PR #3, opened 2023-08-30
create layouts

KonnorRogers:


src/themes/applied.css
```diff
@@ -433,6 +437,10 @@ textarea {
   padding: 0 !important;
 }
 
+.flow > * + * {
+  margin-block-start: var(--wa-flow-spacing);
+}
```
claviska: This should probably be in `sport-awesome.njk` unless we plan on exposing it as a public utility class.

docs/_utilities/scrolling-tables.cjs
```diff
@@ -1,8 +1,13 @@
 /**
- * Turns headings into clickable, deep linkable anchors. The provided doc should be a document object provided by JSDOM.
+ * Turns tables into scrollable tables
  * The same document will be returned with the appropriate DOM manipulations.
  */
 module.exports = function (doc, options) {
+  // We don't want to run this on layouts.
+  if (doc.querySelector("[data-layout='layout-example.njk']")) {
```
claviska, line 7: Is there a more generic way to opt out of this? Or maybe add a class to `<main>` in `docs/_includes/default.njk` and have it target that instead?

docs/_includes/layout-templates/advanced-example.css.njk
claviska, line 1: Do these end in *.css.njk so 11ty will copy them over for us?

docs/_includes/layout-templates/advanced-example.html.njk
claviska, line 1: Same question about *.html.njk extensions.

src/components/details/details.styles.ts
```diff
@@ -76,7 +76,7 @@ export default css`
     display: none;
   }
 
-  .details__body {
+  :not(.details--open) .details__body {
```
claviska, line 79: What was the reason for this again?

claviska (commented):
Huge PR! I've left some comments here and there, but I'm going to dive into the components and use them more before providing feedback there.

docs/_includes/layout-templates/advanced-example.css.njk
KonnorRogers: Yea afaik its the only way to `include` the file with 11ty. Happy to find a better way because whenever these files change it takes like 6seconds to update

src/components/details/details.styles.ts
KonnorRogers: I should add a comment, TLDR is the closing animation gets borked when you dont have this.

src/themes/applied.css
KonnorRogers: I would like to expose it. Happy to consult with @lindsaym-fa .

I ~~stole~~ borrowed it from Andy Bell.

https://andy-bell.co.uk/my-favourite-3-lines-of-css/

KonnorRogers: Moved to sport-awesome for now.

docs/_utilities/scrolling-tables.cjs
KonnorRogers: `<main id='wa-documentation'>` 

```js
if (!doc.querySelector("#wa-documentation")) {
  return
}
```

Maybe? Naming is hard...

docs/_includes/layout-templates/advanced-example.css.njk
KonnorRogers: Found it.

https://www.11ty.dev/docs/copy/#passthrough-by-file-extension

src/themes/applied.css
claviska: We _will_ have utility classes, so exposing it will be fine if we prefix it and document it :)

docs/_includes/layout-templates/advanced-example.css.njk
claviska: TIL

claviska:
I fixed the autoloader in [cd0ad18](https://github.com/shoelace-style/webawesome/pull/3/commits/cd0ad18551c7709e0d63ff4090d9a4c6c2455f70), but it looks like we're still using `<wa-nav-group>` and `<wa-nav-item>` in a few examples.

Do we care about updating these to links or should we remove them?

claviska:
Looks like the CI is failing, but we can probably merge this once that's resolved.

KonnorRogers:
@claviska What's the plan for keeping this gated as a pro component? We dont want it leaking whenever we do a 3.0 release I assume

claviska:
We'll have a pro repo for it. I'm not sure if we want a separate build or if we want to turn the project into a pseudo-monorepo that builds and publishes two targets, though. What are your thoughts on that?

PR #6, opened 2023-09-12
Reduce time it takes to replace strings

KonnorRogers:
Reduce build times by ~6 seconds

### DIFF
--- a/docs/_utilities/replacer.cjs
+++ b/docs/_utilities/replacer.cjs
@@ -13,7 +13,9 @@
  * @param {Replacements} replacements
  */
 module.exports = function (content, replacements) {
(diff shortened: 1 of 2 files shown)


docs/_utilities/replacer.cjs
```diff
@@ -13,7 +13,10 @@
  * @param {Replacements} replacements
  */
 module.exports = function (content, replacements) {
+  let html = content.body.innerHTML;
   replacements.forEach(replacement => {
-    content.body.innerHTML = content.body.innerHTML.replaceAll(replacement.pattern, replacement.replacement);
+    html = html.replaceAll(replacement.pattern, replacement.replacement);
   });
+
+  content.body.innerHTML = html;
```
KonnorRogers: This seems trivial, but by assigning to a string first, THEN using `innerHTML` we reduce the calculations of JSDOM and the replacer function runs in 3 seconds instead of 9 seconds.

claviska (approved):

PR #9, opened 2023-11-02
DRAFT: Layouts review

claviska:
This is the toughest docs page I've worked on so far. 😓 

There's still a lot of room to improve this page, but I think the information is split up into logical sections without too much fluff. We might want to talk about the order of presentation and how we can improve this from users' perspectives. (Maybe @lindsaym-fa has some bandwidth for feedback.)

I made a handful of improvements to the layout component:

- Renamed "skip to main" => "skip to content"
- Renamed "sub-header" => "subheader"
- Some minor CSS adjustments to use design tokens
- JSDoc comment updates
- Removed `main-id` and made the "skip to content" link work automatically by injecting a target into the light DOM using a secret slot 🤫
- Removed the `<wa-button>` dependency since it also imports `<wa-spinner>` and we wanted custom styles for the skip link
- Removed the `<wa-visually-hidden>` dependency to reduce imports since the styles are just a few lines
- Updated the example to use the slot names instead of Title Case.
- Rewrote the docs page based on previous info

The docs page are still a WIP, but I'd like to get some feedback before proceeding. I also have a few suggestions to improve the layout's DX.

## Design Questions

@lindsaym-fa Any suggestions on how to improve the aesthetics of this page? Layout is the most complex component we have and I'm finding it very tricky to document for a few reasons, one being that layout previews need to open in a new tab because they take up the whole page.

Perhaps we can show SVG images of the layout for each example and link out to the preview in a new tab when clicked. Does that sound reasonable? If so, what should those images look like?

Yesterday we talked about a layout anatomy image. Perhaps we can use that as a basis for the previews if you think that's a good way to go.

Do you have any other thoughts or suggestions after looking at the layouts page in the docs?

## Dev Questions

- [ ] How do we want to handle zeroing out the padding/margin of `html/body` elements? @KonnorRogers and I discussed a few ways to handle it programmatically, but I don't think we settled on one. I currently have it documented as a warning in case we decided _not_ to try to handle this automatically.

- [x] Can `nav-state="open"` be changed to a boolean attribute `nav-open`? This is more akin to how we handle open dialogs and drawers.

- [x] Can we add default styles to hide the nav menu on mobile so the user doesn't have to do it? It wasn't obvious to me, and I feel like something like this should just work out of the box.

- [ ] Should we support basic markup in the nav menu? Things like `<h2>` for section headings and `<a>` links? Maybe `<details>` for collapsible areas? I know we have new nav items, but I wonder if those styles are too opinionated and if the component does enough. I think it's worth discussing if we can achieve what we're achieving with nav items using regular markup, even if the answer turns out to be "nope." (Remember, it's easier to give later than to take away later.)

- [ ] What's the plan to clean up the dev examples so they don’t ship with the production website?

--- 

Here's direct link to the preview: https://webawesome-git-layouts-review-font-awesome.vercel.app/components/layout/


KonnorRogers:
1.) I guess we could also check the padding on `<body>` at runtime and emit a console warning? I don't have a good answer here, but we should also include the recommended reset:

```css
html, body {
  min-height: 100%;
  padding: 0;
  margin: 0;
}
```

2.) `nav-open` boolean seems fine to me
3.) Honestly, this seems like the proper default to me.
4.) I can do some exploration work here. The biggest thing is styling, but if we have buttons that could support a `nav-item` like style, then I don't think we need a full blown component. Also means we can defer any abstractions around "button-like" components.
5.) The "example" / "advanced example" layouts?? Or do you mean the the layout examples for things like "sticky header"

claviska:
> The "example" / "advanced example" layouts?? Or do you mean the the layout examples for things like "sticky header"

The stuff located here: https://webawesome-git-layouts-review-font-awesome.vercel.app/layouts/index.html

Some of them we can evolve into the linked previews (if we choose to do it that way) but some seem like dev helpers or POCs.

- https://webawesome-git-layouts-review-font-awesome.vercel.app/layouts/hero-playground/index.html
- https://webawesome-git-layouts-review-font-awesome.vercel.app/layouts/sport-awesome/index.html
- https://webawesome-git-layouts-review-font-awesome.vercel.app/layouts/app-playground/index.html

If we're going to keep any of these examples around for end users, we need to clean them up and make them less janky. For example:

https://github.com/shoelace-style/webawesome/assets/55639/a4025b5b-01c6-4e19-828e-35346854e693

claviska:
> I guess we could also check the padding on <body> at runtime and emit a console warning? I don't have a good answer here, but we should also include the recommended reset:

I've added the recommended styles to the docs.

claviska:
> I guess we could also check the padding on <body> at runtime and emit a console warning? I don't have a good answer here, but we should also include the recommended reset:

I don't love that idea. There might be valid reasons for folks to keep some kind of padding on the body and they wouldn't have a way to suppress it.

KonnorRogers:
@claviska this has been merged into #3 and I'll be continuing work there.

PR #22, opened 2023-12-14
Convert to Starlight

KonnorRogers:
Missing things:

- ~~Search in development (This is tough because Astro seems to lazy load pages in dev to speed things up, but we could leverage some custom record shenanigans in PageFind, definitely not a need to have)~~
- ~~Help & Support under "resources". We autogenerate from a directory. We could add a blank page with a link to issues, or opt-out of "autogenerate" and manually add it.~~
- 404 page doesn't include sidebar and should match previous version.
- ~~Scrolling tables wrapper (Rehype Plugin)~~ Decided not to implement. Only used in ComponentLayout.
- ~~Carousel assets missing~~
- ~~Code / Star / Twitter links in sidebar~~


claviska:
Amazing work! I have no problems merging this as-is, but we'll need some things from the design team.

### Designers

- The docs site is still mostly Starlight. We need to tie in our theming tokens and give the layout/content a facelift.
- Should we center the content? (The alignment looks a bit strange to me, but that might be due to the extra width compared to [Starlight's website](https://starlight.astro.build/getting-started/))
- What other changes/improvements would you suggest for this?

(None of this needs to hold this PR up from being merged!)

claviska (approved):

PR #24, opened 2023-12-22
add the ability to toggle light and dark mode for themes

kelseythejackson:
Adds dark mode for all themes


docs/pages/experimental/themer.md
```diff
@@ -11,7 +11,7 @@ toc: false
     <a href="/">{% include 'logo.njk' %}</a>
     <wa-select name="theme" label="Theme" value="default">
       <wa-option value="default">Default</wa-option>
-      <wa-option value="classic">Classic</wa-option>
+      <wa-option value="default">Classic</wa-option>
```
KonnorRogers: Any reason this was changed from "default" to "classic", when they're two different themes? Was this an accident?

KonnorRogers (commented):
Looks good, just 1 question on themes, and then should be good to go

docs/pages/experimental/themer.md
kelseythejackson: Ooh, that's a slip by me, I'll fix that

kelseythejackson: Okay, that's been updated

PR #29, opened 2024-02-07
Pac Man: Playful theme

frrrances:
This PR adds the Playful theme. 

Playful Light Mode
Playful Dark Mode


src/themes/playful.css
```diff
@@ -244,22 +260,22 @@
   /* Border widths are designed to scale according to a single base value.
      * The base value is intended for calculations and is not used by components directly. */
   --wa-border-width-base: 0.125;
```
lindsaym-fa:
```suggestion
  --wa-border-width-base: calc(2 / 18);
```

If we want to ensure whole pixel values for borders at 100% zoom, one trick I like is using a `calc()` function to divide the desired pixel values by the root font size. That prevents us from having to mess with potentially messy decimals.

Since it looks like we have 2px, 3px, and 4px borders used throughout our components with this theme, we may also want to adjust the calculations below to multiply by `1rem`, `1.5rem`, and `2rem`.

src/themes/playful.css
```diff
@@ -244,22 +260,22 @@
...
-  --wa-panel-border-width: var(--wa-border-width-s);
+  --wa-panel-border-width: 0;
 
   /**
    * Corners
    */
   /* Corners are designed to scale according to a single base value.
    * The base value is intended for calculations and is not used by components directly. */
-  --wa-corners-base: 0.5;
+  --wa-corners-base: 0.75;
```
lindsaym-fa:
```suggestion
  --wa-corners-base: calc(8 / 18);
```

This one's a bit trickier, but we could use a similar `calc()` approach here. Looks like we have 8px, 12px, 24px, and 40px corners at 100% zoom throughout this theme, so 8 might be a good starting point for our base. We can set that in our base and adjust the calculations below multiply by `1rem`, `1.5rem`, `3rem`, and `5rem`.

src/themes/playful.css
```diff
@@ -334,14 +350,404 @@
   --wa-form-controls-required-content-color: inherit;
   --wa-form-controls-required-content-offset: -0.1em;
 
-  --wa-panel-corners: var(--wa-corners-m);
+  --wa-panel-corners: var(--wa-corners-l);
```
lindsaym-fa:
```suggestion
  --wa-panel-corners: var(--wa-corners-m);
```
If we make an adjustment to the `corners` properties above, we can keep using `--wa-corners-m` here. We can reserve `--wa-corners-l` for just the product cards since they have rounder corners than any of our other elements (alerts, message composer, support table, etc.).

src/themes/playful.css
```diff
@@ -334,14 +350,404 @@
...
 
   --wa-flow-spacing: 1.5rem;
 
   /**
    * From 2.x
    */
+  --wa-form-control-toggle-size-s: 0.875rem;
+  --wa-form-control-toggle-size-m: 1.125rem;
+  --wa-form-control-toggle-size-l: 1.375rem;
+
```
lindsaym-fa:
```suggestion
```

These might be remnants from an old version of the stylesheet. These are no longer used by any components.

src/themes/playful.css
```diff
@@ -1,4 +1,7 @@
-@import 'depth_2_chunky.css';
+/* @import 'depth_4_glossy.css'; */
+@import 'depth_3_punchy.css';
+/* @import 'depth_2_chunky.css'; */
```
lindsaym-fa:
```suggestion
```

src/themes/playful.css
```diff
@@ -1,4 +1,7 @@
-@import 'depth_2_chunky.css';
+/* @import 'depth_4_glossy.css'; */
```
lindsaym-fa:
```suggestion
```

src/themes/playful.css
```diff
@@ -296,16 +312,16 @@
...
   --wa-transition-faster: 50ms;
 
   /**
    * Component groups
    */
-  --wa-form-controls-background: var(--wa-color-surface-default);
+  --wa-form-controls-background: transparent;
 
-  --wa-form-controls-corners: var(--wa-corners-s);
+  --wa-form-controls-corners: var(--wa-corners-m);
```
lindsaym-fa:
```suggestion
  --wa-form-controls-corners: var(--wa-corners-pill);
```

We'll have better luck ensuring that form controls are always pill shaped by using this utility property.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+      180deg,
+      color-mix(in oklab, var(--button-gradient-top), var(--wa-color-mix-active)) 0%,
+      color-mix(in oklab, var(--button-gradient-middle), var(--wa-color-mix-active)) 51.88%,
+      color-mix(in oklab, var(--button-gradient-bottom), var(--wa-color-mix-active)) 100%
+    );
+    --border-color-active: var(--button-gradient-middle);
+
+    &::part(base) {
+      padding-inline-start: var(--wa-space-s);
+    }
```
lindsaym-fa:
```suggestion
    &::part(prefix) {
      padding-inline-start: calc(0.25em - var(--border-width));
    }
    &::part(suffix),
    &::part(caret) {
      padding-inline-end: calc(0.25em - var(--border-width));
    }
```

This was causing any button without a prefix icon to look off-center. To even things out, how about we add a bit of padding to the `prefix` and `suffix` slots? I used `em` units so that we don't have to declare separate rules per button size, and I accounted for the button's `--border-width` so that `outline` buttons don't get unnecessarily wide.

(Side note: good heavens, we _have_ to make it easier to change button padding. 😵‍💫 Took me far too long to figure this out...)

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    --label-color-hover: var(--wa-color-neutral-text-on-spot);
+
+    &::part(label) {
+      text-shadow: none;
+    }
+  }
+
+  wa-card {
+    --border-radius: var(--wa-panel-corners);
+  }
```
lindsaym-fa:
```suggestion
```
`wa-card` uses `--wa-panel-corners` by default.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+
+  wa-card img {
+    border-bottom-right-radius: 0;
+    border-bottom-left-radius: 0;
+  }
```
lindsaym-fa:
```suggestion
```
No longer needed with the fixes we merged.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+
+  wa-carousel img {
+    border-radius: var(--wa-panel-corners);
+    box-shadow: none;
+  }
+
+  wa-alert {
+    --border-radius: var(--wa-corners-m);
```
lindsaym-fa:
```suggestion
```
If we adjust the theme's corner calculations and `--wa-panel-corners` above, we won't need to override this.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    --icon-size: var(--wa-font-size-m);
+    --padding: var(--wa-space-s);
+    font-size: var(--wa-font-size-s);
+
+    &::part(message) {
+      padding-left: var(--wa-space-l);
+    }
+  }
+
+  wa-tag {
```
lindsaym-fa, line 558:
```suggestion
  wa-tag {
      --border-radius: var(--wa-corners-pill);
```
We'll probably want to add this. However, it looks like `wa-tag` isn't currently respecting its `--border-radius` property as it should. I can push a commit to fix that.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    &[variant='danger'] {
+      --border-color: var(--wa-color-danger-border-subtle);
+      --background: var(--wa-color-danger-fill-subtle);
+      --content-color: var(--wa-color-danger-text-on-fill);
+    }
+    &[variant='neutral'] {
+      --border-color: var(--wa-color-neutral-border-subtle);
+      --background: var(--wa-color-neutral-fill-subtle);
+      --content-color: var(--wa-color-neutral-text-on-fill);
+    }
```
lindsaym-fa:
```suggestion
```
These are the default values for `wa-alert`, so we should be safe to remove these.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+  }
+
+  wa-badge {
+    --border-radius: var(--wa-corners-l);
```
lindsaym-fa:
```suggestion
    --border-radius: var(--wa-corners-pill);
```

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    margin-inline-end: var(--wa-space-xs);
+  }
+
+  .blog wa-avatar {
+    --size: 2rem;
+  }
+
+  .message-composer wa-card {
+    --background: var(--wa-color-surface-raised);
+    --wa-border-width-s: calc(var(--wa-border-width-base) * 1rem);
```
lindsaym-fa:
```suggestion
```
If we adjust the border width calculations above, we may not need this.

src/themes/playful.css
```diff
@@ -244,22 +260,22 @@
...
   --wa-border-width-m: calc(var(--wa-border-width-base) * 2rem);
   --wa-border-width-l: calc(var(--wa-border-width-base) * 3rem);
 
   --wa-form-controls-border-style: var(--wa-border-style);
-  --wa-form-controls-border-width: var(--wa-border-width-s);
+  --wa-form-controls-border-width: var(--wa-border-width-m);
 
   --wa-panel-border-style: var(--wa-border-style);
-  --wa-panel-border-width: var(--wa-border-width-s);
+  --wa-panel-border-width: 0;
```
lindsaym-fa:
```suggestion
  --wa-panel-border-width: 0rem;
```
Adding a unit here lets us account for border widths in calculations later on.

src/themes/playful.css
```diff
@@ -244,22 +260,22 @@
...
      * The base value is intended for calculations and is not used by components directly. */
   --wa-border-width-base: 0.125;
-  --wa-border-width-s: calc(var(--wa-border-width-base) * 1rem);
+  --wa-border-width-s: calc(var(--wa-border-width-base) * 1.5rem);
   --wa-border-width-m: calc(var(--wa-border-width-base) * 2rem);
   --wa-border-width-l: calc(var(--wa-border-width-base) * 3rem);
 
   --wa-form-controls-border-style: var(--wa-border-style);
-  --wa-form-controls-border-width: var(--wa-border-width-s);
+  --wa-form-controls-border-width: var(--wa-border-width-m);
```
lindsaym-fa:
```suggestion
  --wa-form-controls-border-width: var(--wa-border-width-l);
```

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    font-size: var(--wa-font-size-s);
+
+    &::part(message) {
+      padding-left: var(--wa-space-l);
+    }
+  }
+
+  wa-tag {
+    --border-style: dashed;
+    --border-width: var(--wa-border-width-s);
```
lindsaym-fa:
```suggestion
```
This is the default style for `wa-tag`.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    --size: 2rem;
+  }
+
+  .message-composer wa-card {
+    --background: var(--wa-color-surface-raised);
+    --wa-border-width-s: calc(var(--wa-border-width-base) * 1rem);
+  }
+
+  .message-composer wa-card::part(header) {
+    --wa-color-neutral-border-subtle: var(--wa-color-base-70);
```
lindsaym-fa, line 679:
```suggestion
    --wa-color-neutral-border-subtle: var(--wa-color-base-70);
    border-start-start-radius: calc(var(--border-radius) - var(--border-width));
    border-start-end-radius: calc(var(--border-radius) - var(--border-width));
```

We may want to subtract any borders from our border-radii to ensure we don't get artifacts peeking through (especially in dark mode).

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    padding-top: 0;
+  }
+
+  .product-detail .price {
+    font-weight: var(--wa-font-weight-heavy);
+  }
+
+  .support-table wa-card {
+    --background: var(--wa-color-surface-raised);
+    --border-radius: var(--wa-panel-corners);
```
lindsaym-fa:
```suggestion
```
This is the default style for `wa-card`.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+  }
+
+  .support-table th {
+    font-size: var(--wa-font-size-s);
+    font-weight: var(--wa-font-weight-heavy);
+    background-color: var(--wa-color-surface-lowered);
+  }
+
```
lindsaym-fa, line 712:
```suggestion

  .support-table tr {
    border-top-color: var(--wa-color-surface-lowered);
  }
  
```
This addition should match the mock-ups a bit closer.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+    border-radius: 0 var(--wa-corners-l) 0 0;
+  }
+
+  .support-table tr:last-child td:first-child {
+    border-radius: 0 0 0 var(--wa-corners-l);
+  }
+
+  .support-table tr:last-child td:last-child {
+    border-radius: 0 0 var(--wa-corners-l) 0;
+  }
```
lindsaym-fa:
```suggestion
  .support-table th:first-child {
    border-radius: calc(var(--border-radius) - var(--border-width)) 0 0 0;
  }

  .support-table th:last-child {
    border-radius: 0 calc(var(--border-radius) - var(--border-width)) 0 0;
  }

  .support-table tr:last-child td:first-child {
    border-radius: 0 0 0 calc(var(--border-radius) - var(--border-width));
  }

  .support-table tr:last-child td:last-child {
    border-radius: 0 0 calc(var(--border-radius) - var(--border-width)) 0;
  }
```

Prevents peeking artifacts. This also allows us to change the card's border radius without needing to update these fellas.

src/themes/playful.css
```diff
@@ -351,107 +768,200 @@
...
+
+  --wa-color-neutral-spot: var(--wa-color-base-40);
+  --wa-color-neutral-spot-darker: var(--wa-color-base-70);
+  --wa-color-neutral-fill-subtle: var(--wa-color-base-40);
+  --wa-color-neutral-fill-highlight: var(--wa-color-base-30);
+  --wa-color-neutral-border-subtle: var(--wa-color-base-30);
+  --wa-color-neutral-border-highlight: var(--wa-color-base-70);
+  --wa-color-neutral-text-on-spot: var(--wa-color-base-20);
+  --wa-color-neutral-text-on-fill: var(--wa-color-base-95);
+  --wa-color-neutral-text-on-surface: var(--wa-color-base-70);
```
lindsaym-fa:
```suggestion
  --wa-color-neutral-spot: var(--wa-color-base-40);
  --wa-color-neutral-spot-darker: var(--wa-color-base-70);
  --wa-color-neutral-fill-subtle: var(--wa-color-base-20);
  --wa-color-neutral-fill-highlight: var(--wa-color-base-30);
  --wa-color-neutral-border-subtle: var(--wa-color-base-30);
  --wa-color-neutral-border-highlight: var(--wa-color-base-40);
  --wa-color-neutral-text-on-spot: var(--wa-color-base-20);
  --wa-color-neutral-text-on-fill: var(--wa-color-base-95);
  --wa-color-neutral-text-on-surface: var(--wa-color-base-70);
```
I think this brings us a bit closer to the mock-ups.

src/themes/playful.css
```diff
@@ -351,107 +768,200 @@
...
 
-/* #region Custom styles */
-wa-button {
-  --border-color-hover: var(--border-color);
-  --border-color-active: var(--border-color);
-  --box-shadow-color: var(--border-color);
+  wa-card {
+    --box-shadow: none;
+    --border-width: var(--wa-border-width-s);
+    --border-color: var(--wa-color-base-30);
```
lindsaym-fa:
```suggestion
```

It looks like only the message composer has a lighter border color than the default cards, so could we add a separate rule for that?

src/themes/playful.css
```diff
@@ -351,107 +768,200 @@
...
+  .hero-background {
+    background: linear-gradient(180deg, var(--wa-color-green-60) 69.42%, var(--wa-color-green-30) 100%);
   }
 
-  &[variant='success'] {
-    --border-color: var(--wa-color-success-spot-darker);
+  .message-composer wa-card::part(header) {
+    background: var(--wa-color-surface-default);
+    border-top-left-radius: var(--wa-corners-l);
+    border-top-right-radius: var(--wa-corners-l);
```
lindsaym-fa:
```suggestion
```
If we add `border-start-start-radius: calc(var(--border-radius) - var(--border-width));` and `border-start-end-radius: calc(var(--border-radius) - var(--border-width));` for the message composer above, we don't have to make a change specifically for dark mode.

src/themes/playful.css
```diff
@@ -334,14 +350,415 @@
...
+
+  code {
+    font-weight: var(--wa-font-weight-normal);
+    font-size: var(--wa-font-size-s);
+  }
+
+  img {
+    border-radius: var(--wa-corners-m);
+    box-shadow: var(--wa-shadow-level-1);
+  }
```
lindsaym-fa: Because we wrapped our images with `.landscape-frame` or `.square-frame`, I don't think this rule is having the desired effect. Do we want to declare these styles for those frame classes instead?

src/themes/playful.css
frrrances: OH! Yes I noticed that once i made the change but forgot to change it. Thanks for catching it!

src/themes/playful.css
frrrances: Ooo nice!

src/themes/playful.css
frrrances: Yup to background and content color but it looks like tag borders are set to be `*-border-highlight` instead of `*-border-subtle` - Is that something I affected with another change?

lindsaym-fa (approved):
Looks good! 🎉

PR #32, opened 2024-02-12
Backport 1868

claviska:
Backport [#1868](https://github.com/shoelace-style/shoelace/pull/1868)


package.json
```diff
@@ -47,7 +47,8 @@
     "build": "node scripts/build.js",
     "docs:build": "cd docs && npm run build -- --silent",
     "verify": "npm run prettier:check && npm run lint && npm run build && npm run test",
-    "postinstall": "npx playwright install && ./scripts/clear-vercel-patches.sh && npm i --ignore-scripts && npx patch-package",
+    "postinstall": "./scripts/clear-vercel-patches.sh && npm i --ignore-scripts && npx patch-package",
+    "prepare": "npx playwright install",
```
KonnorRogers: The postinstall is only meant for the docs site. Should that move somewhere else?

claviska: I wasn't sure about it, so I left it as-is. If it's development-only, we want to move it to `prepare` so it won't run when WA is installed as a dependency.

Should I move the rest over?

PR #34, opened 2024-02-13
Pac Man: Font Awesome theme

lindsaym-fa:
- Rounds out the Font Awesome theme with imagery, a header, and fine-tuning
- Adds a missing custom property for `--checked-icon-color` to `wa-checkbox`


src/themes/fa.css
```diff
@@ -459,47 +568,73 @@ wa-button:not([variant='text']) {
...
+  & wa-rating {
+    --symbol-size: var(--wa-font-size-m);
+  }
 }
 
-wa-switch {
-  --background: var(--wa-color-neutral-spot);
-  --border-color: var(--wa-color-neutral-spot);
+wa-carousel {
+  --pagination-color-activated: var(--wa-color-brand-spot);
```
frrrances, line 580: Unrequested opinion... what if the resting color was a little lighter?
```suggestion
  --pagination-color-activated: var(--wa-color-brand-spot);
  --pagination-color-resting: var(--wa-color-gravy-80);  
```

src/themes/fa.css
```diff
@@ -459,47 +568,73 @@ wa-button:not([variant='text']) {
...
-wa-alert {
-  --content-color: var(--wa-color-text-normal);
+wa-checkbox,
+wa-radio {
+  --checked-icon-color: var(--wa-color-surface-default);
+}
 
-  &[variant='brand'] {
-    --icon-color: var(--wa-color-brand-spot-darker);
+wa-radio-group > wa-radio-button {
```
frrrances, line 589: Love the depth on these!

src/themes/fa.css
```diff
@@ -459,47 +568,73 @@ wa-button:not([variant='text']) {
...
-  &[variant='neutral'] {
-    --icon-color: var(--wa-color-neutral-spot-darker);
+  & wa-icon-button {
+    color: var(--wa-color-text-normal);
   }
 }
+
+wa-input > wa-icon,
+wa-select > wa-icon {
+  color: var(--wa-color-text-quiet);
```
frrrances: These seem a little more font-awesome-y without this override...

frrrances (commented):
Looks great, @lindsaym-fa! Code is looking awesome, I left a few small optional comments for ya. 

As for the hero, I'm not 100% into it yet either. @jabronus may have some better thoughts, but here are mine in case they're helpful: 

- swap the grape for Font Awesome blue, or navy?
- bring the hero title color into more alignment with the surrounds, like tint it a little? 
- use more illustration like we do on fontawesome.com, like a variation on one of these that is more baked goods oriented?

open-source-illustration
icons-bg

Take those with a giant lump of salt (or whatever the kids are saying these days) and just use what's useful!

jabronus:
This is looking fantastic 🤩

The think the idea of reusing some illos from the site is a great one! If you want a hand with that in any way, let me know.

The only thing that's throwing me off a tad is the depth on buttons in dark mode. It feels like a photo negative, but there may not be a way around that stylistic translation. Could also be I'm not used to seeing a dark mode FA and I just need to acclimate 😅 It doesn't have the same contrast-y punch as light mode, but I think my brain expects to see something more like this:

buttons

Also take that with a giant lump of salt!

lindsaym-fa:
> The only thing that's throwing me off a tad is the depth on buttons in dark mode. It feels like a photo negative, but there may not be a way around that stylistic translation. Could also be I'm not used to seeing a dark mode FA and I just need to acclimate 😅 It doesn't have the same contrast-y punch as light mode, but I think my brain expects to see something more like this:
> 
> buttons


Totally, totally agree, @jabronus. I was going for "old" Web Awesome's `inverted` class, which seems to serve as a spot dark mode in FA:
Screenshot 2024-02-13 at 4 36 42 PM

I'll try out the more realistic lighting model!

jabronus:
Ah nice, didn't know that existed! Feel free to ignore me

src/themes/fa.css
lindsaym-fa: @frrrances, it looks like most icons in FA inputs are pretty dark...
Screenshot 2024-02-15 at 5 29 23 PM

...so I went a bit darker with my overrides.
Screenshot 2024-02-15 at 5 28 10 PM


Let me know how that sits with you!

frrrances (approved):
Looks amaaaazing, @lindsaym-fa! 🤩 ✨

PR #52, opened 2024-02-15
Pac-Man: Polishing round 1

frrrances:
Small polishing on the kitchen sink and themes.


src/themes/brutalist.css
```diff
@@ -351,6 +351,35 @@
...
   }
 
+  pre,
+  code {
+    font-size: var(--wa-font-size-m);
+    background-color: var(--wa-color-neutral-fill-subtle);
+  }
+
+  /* syntax highlighting */
+  .token {
```
frrrances, line 361: The default syntax highlighting colors were close but looked a bit duller, so I added these overrides. If we don't want overrides, we can probably just do without them and use the default styling. (Same for all the syntax highlighting overrides below...)

docs/pages/experimental/kitchen-sink.md
```diff
@@ -1168,13 +1169,23 @@ toc: false
...
+  }
+
   wa-input::part(input) {
     width: 100%;
   }
 
   .square-frame,
   .landscape-frame {
     overflow: hidden;
+    border-radius: calc(var(--wa-corners-m) - var(--wa-panel-border-width));
```
lindsaym-fa: We have some rounding sneaking into the `wa-card` images from this style.

Screenshot 2024-02-15 at 6 16 36 PM
Screenshot 2024-02-15 at 6 16 57 PM
Screenshot 2024-02-15 at 6 19 32 PM

lindsaym-fa (changes requested):
Changes look great! Just one small thing to touch up.

lindsaym-fa (approved):
All looks good to me! 🤩

PR #54, opened 2024-02-21
Pac-man: Responsify

frrrances:
This PR adds a small amount of responsive layout to the kitchen sink so it's a little more resilient for demoing (but it only handles down to ipad-ish size so not fully responsive).

### DIFF
--- a/docs/pages/experimental/kitchen-sink.md
+++ b/docs/pages/experimental/kitchen-sink.md
@@ -1151,11 +1151,16 @@ toc: false
   }
 
   /* page layout */
+  .content {
+    max-width: none;
+    margin: 0;
+  }
+
   .preview-container {
     background: var(--wa-color-surface-lowered);
     padding: 0;
-    margin: 0;
-    translate: calc((var(--knobs-width) + 2rem) / 2);
+    margin-left: calc((var(--knobs-width) - 2rem) );
+    /* translate: calc((var(--knobs-width) + 2rem) / 2); */
     border: var(--wa-border-width-m) var(--wa-color-neutral-border-subtle) var(--wa-border-style);
   }
 
@@ -1179,7 +1184,7 @@ toc: false
   }
 
   .strata {
-    padding: var(--wa-space-3xl) 10%;
+    padding: var(--wa-space-3xl) 7%;
   }
 
   pre,
@@ -1249,8 +1254,13 @@ toc: false
   }
 
   .product-card {
-    grid-column: span 4;
+    grid-column: span 12;
     position: relative;
+  }
+
+  .product-card {
+    max-width: 350px;
+    margin: 0 auto;
   }
 
   .product-card .title-rating {
@@ -1268,8 +1278,12 @@ toc: false
   /* strata - blog post */
 
   .blog .column-post-header {
-    grid-column: 1 / 5;
+    grid-column: span 12;
     position: relative;
+  }
+
+  .blog .post-body {
+    grid-column: span 12;
   }
 
   .blog .post-header {
@@ -1292,9 +1306,6 @@ toc: false
     gap: var(--wa-space-s);
   }
 
-  .blog .post-body {
-    grid-column: 5 / 13;
-  }
 
   /* strata - message composer */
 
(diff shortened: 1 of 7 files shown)


docs/pages/experimental/kitchen-sink.md
```diff
@@ -1179,7 +1184,7 @@ toc: false
   }
 
   .strata {
-    padding: var(--wa-space-3xl) 10%;
+    padding: var(--wa-space-3xl) 7%;
```
lindsaym-fa: @frrrances How would you feel about using a theme value for padding in the inline direction (could do `var(--wa-space-3xl)` all around, or we may want a smaller inline value) instead of a percentage, then applying something like `max-inline-size: 1280px;` and `margin-inline: auto;` to the strata?

I'm not sure if the `max-inline-size` would cause us headaches in the hero areas, but it should keep all of the body content from spilling beyond readability.

docs/pages/experimental/kitchen-sink.md
```diff
@@ -1249,10 +1254,15 @@ toc: false
   }
 
   .product-card {
-    grid-column: span 4;
+    grid-column: span 12;
     position: relative;
   }
 
+  .product-card {
+    max-width: 350px;
```
lindsaym-fa:
```suggestion
    max-width: 45ch;
```

Completely optional: how would you feel about tying the card widths to the font size/family? Could help ensure consistent, readable line lengths between themes.

docs/pages/experimental/kitchen-sink.md
```diff
@@ -1445,35 +1462,47 @@ toc: false
...
     }
   }
 
-  @media screen and (max-width: 1450px) {
-    .cards {
-      grid-template-columns: 1fr 1fr;
-    }
+  @media screen and (min-width: 1450px) {
+
   }
```
lindsaym-fa:
```suggestion
```

Just some cleanup!

src/themes/brutalist.css
```diff
@@ -699,6 +699,17 @@
...
+  /* responsive */
+  @media screen and (min-width: 1324px) and (max-width: 1505px) {
+    wa-card.product-card {
+      --padding: var(--wa-space-s);
+
+      wa-button {
+        margin-inline-end: 0;
+      }
+    }
+  }
```
lindsaym-fa: Are these rules needed? It's hard for me to tell what we gain from them.

src/themes/playful.css
```diff
@@ -762,6 +762,17 @@
...
+  /* responsive */
+  @media screen and (min-width: 1324px) and (max-width: 1505px) {
+    wa-card.product-card {
+      --padding: var(--wa-space-s);
+
+      wa-button {
+        margin-inline-end: 0;
+      }
+    }
+  }
```
lindsaym-fa: Similar to Brutalist: are these rules needed?

lindsaym-fa (commented):
Awesome work! 🎉 Just a few minor suggestions/questions.

src/themes/brutalist.css
frrrances: They tighten up the spacing on the buttons in the cards so they don't stack before the break. Breaking earlier seemed a bit premature, but I didn't like the stack so I tightened the space. Think that's the right call here?

src/themes/playful.css
frrrances: See above :)

src/themes/brutalist.css
frrrances: Normally I would make the buttons a bit smaller or tighten the padding in them, but shied away from those...

lindsaym-fa: Gotcha! It's mostly the change to the card padding that feels a bit off to me:

https://github.com/shoelace-style/webawesome/assets/126139086/3c229d01-d454-405c-a7cd-0ba00a78d675

Without changing the card padding, it doesn't seem like the buttons get to a point where they stack (at least on my end):

https://github.com/shoelace-style/webawesome/assets/126139086/5572fa0f-b1c7-4716-a8e2-29a9c257cd26

frrrances: Well, that's funny but you're right we don't need those rules anymore. I'll pull 'em out here and in playful!

PR #56, opened 2024-02-22
Pac man: Glassy touchup

frrrances:
This PR adds glassy hero styling and theme images.


lindsaym-fa (approved):
Looks awesome!

I'll move the custom styles into a @container query on my branch, so this is ready to go from my POV.

PR #57, opened 2024-02-23
Pac man: Themer controls

lindsaym-fa:
- Redesigns the themer controls and color mode toggle for our new kitchen sink
- Isolates the styles of the controls so they are largely unaffected by the current theme
- Renames "Chic" to "Active"
- Renames "Headwind" to "Migration"
- Fixes an issue with misplaced tooltips and menus when the highest depth level is applied
- Fixes an issue in Playful where the blog post header didn't scroll with content


src/themes/premium.css
```diff
@@ -435,280 +435,282 @@
...
+      &.attr-value {
+        color: var(--wa-color-yellow-80);
+      }
+    }
+  }
 }
 
 /* responsive */
-@media screen and (min-width: 1324px) {
+@container preview (min-width: 1100px) {
```
frrrances: Thanks for adjusting this.

docs/pages/experimental/kitchen-sink.md
```diff
@@ -922,166 +1223,181 @@ toc: false
   });
 
   // Icons
-    import { registerIconLibrary } from '/dist/utilities/icon-library.js';
+  import { registerIconLibrary } from '/dist/utilities/icon-library.js';
+
+  // Ensure regular icons are always available for the knobs
+  registerIconLibrary('fa-classic-regular', {
+    resolver: name => `https://ka-f.fontawesome.com/releases/v6.5.1/svgs/regular/${name}.svg`
```
frrrances, line 1253: Would it be better to keep them locally?

docs/pages/experimental/kitchen-sink.md
```diff
@@ -922,166 +1223,181 @@ toc: false
...
+        case 'thin':
+          iconLibrary = 'sharp-thin';
           break;
         default:
-          hide(iconStyle);
+          iconLibrary = 'sharp-solid';
       }
+      // Ensures sharp-solid variations are available for ratings, etc.
+      registerIconLibrary('always-solid', {
+        resolver: name => `https://ka-f.fontawesome.com/releases/v6.5.1/svgs/sharp-solid/${name}.svg`
```
frrrances, line 1372: Same question here...

frrrances (approved):
@lindsaym-fa This is fantastic work! I love the look of the controls and the code looks solid. I had a couple small questions/comments but no blockers. The diffs on the theme css files were pretty terrible - let me know If you did more than wrap them in the container query and want me to review specific changes.

docs/pages/experimental/kitchen-sink.md
lindsaym-fa: That's a great question. The default icon library for WA also pulls icons from the same URL, so I'm inclined to keep the same approach here. The main goal is to make sure these icons aren't overwritten by the custom icon sets associated with each theme. There are definitely benefits to keeping these icons locally if we can't connect to that URL, though...

PR #59, opened 2024-02-23
Pac man: Fix tooltip contrast in brutalist message composer header

frrrances:
Fixes: https://3.basecamp.com/4301850/buckets/35892254/todos/7113131633


lindsaym-fa (approved):
Short and sweet. ✔️

PR #62, opened 2024-02-29
[DRAFT]: first pass at form association in `<sl-textarea>`

KonnorRogers:


package.json
```diff
@@ -100,8 +101,8 @@
     "cspell": "^6.18.1",
     "custom-element-jet-brains-integration": "^1.4.0",
     "custom-element-vs-code-integration": "^1.2.1",
-    "dedent": "^1.5.1",
     "custom-element-vuejs-integration": "^1.0.0",
+    "dedent": "^1.5.1",
```
claviska, line 105: I'm not seeing where this is being used. Was it added accidentally?

claviska:
I was looking to see if constraint validation was working properly, and it looks like it is. However, the textarea shows a message before other fields.

CleanShot 2024-03-04 at 10 24 52@2x

I'm assuming this is because the ElementInternals hook runs before our previous form controller, but wanted to make sure this is expected.

claviska:
It's nice that the boilerplate validation methods are offloaded but, because of that, it looks like they're not showing up in the properties table. I'm guessing the CEM analyzer doesn't read members from extended classes.

https://webawesome-git-konnorrogers-form-association-font-awesome.vercel.app/components/textarea/#methods

I applaud your work on the library, but it does raise a tiny concern that we're pulling in another "third-party" dependency. Devil's advocate: is the benefit worth the cost of the abstraction here?

package.json
KonnorRogers: Nope. Its a holdover from 11ty. Don't know why it got reordered.

KonnorRogers:
> I was looking to see if constraint validation was working properly, and it looks like it is. However, the textarea shows a message before other fields.
> 
> CleanShot 2024-03-04 at 10 24 52@2x
> 
> I'm assuming this is because the ElementInternals hook runs before our previous form controller, but wanted to make sure this is expected.

That would be my guess. I'd have to investigate a little further.

> It's nice that the boilerplate validation methods are offloaded but, because of that, it looks like they're not showing up in the properties table. I'm guessing the CEM analyzer doesn't read members from extended classes.
> 
> https://webawesome-git-konnorrogers-form-association-font-awesome.vercel.app/components/textarea/#methods
> 
> I applaud your work on the library, but it does raise a tiny concern that we're pulling in another "third-party" dependency. Devil's advocate: is the benefit worth the cost of the abstraction here?

We could pull in the code internally 🤷‍♂️

PR #118, opened 2024-05-30
Implement roving tabindex for tab-group

KonnorRogers:


src/components/tab/tab.ts
```diff
@@ -52,16 +54,7 @@ export default class WaTab extends WebAwesomeElement {
...
-
-  /** Sets focus to the tab. */
-  focus(options?: FocusOptions) {
-    this.tab.focus(options);
-  }
-
-  /** Removes focus from the tab. */
-  blur() {
-    this.tab.blur();
+    this.tabIndex = -1;
```
claviska: If someone programmatically disables the active tab, the roving tab index is lost so you can't tab into it anymore. Perhaps we should expose a function to tell the Tab Group to reset the tabindex if the active tab gets disabled.

And this will prevent active styles from being applied to disabled tabs.

```suggestion
    this.active = false;    
```

KonnorRogers: 🤔 good call...let me play around with it

claviska (approved):

PR #120, opened 2024-05-31
Events subclass Event instead of using CustomEvent

claviska:
This is a big PR because it touches most components. All tests are currently passing, but there are a few things I'd like to do before we merge this one:

- [x] Update events to subclass `Event` instead of using `CustomEvent` (see https://github.com/shoelace-style/shoelace/issues/1999)
- [x] Remove `WebAwesomeElement.emit()`
- [x] Update events section of contribution guidelines
- [x] the `show` and `hide` events are cancelable as defined, but we only listen for `event.defaultPrevented` in `<wa-details>`. These should be consistent across all components, so we need to add checks to prevent show/hide to the others that use these events.
   - [x] Details (already does)
   - [x] Dialog
   - [x] Drawer
   - [x] Dropdown
   - [x] Select
   - [x] Tooltip


src/components/select/select.styles.ts
```diff
@@ -306,6 +306,7 @@ export default css`
   }
 
   .select__listbox ::slotted(small) {
+    display: block;
```
KonnorRogers, line 309: thanks for backporting this!

src/internal/webawesome-element.ts
```diff
@@ -1,98 +1,12 @@
 import { CustomErrorValidator } from './validators/custom-error-validator.js';
 import { LitElement, type PropertyValues } from 'lit';
 import { property } from 'lit/decorators.js';
-
-// Match event type name strings that are registered on GlobalEventHandlersEventMap...
-type EventTypeRequiresDetail<T> = T extends keyof GlobalEventHandlersEventMap
```
KonnorRogers, line 6: so happy this is gone

KonnorRogers (approved):
LGTM!!

PR #123, opened 2024-06-07
Tooltips associated via selectors

### LINKED ISSUE #93: Tooltips should stop wrapping targets and link to them using ids instead
This will prevent tooltips from being clipped. _Or_ should we keep it as is and use the Popover API to prevent clipping and remove hoist?

- Tooltips should disappear when the anchor is clicked
- We should probably remove the `trigger` property and follow ARIA APG guidelines (hover/focus)

---

Let's make sure this also fixes:

- https://github.com/shoelace-style/shoelace/issues/1712

KonnorRogers:
Tooltips are now associated by doing:

```html
<wa-button id="my-button">I'm a button with a tooltip</wa-button>
<wa-tooltip for="my-button">I'm a tooltip</wa-tooltip>
```

Fixes #93


src/components/tooltip/tooltip.ts
```diff
@@ -292,13 +350,10 @@ export default class WaTooltip extends WebAwesomeElement {
...
+        .anchor=${this.anchor}
       >
-        ${'' /* eslint-disable-next-line lit-a11y/no-aria-slot */}
-        <slot slot="anchor" aria-describedby="tooltip"></slot>
-
-        ${'' /* eslint-disable-next-line lit-a11y/accessible-name */}
-        <div part="body" id="tooltip" class="tooltip__body" role="tooltip" aria-live=${this.open ? 'polite' : 'off'}>
-          <slot name="content">${this.content}</slot>
+        <div part="body" class="tooltip__body">
+          <slot></slot>
```
claviska, line 353: Nice. We should eliminate the `content` property now that content is slotted in.

claviska (commented):
Amazing — and they read properly in screen readers! 🤩

I left some comments in the code and a couple more questions:

- In tooltip.md, we need to update the content to reflect this update. One example (but there may be more) is "A tooltip's target is its _first child element_"
- Are we still planning to add the Popover API to this one, or will that be a separate PR?

src/components/tooltip/tooltip.ts
KonnorRogers: Good call. Content property has been removed! Docs updated!

PR #128, opened 2024-06-14
Add new button appearances

lindsaym-fa:
Updates `<wa-button>` to support a number of new styles with an `appearance` attribute —

- `filled` is a classic, bold call-to-action
- `tinted` is a softer, lower emphasis style
- `outline` is the same outline style we know and love
- `text` is a backgroundless button that now supports all semantic color `variant`s

`outline` and `text` are no longer standalone attributes.


lindsaym-fa:
@claviska Here's a that `appearance` attribute we were talking about!

Curious on your thoughts on naming. I used `filled` to match the similar attribute for `wa-input` and `wa-select`, but it's a bit inconsistent with the existing `outline` style. Do we want to rename `outline` to `outlined`? Do we want to swap `filled` > `fill` and `tinted` > `tint`? Something else entirely?

claviska:
Those tinted buttons look great! The name seems representative of what they are, so no complaints on that.

> Do we want to rename outline to outlined? Do we want to swap filled > fill and tinted > tint? Something else entirely?

I don't have a strong opinion on it, but agree they should be consistent. If I _had_ to pick, I'd probably go with the -ed variations just because they read better in my head — "button appearance equals filled" vs. "button appearance equals fill".

Either works fine!

claviska (approved):
These look great! Thanks for the last minute add.

PR #134, opened 2024-06-18
fix timing issues with dropdown updating triggers

KonnorRogers:


claviska (approved):
I'm good with this as a quick fix. I still have to merge dropdown + menu so this will be moot soon anyways. Thanks!

PR #143, opened 2024-06-21
updated usage example

kelseythejackson:
Since the js way can't be set declaratively, I deleted that example completely. While it's still possible to do, deemphasizing it may be the best course of action. 

Also, I added a link to the [html named colors](https://www.w3schools.com/colors/colors_names.asp) so customers would have a jumping off point should they need it.


claviska (approved):

claviska:
Thanks, @kelseythejackson!

PR #153, opened 2024-07-01
fix radio group not reading number of radios in chrome

KonnorRogers:


claviska (approved):
Is there some documentation explaining why we need to do this, or is it more of a hack to get it working?

I'm fine with the change, just wondering how/when this pattern is required :)

KonnorRogers:
> Is there some documentation explaining why we need to do this, or is it more of a hack to get it working?
> 
> I'm fine with the change, just wondering how/when this pattern is required :)

I just jiggled the handle. Honestly...I have no clue. I could probably go into the accessibility inspector, I kinda just discovered it on accident 🙃

PR #155, opened 2024-07-05
Fix component alignment

### LINKED ISSUE #91: Design tweaks and improvements
Some alignment issues and other tweaks have been brought up by users. I know we're redesigning things, but I'm dropping this here to make sure we don't forget to check this for 3.0.

https://github.com/shoelace-style/shoelace/issues/1892

lindsaym-fa:
Covers issue #91, addressing [#1892 from Shoelace](https://github.com/shoelace-style/shoelace/issues/1892).

These changes ensure that all bite-sized components (inputs, ranges, and the like) share the same center alignment and that none of these have excess space at the top or bottom.

Notably, `--tooltip-offset` for `<wa-range>` is now relative to the _thumb_ rather than the _track_. Tooltip positioning had to be refactored to account for the using of `display: flex;` for the internal input, and a thumb-relative `--tooltip-offset` yields more predictable results when other properties like `--track-height` or `--thumb-size` are manipulated.

Light blue lines in the screenshots below show the top and bottom boundaries of the components. Red lines show the center of the components.

Before:
Screenshot 2024-07-05 at 7 00 10 PM
After:
Screenshot 2024-07-05 at 6 58 04 PM


claviska (approved):
Awesome! Thanks

PR #157, opened 2024-07-17
Initial SSR implementation

KonnorRogers:


.gitignore
```diff
@@ -4,6 +4,7 @@ _site
 package.json
 package-lock.json
 dist
+unbundled-dist
```
KonnorRogers: This is temporary. Ideally it should be "/dist"

docs/docs/components/rating.md
```diff
@@ -110,8 +110,12 @@ You can provide custom icons by passing a function to the `getSymbol` property.
 ```html {.example}
 <wa-rating label="Rating" class="rating-hearts" style="--symbol-color-active: #ff4136;"></wa-rating>
 
-<script>
+<script type="module">
   const rating = document.querySelector('.rating-hearts');
+
+  await customElements.whenDefined('wa-rating')
+  await rating.updateComplete
```
KonnorRogers: This seems small, but without this, Lit will raise a hydration error. This is going to be problematic as time goes on. I know there's discussion about hydration being more error tolerant, but it will be important for us to have good dev tools to be able to help users who do end up using SSR be able to navigate these issues.

docs/assets/scripts/turbo.js
```diff
@@ -43,6 +43,30 @@ function restoreScrollPosition(event) {
...
+    });
+  })(newElement);
+}
+
+// Fixes an issue with DSD keeping the `<template>` elements hanging around in the lightdom.
+// https://github.com/hotwired/turbo/issues/1292
+['turbo:before-render', 'turbo:before-stream-render', 'turbo:before-frame-render'].forEach(eventName => {
+  document.addEventListener(eventName, fixDSD);
+});
+
```
KonnorRogers, line 69: Feels like we need to add a section about Turbo integration.

src/components/radio-button/radio-button.ts
```diff
@@ -84,6 +84,21 @@ export default class WaRadioButton extends WebAwesomeFormAssociatedElement {
...
+  /**
+   * Used for SSR. if true, will show slotted suffix on initial render.
+   */
+  @property({ type: Boolean, attribute: 'with-suffix' }) withSuffix = false;
+
+  /**
+   * Used for SSR. if true, will show slotted suffix on initial render. (should this be withDefault, since its the default slot??)
+   */
+  @property({ type: Boolean, attribute: 'with-label' }) withLabel = false;
+
```
KonnorRogers, line 101: This is the pattern I was referring to.

For SSR users they can pass this in. For CSR users, they just use the components like normal

docs/assets/styles/docs.css
```diff
@@ -92,19 +83,6 @@ wa-page > header {
   }
 }
 
-/* Navigation sidebar */
-wa-page[view='desktop']::part(navigation) {
-  border-right: var(--wa-border-style) var(--wa-panel-border-width) var(--wa-color-neutral-border-quiet);
-}
```
KonnorRogers, line 98: These all moved because they're not really compatible with SSR.

src/themes/applied.css
```diff
@@ -31,17 +36,6 @@ body {
...
-:not(:defined),
-:not(:defined) * {
-  opacity: 0;
-}
-
-:defined {
-  opacity: 1;
-  transition: 0.1s opacity;
-}
-
```
KonnorRogers: this defeats the entire purpose of SSR. Do we still want to leave this? Should we make this opt-in / opt-out?


Technically we can do:

```css
/* Show custom elements only after they're registered */
:not(:defined, [defer-hydration]),
:not(:defined, [defer-hydration]) * {
  opacity: 0;
}

:defined {
  opacity: 1;
  transition: 0.1s opacity;
}
```

.gitignore
```diff
@@ -4,6 +4,7 @@ _site
 package.json
 package-lock.json
 dist
```
KonnorRogers: dist-cdn/

docs/assets/scripts/turbo.js
KonnorRogers: Add note about Turbo in SSR

src/components/carousel/carousel.ts
```diff
@@ -111,6 +119,7 @@ export default class WaCarousel extends WebAwesomeElement {
   }
 
   protected firstUpdated(): void {
+    // this.firstUpdatedHasBeenCalled = true
```
KonnorRogers: get rid of this

web-test-runner.config.js
```diff
@@ -1,16 +1,31 @@
...
 const metadata = JSON.parse(readFileSync('./dist/custom-elements.json'), 'utf8');
+const serverComponents = [];
 const componentImports = getAllComponents(metadata).map(component => {
   const name = component.tagName.replace(/^wa-/, '');
-  return `./dist/components/${name}/${name}.js`;
+
+  // "qr-code" is the only component currently not supported by SSR.
+  if (name !== 'qr-code') {
+    serverComponents.push(`/unbundled-dist/components/${name}/${name}.js`);
+  }
```
KonnorRogers: does get SSRed now

PR #159, opened 2024-08-05
Components Overview/Index

### LINKED ISSUE #156: Component index
We've been wanting a component index, and a friend even started [a discussion with the idea](https://github.com/shoelace-style/webawesome-alpha/discussions/46), so let's do it!

lindsaym-fa:
Adds a page with a thumbnail grid of all standalone Web Awesome components. The goal is to give WA users a better platform to browse WA component offerings, whereas the sidebar provides a scannable list effective for when users already have a specific component in mind.

I opted to exclude child components (wa-carousel-item, wa-breadcrumb-item, etc.) since these are discoverable through their parent components. (Users may not be as likely to have success discovering these by 'browsing.') 

Components are organized into categories in the index but kept as a single, alphabetical list in the sidebar. This supports the distinct goals of each entry point: users looking for a specific component can scan the list predictably in alphabetical order without the barrier of having to guess its category, while users who are browsing to discover components in the index get an extra hint at each components' purpose.

Feedback appreciated. 😊


claviska:
These look fantastic! 🥳 

If you haven't already thought about it, is there an illustration for components that don't have one yet? For example, if we create new component I wouldn't expect to have an image for it right away, so a placeholder might be useful.

> I opted to exclude child components (wa-carousel-item, wa-breadcrumb-item, etc.) since these are discoverable through their parent components. (Users may not be as likely to have success discovering these by 'browsing.')

Good call. I've also been thinking that we should probably remove the examples from those child component pages and just link back to the main page. Most of them are duplicates of the main component anyways and a bit out of context. (The pages do need to exist, however, for the API tables.)

lindsaym-fa:
@claviska Added this little number for future components that don't yet have thumbnails:
Screenshot 2024-08-06 at 5 37 06 PM
Screenshot 2024-08-06 at 5 37 10 PM

claviska (approved):

PR #161, opened 2024-08-13
Fix card footer border radius

lindsaym-fa:
A test site from one of our backers revealed that card footers don't have any border radius. So, when a background color is applied specifically to the footer, the corners don't match those of the parent card.

Their example:
Screenshot 2024-08-13 at 1 33 00 PM

This simple change resolves the issue.


claviska (approved):
Nice catch!

PR #169, opened 2024-09-04
Layout utilities

### LINKED ISSUE #165: Additional layout utilities
`wa-page` is a brilliant tool for full-page layouts, but how might we make it easier for devs to organize content within sections of a page?

We'll draw inspiration from https://every-layout.dev/, global design systems like Material, and front-end libraries like Bootstrap to see how others approach the problem and evaluate what devs might be familiar/comfortable with (and how we can make it better). Extra layout utilities should also benefit our own patterns, too.

lindsaym-fa:
This branch introduces a number of utility classes with the goal to make complex layouts easy to compose. Each class distills some aspect of flexbox or grid into a distinct, digestible part that is automatically responsive.

New layout classes:
- `wa-cluster`
- `wa-flank`
- `wa-frame`
- `wa-grid`
- `wa-split`
- `wa-stack`

New style utility classes:
- `wa-align-items-*`
- `wa-border-radius-*`
- `wa-gap-*`
- Text classes including `wa-heading-*`, `wa-body-*`, and `wa-caption-*`


claviska (approved):
Great work on these! Nice visuals too.

LeaVerou:
I left comments wrt the API [here](https://github.com/shoelace-style/webawesome/issues/165#issuecomment-2356534266).

Looking at the code, one thing I noticed is: the way the modifiers are implemented, you can only have one per element. Which is currently ok, because it doesn't make sense to combine these classes on the same element. But it made me wonder: do we actually _want_ these as modifiers? What if they were utility classes as well? E.g. `class="wa-stack wa-align-right wa-gap-xl"`? Then users can compose them however they want. We could even design them so that they can be set on an ancestor and set the gap or alignment across multiple layout utilities.

lindsaym-fa:
Bringing this out of 'draft' status and would love another set of eyes before we merge. @LeaVerou or @claviska, any blocking feedback or suggestions?

claviska:
These looks good to me. Feel free to merge!

@kelseythejackson do you think these will help reduce styles/markup in patterns?

claviska (approved):
Looks great! Nice examples in the style utilities docs too.

PR #182, opened 2024-10-03
Kj/app patterns

kelseythejackson:


claviska:
Is there an index page to see what patterns are available so far, or should I just dive into the file structure for the URLs?

kelseythejackson:
> Is there an index page to see what patterns are available so far, or should I just dive into the file structure for the URLs?

I haven't made an index yet, all the link should be in the side bar though

PR #186, opened 2024-10-24
fix progress ring in safari

KonnorRogers:
Fixes https://github.com/shoelace-style/webawesome-alpha/issues/89

### DIFF
--- a/src/components/progress-ring/progress-ring.styles.ts
+++ b/src/components/progress-ring/progress-ring.styles.ts
@@ -3,7 +3,7 @@ import { css } from 'lit';
 export default css`
   :host {
     --size: 8rem;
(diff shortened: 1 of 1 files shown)


src/components/progress-ring/progress-ring.styles.ts
```diff
@@ -3,7 +3,7 @@ import { css } from 'lit';
 export default css`
   :host {
     --size: 8rem;
-    --track-width: 0.25rem;
+    --track-width: 0.25em;
```
claviska:
```suggestion
    --track-width: 0.25em; // avoid using rems here - https://github.com/shoelace-style/webawesome-alpha/issues/89
```

claviska (approved):

PR #188, opened 2024-10-28
fix labels when updated dynamically

claviska:
Context: https://github.com/shoelace-style/shoelace/issues/1971

@KonnorRogers please check my work on this one. Tests are passing and the labels update. No need for a mutation observer.

If we're OK with this, I can backport it to SL.


KonnorRogers:
Looks good to me! Even better that there's no MO needed!

PR #240, opened 2024-11-22
Indices of components by attribute, slot, part, etc.

LeaVerou:
Adds a `docs/reference` route (better naming suggestions welcome!) that displays tables of components by  attribute, slot, part, etc. in descending order of component count. I needed it myself and figured it could be useful more broadly.

image

Also cleans up the code that exposes component metadata to 11ty:
- Exposes it as [global data](https://www.11ty.dev/docs/data-global/) that is there for every page rather than a function call
- Exposes as an object literal (`components`), an array (`componentList`) and by thing (e.g. `componentsBy.attribute`)


claviska:
Neat idea! I'd suggest adding a description at the top of the page so users know what it is. And maybe a link in the ever-growing sidebar (which we can better organize in a later PR).

I really like moving the insane (it really is) logic into componentList.js! One nit: can we use kebab-case for filenames? I don't think we have any other camelcase names in the lib at the moment.

LeaVerou:
> Neat idea! I'd suggest adding a description at the top of the page so users know what it is. And maybe a link in the ever-growing sidebar (which we can better organize in a later PR).

Yeah, that's the plan! Or an icon, once #241 is merged. Just wanted to get your thoughts first.

> I really like moving the insane (it really is) logic into componentList.js! One nit: can we use kebab-case for filenames? I don't think we have any other camelcase names in the lib at the moment.

I also tend to use kebab-case for filenames, but anything in `_data` becomes a global JS variable by 11ty so not sure how that would work. But maybe 11ty is smart enough to do the conversion. @zachleat?

lindsaym-fa:
Nice! Curiosity question, if you have the bandwidth: what's your use case for this page, @LeaVerou?

claviska:
> I also tend to use kebab-case for filenames, but anything in _data becomes a global JS variable by 11ty so not sure how that would work. But maybe 11ty is smart enough to do the conversion. @zachleat?

Oh, right. I try to avoid that pattern because it feels a bit too magical 😆 

We currently assign global data here: https://github.com/shoelace-style/webawesome/blob/4522b408793a05142719cd5d5091c3d6bc897217/docs/.eleventy.js#L30

Aside: something I love about 11ty is how you can do things in different ways, but somethind I dislike about 11ty is how you can do ALL OF THESE THINGS by default, which encourages mixing paradigms. Maybe that's a good default, but it would be nice to opt out of certain features (and maybe show a warning when you try to use it) in the config.

This would help keep things consistent (read: sane) in projects where multiple people are working on them.

11ty: "It looks like you're adding global data using `[filename.ext]`, but this project has been configured to disallow that. Try adding data using [permitted ways] instead."

LeaVerou:
@lindsaym-fa 
> Nice! Curiosity question, if you have the bandwidth: what's your use case for this page, @LeaVerou?

My _own_ use case was extremely niche: to help with the https://github.com/shoelace-style/webawesome/labels/ux%20audit . But I thought there are many more. E.g. for a WA user to quickly see at a glance what other elements support a given pattern, or for us (and contributors) to see what components are affected by something (e.g. @claviska is working on removing the `base` part, this would have shown him which components are affected at a glance). Using the browser's search in page feature would also be useful to find a piece of syntax (e.g. I often use MDN's syntax references in that way).

@claviska 
> > I also tend to use kebab-case for filenames, but anything in _data becomes a global JS variable by 11ty so not sure how that would work. But maybe 11ty is smart enough to do the conversion. @zachleat?
> 
> Oh, right. I try to avoid that pattern because it feels a bit too magical 😆
> 
> We currently assign global data here:
> 
> https://github.com/shoelace-style/webawesome/blob/4522b408793a05142719cd5d5091c3d6bc897217/docs/.eleventy.js#L30

This is a useful pattern for certain things (e.g. it's the only way plugins can add global data AFAIK) and I've used it extensively, but I don't think it's a good idea for managing ALL data. Putting data in `_data`  [^1] means that it lives in a centralized place where all contributors who know 11ty can find it (and even for those who are not familiar with 11ty, it's quite self-explanatory).

Btw I asked @zachleat on Slack and he confirmed that no kebab-case filenames would not automatically convert. See https://github.com/11ty/eleventy/issues/567

[^1]: or whatever else you want to call it, there's a config option and I often change it to `data` so that it's deployed and accessible on the Web, which doubles as a makeshift "API" of sorts.

> Aside: something I love about 11ty is how you can do things in different ways, but something I dislike about 11ty is how you can do ALL OF THESE THINGS by default, which encourages mixing paradigms. 

Haha I’m afraid that’s exactly what I like about 11ty: that you get all these things for free. It's very much designed in the spirit of making simple things easy and complex things possible, with a very smooth curve between the two. While that provides tangible usability benefits, the downside you are describing sounds a lot like — wait for it — theoretical purity 😁😁😁

claviska:
> theoretical purity 😁😁😁

you spelled ease of maintainability wrong 😝

zachleat:
Filed this one after reading this feedback: https://github.com/11ty/eleventy/issues/3554

docs/docs/components/reference.njk
```diff
@@ -0,0 +1,34 @@
...
+title: Component Reference
+layout: docs
+---
+
+<style>
+table code {
+	white-space: nowrap;
+}
+</style>
+
```
LeaVerou, line 11:
```suggestion

<p>WA components tend to reuse patterns so that you can port knowledge from one to others. While in individual component pages only the syntax of that component is listed, here you can find components grouped by the syntax they implement, so you can at a glance see which other components share each pattern (or just find something!).
```

@claviska is that the kind of description you had in mind?

LeaVerou:
- I took a stab at a description, posted as a suggestion for now.
- I would prefer we resolve on #241 (and figure out if there's a better URL for it) before I add it to the sidebar, but merging it means we can already use it by manually going to its URL

So, what are the blockers for merging?

LeaVerou:
@claviska said to go ahead with merging and we can figure out the IA later

PR #241, opened 2024-11-22
Make sidebar headings links whenever an overview is available

LeaVerou:
This has bugged me for a while — I keep clicking on the headings expecting some kind of index or overview, but nothing happens. Integrating the overviews into the headings of each category creates a consistent association for where the overview is AND saves space.

I didn't do anything to style these better so the styling leaves a lot to be desired, but we can fix that later:

image

### DIFF
--- a/docs/_includes/sidebar.njk
+++ b/docs/_includes/sidebar.njk
@@ -20,11 +20,8 @@
 </ul>
 
 {# Components #}
-<h2>Components</h2>
+<h2><a href="/docs/components/">Components</a></h2>
 <ul>
-  <li>
-    <a href="/docs/components/">Components Overview</a>
-  </li>
   <li>
     <a href="/docs/components/animated-image">Animated Image</a>
   </li>
@@ -216,9 +213,8 @@
 </ul>
 
 {# Layout #}
-<h2>Layout</h2>
+<h2><a href="/docs/layout">Layout</a></h2>
 <ul>
-  <li><a href="/docs/layout">Layout Overview</a></li>
   <li><a href="/docs/components/page">Page</a></li>
   <li><a href="/docs/layout/cluster">Cluster</a></li>
   <li><a href="/docs/layout/flank">Flank</a></li>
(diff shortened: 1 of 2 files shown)


claviska:
Good call. My only concern is the visual inconsistency (which is necessary) and the behavioral inconsistency (which we probably don't need to solve by adding indexes for "Getting Started", "Resources", etc.

CleanShot 2024-11-22 at 08 11 09@2x

Would adding an index of some sort as the first item feel better?

CleanShot 2024-11-22 at 08 12 55@2x

Or maybe with an icon to help it stand out?

CleanShot 2024-11-22 at 08 15 01@2x

claviska:
oh, I've just proposed what we already have 🙃 

CleanShot 2024-11-22 at 08 19 24@2x

hmm

LeaVerou:
Yup, you did just suggest exactly what I was trying to change 😅

- I think we want to avoid dead clicks. Without a user test I can't know how many users are like me and kept clicking on the heading, but I'm pretty sure it's nonzero (and for those that don't already expect to find a link there there is no downside assuming we make it clear that one exists). 
- Having "Components" (etc) be the link to the index of components is pretty natural and closely mirrors RESTful URLs too. In some ways it's similar to the convention of a website logo linking to the homepage.
- Having it on the same level as any other component makes it look like a component. And we'd want components to eventually have icons, so adding an icon is not a long term solution.

WRT visuals, I was hoping for @lindsaym-fa's input on what it could look like, I hadn't dealt with styling at all yet. It should definitely be more subtle.

claviska:
Visuals could indeed solve it, which is 50% of my concern. The other 50% is around other groups _not_ having an index page, and it probably doesn't make sense to add one to each _just_ for consistency.

lindsaym-fa:
> The other 50% is around other groups _not_ having an index page, and it probably doesn't make sense to add one to each _just_ for consistency.

I'd argue that **_is_** actually the ideal solution, though obviously not one we need to drop everything for. 😄 _Not_ including click signifiers on items in a sidebar like this doesn't do anything to communicate that a term isn't clickable since the links intentionally look like text and are contextually recognized as links (styling everything in a sidebar to look clickable makes the sidebar terribly noisy/impedes scanning). It's no surprise that Lea and likely many others have tried to click the plain-text headers, so the best solution would be to match user expectations by ensuring every header links to an index page.

For now, I wonder if either using an icon (like @claviska suggested above but on the header itself) or keeping the link decoration while using a normal text color would be effective.

I can implement a styling change. I'm 100% happy with the change otherwise!

LeaVerou:
Alright, how about this:

image

Hovered:
image

lindsaym-fa:
I think that works well! Sorry that this fell off of my radar. 😖

LeaVerou:
@claviska any objections?

lindsaym-fa (approved):

LeaVerou:
Going once, going twice, …merged.

PR #242, opened 2024-11-22
Display attribute and property in separate columns

LeaVerou:
Before:
image

After:
image

Rationale:
- It wasn't super clear that the second identifier was an attribute
- There was no attribute section, so those looking for it couldn't find it

I did try to just use `colspan` when property and attribute where the same, but because of the left alignment, it just looked like the attribute is missing, so I ended up showing them even when they're the same.

(This is probably for another issue, but what does reflect mean? Reflects *from* an attribute *to* an attribute, or both?)

### DIFF
--- a/docs/_layouts/component.njk
+++ b/docs/_layouts/component.njk
@@ -62,13 +62,14 @@
 
   {# Properties #}
   {% if component.properties.length %}
-    <h2>Properties</h2>
+    <h2>Properties & Attributes</h2>
 
     <div class="table-scroll">
       <table class="component-table">
         <thead>
           <tr>
-            <th class="table-name">Name</th>
+            <th class="table-name">Property</th>
+            <th class="table-name">Attribute</th>
             <th class="table-description">Description</th>
             <th class="table-reflects">Reflects</th>
           </tr>
@@ -78,8 +79,10 @@
             <tr>
               <td class="table-name">
                 <code>{{ prop.name }}</code><br>
+              </td>
+              <td class="table-name">
                 {% if prop.attribute %}
-                  <div><small><code>{{ prop.attribute }}</code></small></div>
+                <code>{{ prop.attribute }}</code>
                 {% endif %}
               </td>
               <td class="table-description">


claviska:
I've passively struggled with this for so long because the extra column is redundant more than 50% of the time (maybe closer to 80% tbh) and leads to things like this (from the dropdown page).

CleanShot 2024-11-22 at 08 02 49@2x

And on mobile, the story is even worse.

CleanShot 2024-11-22 at 08 03 44@2x

In, uh, other explorations I've settled on combining them and only showing the attribute when it's actually different.

CleanShot 2024-11-22 at 08 04 30@2x

LeaVerou:
> In, uh, other explorations I've settled on combining them and only showing the attribute when it's actually different.

I like that! But then how do you communicate that no attribute exists for JS-only props?

claviska:
here's the best i got 😆 

CleanShot 2024-11-22 at 10 04 39@2x

LeaVerou:
I don't get it. Which one of these has no attribute?

LeaVerou:
What about icons? We could us a [js](https://fontawesome.com/icons/js?f=brands&s=solid) or [square-js](https://fontawesome.com/icons/square-js?f=brands&s=solid) icon for prop and an [at](https://fontawesome.com/icons/at?f=classic&s=solid) for attribute. Then those that share the same name would use both icons, to indicate this. 

Thoughts? I worry it could get noisy…

claviska:
> I don't get it. Which one of these has no attribute?

Both of the properties shown in the screenshot above have attribute _and_ properties. There are only a few properties that don't have attributes, but they exist. For example, dropdown has `containingElement` which references to an HTML element. Popup has `flipBoundary`, `shiftBoundary`, etc. which also reference HTML elements.

Basically, anything that requires non-scalar data (which we do try to avoid).

LeaVerou:
Right. So looking at this, there is nothing there actually indicating that e.g. `containingElement` does not correspond to an attribute. The Reflects column is empty, but that could just mean that it doesn't reflect, not necessarily that it's not an attribute. Perhaps that's not even possible as a pattern with Lit, but we shouldn't expect that people understand the Lit to parse our docs (I'm not even sure all our target users understand what "Reflects" means).

LeaVerou:
Closing this, I’ll just change the heading to Properties & Attributes if you have no objections @claviska (which is small enough to do as a direct commit). That would address the issue of not being able to find where the attributes are, without the issues that two columns would introduce.

claviska:
Works for me. We can change this anytime!

LeaVerou:
Done!

Issue #206, opened 2024-04-22
Remove `@watch()` decorator


justinfagnani:
After discussing this with @KonnorRogers, we feel it best to remove the `@watch` decorator and use Lit's `updated()` lifecycle method instead of being clever. This change should be somewhat trivial to make in WA.

The original bug regarding this decorator is posted below for historical purposes.

---

### Describe the bug

`@watch()` currently patches the element prototype to override `update()` each time it's applied. Patching the prototype at all is less-then-ideal for performance, but creating a potentially long chain of calls will be worse. As decorators migrate towards standard decorators such patching will be impossible anyway, and there will be no other patching from experimental decorators, potentially making this patch (if it can be done) the only thing to change the class shape, slowing class initialization.

To fix this, I'd recommend making `@watch()` cooperate with a base class that allows watchers. The directive stores the information about the watcher in a cache, and the base class looks up the watchers to invoke them.

Here's some code from a similar utility I have:

```ts
type Observer = {
  method: (...args: unknown[]) => void;
  keys: Array<PropertyKey>;
};

const observersForClass = new WeakMap<
  DecoratorMetadataObject,
  Array<Observer>
>();

export class BaseElement extends ReactiveElement {
  override update(changedProperties: PropertyValues) {
    const meta = (this.constructor as typeof ElematicElement)[Symbol.metadata];
    const observers = observersForClass.get(meta);
    if (observers !== undefined) {
      for (const {keys, method} of observers) {
        if (keys.some((p) => changedProperties.has(p))) {
          const oldValues = keys.map((p) => changedProperties.get(p));
          method.apply(this, oldValues);
        }
      }
    }
    super.update(changedProperties);
  }
}

export function observe<T extends BaseElement>(...keys: Array<keyof T>) {
  return <C, V extends (this: C, ...args: any) => any>(
    method: V,
    context: ClassMethodDecoratorContext<C, V>,
  ) => {
    let observers = observersForClass.get(context.metadata);
    if (observers === undefined) {
      observersForClass.set(context.metadata, (observers = []));
    }
    observers.push({keys, method});
  };
}
```

This is a standard decorator, but it can be adapted to experimental decorators by using the class as the key into the `observersForClass` map.

One benefit of this approach is that the watcher run in definition order, which is probably more intuitive. Right now watchers run in reverse definition order because the later methods make the outer wrappers for `update()`.

[...]

PR #243, opened 2024-11-22
[DRAFT] Watch removal

claviska:
This PR removes the `@watch` decorator and alleviates any performance concerns in #206. It also sneaks in vertical orientation in `<wa-button-group>`.

I've noticed only three things that still need to be fixed, and they all seem related.

- [ ] Dropdown - opens and instantly closes
- [ ] Tooltip - opens and instantly closes
- [ ] Select - freezes the page due to recursion (probably from `open`)


LeaVerou:
This PR seems to have a bunch of things that are completely unrelated to the title. Is that intentional?

claviska:
whoops, yep, I should have split out vertical orientation in button group. Since it's here already, I've updated the description above.

The pattern page was a prettier error and then I also made some indentation fixes.

claviska:
@KonnorRogers after digging, it looks like dropdown and tooltip are related to SSR. Disabling SSR makes them work. 🙈 

Select appears to be unrelated. Working on that one now.

claviska:
Update: I've fixed the infinite loop in select and it _is_ also affected by SSR.

claviska:
Thanks for the fixes, @KonnorRogers! I'm going to sort out the tests and get this one done tomorrow.

claviska:
Ugh, abandoning this because tests are going crazy and it's become a rabbit hole. This was supposed to be a quick update, but it's not a priority for launch.

PR #247, opened 2024-11-22
automatically slot tab into tab group; fixes #239

### LINKED ISSUE #239: [Tab] Auto-slot into the `tab` slot
As discussed today, `<wa-tab>` only has one possible slot it can ever go into so there’s no reason for users to need to specify it.

As @KonnorRogers pointed put, this is easier than other types of auto-slotting (#197) because no logic is involved, there is only one possible slot.

claviska:
Most of the changes are removing the slot attribute from the docs. The [big change](https://github.com/shoelace-style/webawesome/compare/autoslot?expand=1#diff-91d78201c2c39ef129491fea401203d95518fae0195ff91c9ee397a89d425dceR29) here is in `tab.ts`:

```diff
+  public slot = 'nav';
```

This makes the tab component "self-slotting" so users can do this:

```html
<wa-tab-group>
  <wa-tab panel="general">General</wa-tab>
  <wa-tab panel="custom">Custom</wa-tab>
  <wa-tab panel="advanced">Advanced</wa-tab>

  <wa-tab-panel name="general">This is the general tab panel.</wa-tab-panel>
  <wa-tab-panel name="custom">This is the custom tab panel.</wa-tab-panel>
  <wa-tab-panel name="advanced">This is the advanced tab panel.</wa-tab-panel>
</wa-tab-group>
```

Instead of this:

```html
<wa-tab-group>
  <wa-tab slot="nav" panel="general">General</wa-tab>
  <wa-tab slot="nav" panel="custom">Custom</wa-tab>
  <wa-tab slot="nav" panel="advanced">Advanced</wa-tab>

  <wa-tab-panel name="general">This is the general tab panel.</wa-tab-panel>
  <wa-tab-panel name="custom">This is the custom tab panel.</wa-tab-panel>
  <wa-tab-panel name="advanced">This is the advanced tab panel.</wa-tab-panel>
</wa-tab-group>
```

This PR will establish a pattern for future elements where slots can be assumed based on their position in another component.

### DIFF
--- a/src/components/tab/tab.ts
+++ b/src/components/tab/tab.ts
@@ -26,7 +26,7 @@ let id = 0;
 @customElement('wa-tab')
 export default class WaTab extends WebAwesomeElement {
   static styles: CSSResultGroup = [componentStyles, styles];
-  public slot = 'nav';
+  public slot = 'nav'; // Auto-slot into nav slot
 
   private readonly attrId = ++id;
   private readonly componentId = `wa-tab-${this.attrId}`;
--- a/src/components/tab/tab.ts
+++ b/src/components/tab/tab.ts
@@ -26,6 +26,7 @@ let id = 0;
 @customElement('wa-tab')
 export default class WaTab extends WebAwesomeElement {
   static styles: CSSResultGroup = [componentStyles, styles];
+  public slot = 'nav';
 
   private readonly attrId = ++id;
   private readonly componentId = `wa-tab-${this.attrId}`;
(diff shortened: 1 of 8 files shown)


src/components/tab/tab.ts
```diff
@@ -26,6 +26,7 @@ let id = 0;
 @customElement('wa-tab')
 export default class WaTab extends WebAwesomeElement {
   static styles: CSSResultGroup = [componentStyles, styles];
+  public slot = 'nav';
```
LeaVerou:
```suggestion
  public slot = 'nav'; // Auto-slot into nav slot
```

LeaVerou (approved):
LGTM!

PR #264, opened 2024-12-03
Leave existing ids intact when generating anchor headings

claviska:
If an id is explicitly set on a heading, it will be left intact now.

### DIFF
--- a/docs/_utils/anchor-headings.js
+++ b/docs/_utils/anchor-headings.js
@@ -53,7 +53,7 @@ export function anchorHeadingsPlugin(options = {}) {
         if (!id) {
           const slug = createId(clone.textContent ?? '') ?? uuid().slice(-12);
           id = slug;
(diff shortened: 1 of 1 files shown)


docs/_utils/anchor-headings.js
```diff
@@ -39,22 +39,26 @@ export function anchorHeadingsPlugin(options = {}) {
...
+          return;
         }
 
-        if (hasAnchor || !id) {
-          return;
+        let id = existingId;
+        if (!id) {
+          const slug = createId(clone.textContent ?? '') ?? uuid().slice(-12);
+          id = slug;
+          let suffix = 0;
```
LeaVerou:
```suggestion
          let suffix = 1;
```

Unrelated to the current change, but I think `thing-2` makes a lot more sense for naming a duplicate than `thing-1`.

PR #272, opened 2024-12-05
set outline false on page, add pattern.njk

KonnorRogers:
Patterns now use their own layout with `outline` set to false.


claviska:
This fixes this.

CleanShot 2024-12-05 at 13 37 44@2x

claviska (approved):

PR #279, opened 2024-12-06
Page spacing

### LINKED ISSUE #245: [Page] Default spacing/page widths
https://webawesome-git-next-font-awesome.vercel.app/docs/components/page#how-to-apply-spacing-to-your-layout

> The layout component does not apply spacing for you. You can apply the appropriate paddings or margins directly to the elements you slot in to fine tune your spacing needs.
> When using <wa-page>, make sure to zero out all paddings and margins on <html> and <body>, otherwise you may see unexpected gaps. 

As discussed yesterday with @claviska and @KonnorRogers, ideally we want `<wa-page>` to add nice spacing etc so that simply using it immediately yields benefits, and users can always override them. We'd need @lindsaym-fa's help about what these defaults should look like.

This would also mean adding something like this to Applied CSS (which we currently recommend people add manually):

```css
:is(html, body):has(wa-page) {
  min-height: 100%;
  height: 100%;
  padding: 0;
  margin: 0;
}
```

I’d be inclined to tag this high priority since it _really_ diminishes the onboarding experience for `<wa-page>` but if you folks disagree feel free to change it!

lindsaym-fa:
This adds default `flex`, `padding`, and `gap` styles for each slot in `<wa-page>`.

[See the default styles in action](https://webawesome-git-page-spacing-font-awesome.vercel.app/docs/experimental/default-page-spacing). Here's an image preview:
Screenshot 2024-12-06 at 5 03 58 PM

I've also added a demo page that shows off a documentation page example that accepts some of the new default styles and overrides others. It still needs a little TLC, but [you can find it in a new `/components/page-samples/` folder](https://webawesome-git-page-spacing-font-awesome.vercel.app/docs/components/page-samples/documentation).


src/themes/applied.css
```diff
@@ -9,6 +9,13 @@ html {
   margin: 0;
 }
 
+:is(html, body):has(wa-page) {
+  min-height: 100%;
+  height: 100%;
+  padding: 0;
+  margin: 0;
+}
+
```
KonnorRogers, line 18: smart!

src/components/page/page.styles.ts
```diff
@@ -14,6 +15,82 @@ export default css`
...
+          [slot='navigation-header'],
+          [slot='navigation'],
+          [slot='navigation-footer'],
+          [slot='menu'],
+          [slot='main-header'],
+          [slot='main-footer'],
+          [slot='aside'],
+          [slot='footer']
+        )
+    ) {
```
KonnorRogers, line 32: TIL `::slotted()` supports `:is()` dont think I ever thought to try that 👀

docs/docs/experimental/default-page-spacing.md
```diff
@@ -0,0 +1,100 @@
...
+
+<style>
+  [slot='banner'] {
+    background-color: pink;
+  }
+  [slot='header'] {
+    background-color: peachpuff;
+  }
+  [slot='subheader'] {
+    background-color: papayawhip;
```
KonnorRogers, line 15: Found myself a new favorite CSS color name

KonnorRogers (approved):
looks good to me! Thanks for making this! Much nicer! and by using `::slotted()` makes it easy for users to opt out of the defaults.

Issue #270, opened 2024-12-04
Support isolated code examples that are rendered in an `<iframe>`


LeaVerou:
This is needed for #257 as we can't have `<wa-page>` demos that just live in a `<div>`, `<wa-page>` is intended for whole pages.

We could use a class name flag for this, e.g. `.isolated`.

Many but not all isolated code examples would also need zooming (e.g. without that all `<wa-page>` demos would render a mobile view!). We could do that with an attribute (e.g. `data-zoom="2"`) or a CSS property (`style="--zoom: 2"`).

The way I proposed handling this is:
- We use `<iframe srcdoc>` to render the example. That way we don't need to build a separate page.
- If zooming is needed, we render a larger iframe, and then use transforms to scale it down (and clipping)
- @claviska proposed also using `inert` on these to prevent them from trapping focus. I wonder if `tabindex="-1"` might be a better solution so that they can remain interactive.

KonnorRogers:
Things to note when I did this while working on `<light-preview>` , if the `<iframe>` is meant to resize based on content, we need to add a ResizeObserver within the `<iframe>` and tell the host page about the resize to resize the code container. If not, we'll generally want a static `height` for the iframes.

The other thing is obviously the extra boilerplate around making sure to import the global stylesheet, and import the autoloader. I can't remember if there was any issues around content escaping for `srcdoc`, but being an attribute im sure they exist.

LeaVerou:
Closed via #281

Issue #167, opened 2024-09-03
Bring back resizable code examples


lindsaym-fa:
Our code examples would benefit from being resizable to highlight each component's responsiveness, and the [`wa-resize-observer`](https://backers.webawesome.com/docs/components/resize-observer) example explicitly relies on it. Let's bring back the resizable code examples we enjoyed with Shoelace. 😊

claviska:
That's fair. Would it be better to use preset buttons for mobile/tablet/desktop or do you want it to behave the same as before?

lindsaym-fa:
Free resizing has my vote. The flexibility would better satisfy folk's curiosity, I think.

LeaVerou:
@lindsaym-fa do you think this should be an opt-in or an opt-out?

lindsaym-fa:
I'd argue for opt-out. I feel like I see this feature on code examples more often than not.

LeaVerou:
Agreed — that’s how I've implemented it so far. In fact, the opt-out is not yet implemented 😁

LeaVerou:
Closed via #281

PR #281, opened 2024-12-06
Add `<wa-code-demo>` and `<wa-viewport-demo>` and use them by default

LeaVerou:
This adds a `<wa-code-demo>` component, that componentizes our existing code snippets and adds certain features.

Things done:
- [x] Ported existing code example functionality (except `no-edit` and `no-buttons` but these are a) trivial to add b) can be done already by doing `::part(edit button) { display: none }`
- [x] #270  
- [x] Viewport emulation
- [x] #167 
- [x] Custom slotted-in previews, to account for code transforms, whitespace trimming etc.
- [x] Syntax to include resources (scripts, CSS, etc.).
- [x] [**Docs!**](https://webawesome-git-code-demo-font-awesome.vercel.app/docs/components/code-demo)

Things remaining:
- [ ] Better documentation for the CEM
- [ ] Tests

Once this is in good shape, I could do one of the following, in a separate PR:
1. Move all code examples to use this instead, by editing `docs/_utils/code-examples.js` to spit out uses of this component and delete `docs/assets/scripts/code-examples.js` as it will be no longer needed.
2. Take a more conservative approach and introduce a modifier class to opt-in to the new rendering, so that @lindsaym-fa can try it out on the `<wa-page>` docs without risking making potentially breaking changes right before a release.
3. Or something in the middle: Instead of a modifier class, we could introduce an opt-in on a page by page basis.


claviska:
CleanShot 2024-12-09 at 12 23 51@2x

Should we add a reasonable min-width to prevent this?

LeaVerou:
Absolutely. I thought I had one already, will check once I'm done with the Safari bug

claviska:
CleanShot 2024-12-09 at 12 28 29@2x

I agree on the spacing issue here. Where do you think the best place to solve it it?

- `docs.css` will fix it for us, but nobody else
- `applied.css` will fix it for some users, but adds WA-specific selectors which we don't currently do (this could easily bloat the stylesheet too)

Maybe it should exist directly in callout's styles, but the selector might be nasty unless we only want to scope it to certain tags. (We can't reliably use `:first-child` or similar because of the structure — the slotted icon will be targeted instead.)

CleanShot 2024-12-09 at 12 31 07@2x

claviska:
Opening a code example is instant, but closing it has a subtle animation. Is that intentional?

LeaVerou:
> Opening a code example is instant, but closing it has a subtle animation. Is that intentional?

The intention was to have a subtle animation both ways, but I couldn't get it to work and decided the subtle animation while closing is better than no animation (the animation when opening should be faster anyway).

LeaVerou:
Re:callout styles, I can take a look once this is done.

src/components/code-demo/code-demo.styles.ts
```diff
@@ -0,0 +1,199 @@
...
+    #source,
+    [part~='toggle'] wa-icon {
+      transition-duration: var(--code-expand-duration);
+    }
+
+    #source {
+      /* Expanded */
+      height: auto;
+      display: block;
+    }
```
claviska, line 117: This transitions the height on expand:

```suggestion
    #source {
      /* Expanded */
      height: auto;
      display: block;
    }

    @starting-style {
      #source {
        height: 0px;
        display: block;
      }
    }
```

claviska (approved):
Let some comments and a suggestion to fix the opening transition. Looks good!

src/components/code-demo/code-demo.styles.ts
LeaVerou: Have you tried it? I'd swear it was among the many things I tried that didn't work…

claviska: Yes, it works in Chrome. My first go was a fail because I forgot to add the `#source` selector inside `@starting-style` but it works with the above.

PR #291, opened 2024-12-10
Fix FOUC for default theme

KonnorRogers:
There is still some minor layout shift if you're not using the default theme, but the default theme has no shifting. Will create an issue to fix it.


claviska (approved):

ricky11:
I am pre-rendering my single page app, and i still see layout shift right before the JS kicks in.. I am also using the fouc utility class class="wa-cloak" on the navigation <wa-popover>.. The shift does not happen in SPA mode ofcourse, but does happen when i pre-render.  On page load i see the popover menu details flash, and then hide.

PR #296, opened 2024-12-11
Fix turbo caching on script tag

KonnorRogers:
While this one feels weird to have the cache busting, its much closer to the expected behavior.

The other version [here](https://github.com/shoelace-style/webawesome/pull/295) with the event listener + `turbo:load` will fire on every page.

### DIFF
--- a/docs/_includes/base.njk
+++ b/docs/_includes/base.njk
@@ -6,7 +6,6 @@
     <meta name="description" content="{{ description }}">
     <meta name="theme-color" content="#f36944">
     {% if noindex %}<meta name="robots" content="noindex">{% endif %}
-    <meta name="turbo-cache-control" content="no-preview">
 
     <title>{{ title }}</title>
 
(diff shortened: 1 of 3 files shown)


claviska:
Thanks for looking into this. Why doesn't it work with `data-turbo="false"`, though?

Maybe Turbo does too much. Are there alternatives that are lighter and reliable?

KonnorRogers:
> Thanks for looking into this. Why doesn't it work with `data-turbo="false"`, though?
> 
> Maybe Turbo does too much. Are there alternatives that are lighter and reliable?

data-turbo="false" is for navigation. So when I click on a link with data-turbo="false", it'll behave like a normal browser link without doing the normal `<head>` / `<body>` swapping.

as for alternatives, I'm not really aware of any, I think view transitions should in theory solve our problem, obviously not supported in every browser, but it's kind of just a progressive enhancement really.

claviska (approved):

claviska (approved):

PR #298, opened 2024-12-12
Move component styles to use CSS files

LeaVerou:
As we gear up to release Applied Styles, this paves the way for sharing styles and utilities between components and Applied Styles. 

`@import` not supported currently, but still an improvement over our current CSS-in-JS! We can manage imports in the component itself, which also produces better results with the browser’s own optimizations (that Lit leverages).


claviska (approved):
Reviewed and tested locally and I don't see any obvious issues. The styles are compiled into JS chunks, which I believe is expected. 👍

PR #302, opened 2024-12-13
Use `getComputedStyle()` util in `tabbable.ts`

LeaVerou:
Tiny PR, but didn't want to commit directly because I can't figure out where this code is used


claviska:
@KonnorRogers wasn’t this for focus trapping? Can we remove it now?

KonnorRogers:
@claviska correct. this should be removed iirc. was only used for focus trapping.

LeaVerou:
You mean the entire file?

KonnorRogers:
yup, pretty sure the whole file can be deleted!

LeaVerou:
Okay, I will close this and do that then

PR #303, opened 2024-12-13
Support any CSS length in `mobile-breakpoint`

LeaVerou:


src/components/page/page.ts
```diff
@@ -138,17 +140,21 @@ export default class WaPage extends WebAwesomeElement {
   @property({ attribute: 'nav-open', reflect: true, type: Boolean }) navOpen = false;
 
   /**
-   * At what "px" to hide the "menu" slot and collapse into a hamburger button
+   * At what page width to hide the "menu" slot and collapse into a hamburger button.
+   * Accepts both numbers (interpreted as px) and CSS lengths (e.g. `50em`), which are resolved based on the root element.
```
KonnorRogers:
```suggestion
   * At what page width to hide the "navigation" slot and collapse into a hamburger button.
   * Accepts both numbers (interpreted as px) and CSS lengths (e.g. `50em`), which are resolved based on the root element.
```

Little nitpicky, but in talking with @lindsaym-fa it is a little confusing that the menu isn't technically what collapses, but the "navigation" slot that collapses. (turns from the `<nav>` to a `<dialog>`)

src/internal/css-values.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  if (resolved?.endsWith('px')) {
+    return parseFloat(resolved);
+  }
+
+  return Number(resolved);
+}
+
+export function toLength(px: number | string): string {
+  return Number.isNaN(Number(px)) ? `${px}px` : (px as string);
```
KonnorRogers: is this backwards?? Or am I misunderstanding the logic?

If you pass a `number` you should get back `${number}px` if you pass anything else, we just use the string.

```suggestion
  return Number.isNaN(Number(px)) ? (px as string) : `${px}px`;
```

Am I missing something obvious?

src/internal/css-values.ts
```diff
@@ -0,0 +1,63 @@
...
+  { on = document.documentElement, as = 'length', initialValue = initialValues[as] }: ResolveOptions = {}
+) {
+  const resolver = `--wa-${as}-resolver`;
+
+  if (!window.CSS || !CSS.registerProperty) {
+    return value;
+  }
+
+  if (!definedProperties.has(resolver)) {
+    CSS.registerProperty({
```
KonnorRogers, line 28: Support on this was way larger than I expected 👀 

https://caniuse.com/?search=registerProperty

KonnorRogers (approved):
This is insanely clever! Thanks for making this happen. A couple of questions, otherwise good to go from me!

src/internal/css-values.ts
LeaVerou: Yikes, yes it is. Now I question my testing 🤣

PR #314, opened 2024-12-17
Custom states

claviska:
- Adds the custom states section to the docs for components that have it
- Removes `[data-wa-state]` fallback in favor of the standard `:state()`
- Removes stateful BEM-style parts from checkbox, radio, and tree item, e.g. `control--checked`
- Fixes a handful of misnamed parts
- Hoists `ElementInternals` to `WebAwesomeElement` so all components can use it (previously on `WebAwesomeFormAssociatedElement`)
- Hoists custom states methods to `WebAwesomeElement`
- Updates tests to use the custom states methods

### DIFF
--- a/docs/docs/form-controls.md
+++ b/docs/docs/form-controls.md
@@ -165,12 +165,12 @@ Custom validation can be applied to any form control that supports the `setCusto
 
 Due to the many ways form controls are used, Web Awesome doesn't provide out of the box validation styles for form controls as part of its default theme. Instead, the following attributes will be applied to reflect a control's validity as users interact with it. You can use them to create custom styles for any of the validation states you're interested in.
 
-- `data-wa-required` - the form control is required
-- `data-wa-optional` - the form control is optional
-- `data-wa-invalid` - the form control is invalid
-- `data-wa-valid` - the form control is valid
-- `data-wa-user-invalid` - the form control is invalid and the user has interacted with it
-- `data-wa-user-valid` - the form control is valid and the user has interacted with it
+- `required` - the form control is required
+- `optional` - the form control is optional
+- `invalid` - the form control is invalid
+- `valid` - the form control is valid
+- `user-invalid` - the form control is invalid and the user has interacted with it
+- `user-valid` - the form control is valid and the user has interacted with it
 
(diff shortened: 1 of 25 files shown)


docs/docs/form-controls.md
```diff
@@ -165,12 +165,12 @@ Custom validation can be applied to any form control that supports the `setCusto
 
 Due to the many ways form controls are used, Web Awesome doesn't provide out of the box validation styles for form controls as part of its default theme. Instead, the following attributes will be applied to reflect a control's validity as users interact with it. You can use them to create custom styles for any of the validation states you're interested in.
 
-- `data-wa-required` - the form control is required
-- `data-wa-optional` - the form control is optional
-- `data-wa-invalid` - the form control is invalid
-- `data-wa-valid` - the form control is valid
-- `data-wa-user-invalid` - the form control is invalid and the user has interacted with it
-- `data-wa-user-valid` - the form control is valid and the user has interacted with it
+- `required` - the form control is required
```
LeaVerou, line 168: Are these attributes actually applied or did you mean to edit this to talk about states?

src/internal/webawesome-element.ts
```diff
@@ -6,6 +6,18 @@ import componentStyles from '../styles/shadow/component.css';
 import { CustomErrorValidator } from './validators/custom-error-validator.js';
 
 export default class WebAwesomeElement extends LitElement {
+  constructor() {
+    super();
+
+    try {
+      this.internals = this.attachInternals();
```
LeaVerou, line 13: We probably don’t want to attach internals without anything to do with them, right? This is something you can only do once, so once we do it here, components can no longer call `attachInternals`. Though given we expose it on a public property it may be okay. 🤷🏽‍♀️ 

What if we do it lazily? (by having an accessor that overwrites itself, though no idea how TS deals with this pattern)

src/internal/webawesome-element.ts
```diff
@@ -6,6 +6,18 @@ import componentStyles from '../styles/shadow/component.css';
 import { CustomErrorValidator } from './validators/custom-error-validator.js';
 
 export default class WebAwesomeElement extends LitElement {
+  constructor() {
+    super();
+
+    try {
```
LeaVerou, line 12: Why a `try...catch` and not just feature detection?

LeaVerou (approved):
See comments, though I don't think any of these is a blocker.

docs/docs/form-controls.md
claviska: These are actually applied by our form validation logic.

src/internal/webawesome-element.ts
claviska: Existing code. Fixed in b2353ca9bddb9fc3105132fd9f62ce042758cfed

src/internal/webawesome-element.ts
claviska: Not sure it's worth the complexity. We can always revisit this if it's a problem.

PR #315, opened 2024-12-17
fix component reloading

KonnorRogers:


claviska:
\*grumbles in javascript*

THANK YOU KONNOR

PR #317, opened 2024-12-17
Add search filter to component index page

claviska:
- Adds a much needed search filter to the component index using basic text matching
- Headings collapse when a search filter is present
- Supports keyword stuffing using `data-keywords="these will match too"`
- Light and easily adaptable for when we generate the index programmatically

Preview:

CleanShot 2024-12-17 at 14 41 35@2x

@lindsaym-fa not sure about the full-width search input, so feel free to adjust styles as necessary.


LeaVerou (approved):
Nit: I'd move the script to `/assets/` but otherwise LGTM!

claviska:
> Nit: I'd move the script to /assets/ but otherwise LGTM!

Then we'd have to pull it in on every page 😕

claviska:
#lazyDev — added a quick fix to this PR for `size.css` since I noticed the 404 in the JS console. It was previously:

```css
@import url('shadow/size.css');
```

Now

```css
@import url('utilities/size.css');
```

PR #334, opened 2024-12-18
fixed label ::after

kelseythejackson:


src/styles/native/form-control.css
```diff
@@ -16,4 +17,7 @@ label:has(input:required)::after {
   content: var(--wa-form-control-required-content);
   margin-inline-start: var(--wa-form-control-required-content-offset);
   color: var(--wa-form-control-required-content-color);
+  position: absolute;
+  top: 0;
+  left: 125px;
```
lindsaym-fa: Unfortunately, this won't work for _any_ input with a `required` attribute since the length of the label may vary.

I didn't consider implementations where `<label>`s wrap `<input>`s when I wrote this rule. Our best bet may be to strip it out. We may not be able to effectively support matching the `required` appearance of our components with native elements. :(

kelseythejackson: updated

lindsaym-fa (approved):

lindsaym-fa:
Thanks, @kelseythejackson!

PR #339, opened 2024-12-19
Fix missing `<wa-input>` styles in Firefox and Safari

### LINKED ISSUE #340: Fix missing input styles
Firefox and Safari aren't rendering the styles for `<wa-input>` —
Image

### LINKED ISSUE #345: Input + textarea components don't have border in Safari + Firefox
Image

Image

Image

### LINKED ISSUE #347: <wa-input type="number"> is missing a border
Image

- [ ] Fix in Firefox
- [ ] Fix in Safari

lindsaym-fa:
Fixes issue #340. Also sneaks in a fix for missing disabled `<wa-input>` styles in all  browsers.

### DIFF
--- a/src/styles/native/input.css
+++ b/src/styles/native/input.css
@@ -51,7 +51,7 @@ input:where(:not(
     }
 
     /* Style disabled inputs */
-    &:has(:disabled:only-child),
+    &:has(:disabled),
     &:disabled {
       cursor: not-allowed;
       opacity: 0.5;
--- a/src/styles/native/input.css
+++ b/src/styles/native/input.css
@@ -1,6 +1,6 @@
 .wa-text-field,
 :host,
-textarea,
+textarea:where(:not(:host *, .wa-text-field *)),
 input:where(:not(
     /* Exclude inputs that don't accept text */
     [type='button'],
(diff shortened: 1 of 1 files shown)


src/styles/native/input.css
```diff
@@ -13,31 +13,29 @@ input:where(:not(
     [type='range'],
     [type='reset'],
     [type='submit']
-  )) {
+  )):where(:not(:host *, .wa-text-field *)) {
```
LeaVerou: What was the issue that this fixes? I wonder if there's a more local fix…

lindsaym-fa: I haven't pinpointed the exact issue, but Firefox doesn't like something about this combination:
```css
:host {
  &:where(:not(.wa-text-field *, :host input, :host textarea)) {
    ...
  }
}
```

Eliminating the nesting and removing any individual part of the nested selector didn't produce any better results. Scoping each selector within `:where(:not(...)` to the specific selectors in the parent rule has been the only change I've had any success with so far.

lindsaym-fa: Oh! Alternatively, we could remove `:where()` from that nested rule, so:
```css
.wa-text-field,
:host,
textarea,
input:where(:not(
    /* Exclude inputs that don't accept text */
    [type='button'],
    [type='checkbox'],
    [type='color'],
    [type='file'],
    [type='hidden'],
    [type='image'],
    [type='radio'],
    [type='range'],
    [type='reset'],
    [type='submit']
  )) {
  /* Style native inputs and <wa-input>'s visible container */
  &:not(.wa-text-field *, :host input, :host textarea) {
    ...
  }
}
```

Unsure if that yields any unexpected results. Still testing.

ETA: So far, so good.

lindsaym-fa: Going to merge with the latter option so that we can stop distracting ourselves with missing style issues in testing. We can revisit if needed.

PR #351, opened 2024-12-19
First stab at using the Popover API for PE

### LINKED ISSUE #273: Drop user-facing `hoist`, use Popover API (keep `hoist` code as internal polyfill)
We currently have 5 components that use `hoist`:

- [ ] [`<wa-color-picker>`](https://webawesome-git-next-font-awesome.vercel.app/docs/components/color-picker) (to pass it to its internal `<wa-dropdown>`)
- [ ] [`<wa-copy-button>`](https://webawesome-git-next-font-awesome.vercel.app/docs/components/copy-button) (to pass it to its internal `<wa-tooltip>`)
- [ ] [`<wa-dropdown>`](https://webawesome-git-next-font-awesome.vercel.app/docs/components/dropdown) (to pass it to its internal `<wa-popup>`)
- [ ] [`<wa-select>`](https://webawesome-git-next-font-awesome.vercel.app/docs/components/select) (to pass it to its internal `<wa-popup>`)
- [ ] [`<wa-tooltip>`](https://webawesome-git-next-font-awesome.vercel.app/docs/components/tooltip) (to pass it to its internal `<wa-popup>`)
- [ ] [`<wa-popup>`](https://webawesome-git-next-font-awesome.vercel.app/docs/components/popup#properties) calls it `strategy` but it's the same thing (`hoist` == `strategy=fixed`)

Presumably because of implementation complexity / perf, this is not the default behavior, which introduces subtle bugs that authors may not even notice (or may not know how to fix). For example, this bug in our own docs, that we had to fix with a JS snippet that just adds `hoist` to all selects:

Image

At this point the [Popover API is supported everywhere](https://caniuse.com/mdn-api_htmlelement_popover) and also fixes it properly, as it uses a UA-level primitive that regular CSS does not have access to. Are there any blockers from us using just that, or is it just about doing the work?

LeaVerou:
Fixes #273

~Seems to work great with dropdown and select, but there's some weirdness with hover bridge on the actual popup page:~ (edit: @KonnorRogers fixed that, thank you!)

image

Remaining issues:
- [ ] Some weirdness with scroll 
image
- [ ] Fall back to fixed positioning if popover is not supported
- [ ] Remove `hoist` and `strategy` from the public APIs


LeaVerou:
@claviska is on board for landing this soon, so I just updated this branch from `next` and went through all examples in the [`popup` docs page](http://localhost:4000/docs/components/popup/) to see what regressions remain (comparing with [the one on backers](https://backers.webawesome.com/docs/components/popup)).

The only issue I spotted is that [Flip fallback](http://localhost:4000/docs/components/popup/#flip-fallbacks), [Shift](http://localhost:4000/docs/components/popup/#shift), and [auto-size](http://localhost:4000/docs/components/popup/#auto-size) no longer work for containers, but only for the whole page. Presumably whatever code calculates the parent scroll container is off.

**Note:** An earlier version of this comment reported additional regressions, but it turned out these were just caused by default UA styles on `[popover]`.

LeaVerou:
I did some digging yesterday. This doesn't appear to be an issue in Floating-UI (it [used to be](https://github.com/floating-ui/floating-ui/issues/1842), but it was fixed). In fact, using dev tools to edit one of the tooltips on their homepage into a popover still does flipping correctly, so there must be _something_ in our code.

We do have some custom logic around getting offset parents:

```js
// Use custom positioning logic if the strategy is absolute. Otherwise, fall back to the default logic.
//
// More info: https://github.com/shoelace-style/shoelace/issues/1135
//
const getOffsetParent =
  this.strategy === 'absolute'
    ? (element: Element) => platform.getOffsetParent(element, offsetParent)
    : platform.getOffsetParent;
```

but even if I revert that fix the issue is still there.

This may be relevant too: https://github.com/floating-ui/floating-ui/discussions/2916

I think this is at the point where someone who has worked more on `<wa-popup>` needs to take a look (@claviska said he would try to look today), but it looks like we're almost there, just this fix and it may be ready to ride the trains! 🎉

LeaVerou:
Still investigating, looks like if I manually set the `flipBoundary` etc it works fine. Also, `elementContext: 'reference'` fixes it, though it works slightly differently: the tooltip can still escape the container partially, but I think that might be okay?

Old:

https://github.com/user-attachments/assets/912eec19-b36e-4aaf-ac25-7b0de5571236

New:

https://github.com/user-attachments/assets/4be38d6b-cb2a-4741-8f91-329bdeab50d4

Edit: So `elementContext: 'reference'` fixes it for flip, but not shift. The option is not available there. So it looks like we may need to do our own boundary detection after all, though that's not that hard, as floating-ui includes a `getOverflowAncestors()` helper.

LeaVerou:
Okay, I've managed to fix all three regressions by calculating a default boundary via `getOverflowAncestors(this.anchorEl)` for shift and autosize when no explicit boundary is provided. 
I've ended up switching flip to use this as well, and that way we get the previous behavior, as @claviska pointed out that this would not work well in certain cases.

I think this is ready for review! Some changes to the docs still needed, but that can happen in parallel with checking for regressions.

**Edit**: Docs should be done too!

src/components/popup/popup.css
```diff
@@ -19,6 +19,19 @@
   isolation: isolate;
   max-width: var(--auto-size-available-width, none);
   max-height: var(--auto-size-available-height, none);
+
+  /* Clear UA styles for [popover] */
+  :where(&) {
```
claviska, line 24: TIL

claviska:
Nice work! Looking at `<wa-select>`, we may need to adjust the boundary:

CleanShot 2025-04-09 at 13 35 54@2x

claviska:
One more 🫠 

CleanShot 2025-04-09 at 13 39 48@2x

Strangely, I expected tooltip to also do this but it appears to work fine.

CleanShot 2025-04-09 at 13 41 23@2x

claviska (approved):
Nice job! This looks great, as discussed.

PR #393, opened 2024-12-22
fix missing components and format number

KonnorRogers:
Fixes https://github.com/shoelace-style/webawesome-alpha/issues/119#issue-2754567030

In checking:

https://github.com/shoelace-style/webawesome/commit/7d3f8b175c3efc218490f7562cbfc595cfab19dc#r150650130

This doesnt seem to affect anything because the native checkbox links are overriding the `<wa-checkbox>`.

@LeaVerou Seems like you were looking to do something with the new "tags", so probably needs both tags and then filter out the "component" collection? idk. But changing this back for now so links appear again.

### DIFF
--- a/docs/docs/components/checkbox.md
+++ b/docs/docs/components/checkbox.md
@@ -1,7 +1,7 @@
 ---
 title: Checkbox
 description: Checkboxes allow the user to toggle an option on or off.
-tags: forms
+tags: component
 native: checkbox
 ---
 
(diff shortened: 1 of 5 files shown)


LeaVerou:
> @LeaVerou Seems like you were looking to do something with the new "tags", so probably needs both tags and then filter out the "component" collection? idk. But changing this back for now so links appear again.

Yeah, the bug here was that the listing was using `collection.component` rather than `collection.components` which is an auto-added tag so it shouldn't be affected by human error. I'll push a change shortly. But thanks for fixing the immediate issue!

PR #407, opened 2025-01-02
tweaked the tag component

kelseythejackson:
Small fix to the multiple select component.


docs/docs/components/select.md
```diff
@@ -114,7 +114,7 @@ To allow multiple options to be selected, use the `multiple` attribute. It's a g
 <wa-select label="Select a Few" value="option-1 option-2 option-3" multiple clearable>
   <wa-option value="option-1">Option 1</wa-option>
   <wa-option value="option-2">Option 2</wa-option>
-  <wa-option value="option-3">Option 3</wa-option>
+  <wa-option value="option-3">Really super extra long option that has too many characters in it and just looks silly, and a few more to make it even more obnoxious</wa-option>
```
claviska: Great for testing, but we should probably leave it out of the example.

src/components/select/select.ts
```diff
@@ -257,7 +257,9 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
           removable
           @wa-remove=${(event: WaRemoveEvent) => this.handleTagRemove(event, option)}
         >
-          ${option.getTextLabel()}
+          <span style="white-space: nowrap;text-overflow: ellipsis;overflow: hidden;max-width: 10ch;">
```
claviska: Thanks for fixing this. Can these styles be moved to `select.css`? You might need to add an id or class to reference it, so feel free to look at other selectors to get a gist of the conventions we're using :)

claviska (changes requested):

docs/docs/components/select.md
kelseythejackson: Done

src/components/select/select.ts
kelseythejackson: will do

src/components/select/select.ts
```diff
@@ -257,7 +257,9 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
           removable
           @wa-remove=${(event: WaRemoveEvent) => this.handleTagRemove(event, option)}
         >
-          ${option.getTextLabel()}
+          <span>
+            ${option.getTextLabel()}
+          </span>
```
claviska: Thanks for updating this. The _only_ thing I'm wondering at this point is if we can do the same thing without the extra span. I _think_ applying the styles to `wa-tag::part(content)` might do the trick. 🤔

kelseythejackson: I tried to do it without adding an extra element. Something about the `display` of the element prevented the ellipses from showing

kelseythejackson: were you able to make it work with just using the content part?

kelseythejackson: scratch that, It did work.

claviska (approved):

PR #410, opened 2025-01-02
Fix tests

claviska:
This fixes most of the tests that were broken.

- [ ] @KonnorRogers I struggled with `<wa-tree>` and had to skip a number of tests that should pass. Take a look at `tree.tests.ts` and search for instances of `it.skip`. It almost seems like a timing issue or maybe something to do with SSR, because the component appears to function properly when testing manually.
   - See: https://github.com/shoelace-style/webawesome/commit/89c95f700866e34460b83930c3ac1f4848191498
- [ ] @KonnorRogers I also skipped the first test in `form.test` because it crashes the browser on every run for some reason
   - See: https://github.com/shoelace-style/webawesome/commit/71bf742af1bfd3c6eea4b6ee69150073f4929000

There are a few other skips still remaining, but these are the ones that seem more important and many of the others are tech debt from before, so we'll address that in a separate issue.


claviska:
sigh

works on my machine

CleanShot 2025-01-02 at 17 17 55@2x

PR #426, opened 2025-01-06
Allow customization of outlined styles, fixes #412

### LINKED ISSUE #412: `wa-input`, `wa-textarea`, and `wa-select` don't inherit all `--wa-form-control-*` theme properties
It looks like our input, textarea, and select components don't respect the values set for `--wa-form-control-*` properties in each theme. 

See this comparison of native vs. WA elements with different custom properties set on the element. The first pair in each example has no custom properties set.

FA theme
Default theme

LeaVerou:
Fixes #412


lindsaym-fa:
It looks like `--wa-form-control-value-color` is the only remaining property that isn't working quite right, though why is not immediately clear to me...
Screenshot 2025-01-06 at 4 54 31 PM

All else looks great. Thanks, @LeaVerou!

LeaVerou:
Should be fixed now!

lindsaym-fa (approved):
Looks great on my end! 🎉

Issue #431, opened 2025-01-06
Callouts are inheriting `--background` from cards


claviska:
Reported here: https://github.com/shoelace-style/webawesome-alpha/discussions/131#discussioncomment-11737762

Repro: https://codepen.io/SupremeSimon/pen/MYgOXjB?editors=1000

Version: 3.0.0-alpha.7

Here's a sanity check, for reference (this is how we intend for it to work):

https://codepen.io/claviska/pen/GgKyvGb?editors=0010

LeaVerou:
I believe the fix for #422 will inadvertently fix that too, so I'll prioritize it. That said, there's no reason for `<wa-card>` to use `--background-color`, it seems to be one of those components where `background-color` should just work (though I haven't looked at its internals, so I could be wrong).

PR #432, opened 2025-01-06
[Callout] Add `appearance` & `size`, closes #422

### LINKED ISSUE #422: [Callout] Add `appearance` and `size`
As discussed with @lindsaym-fa

LeaVerou:
Closes #422 and possibly #431 (though I haven't tested that).

### DIFF
--- a/docs/docs/components/callout.md
+++ b/docs/docs/components/callout.md
@@ -58,6 +58,69 @@ Set the `variant` attribute to change the callout's variant.
 </wa-callout>
 ```
 
+### Appearance
+
+Use the `appearance` attribute to change the callout's visual appearance (the default is `outlined filled`).
+
+```html {.example}
+<wa-callout variant="brand" appearance="outlined accent">
+  <wa-icon slot="icon" name="circle-info" variant="regular"></wa-icon>
+  <strong>This is super informative</strong><br />
+  You can tell by how pretty the callout is.
+</wa-callout>
+
+<br />
+
+<wa-callout variant="success" appearance="accent">
+  <wa-icon slot="icon" name="circle-check" variant="regular"></wa-icon>
+  <strong>Your changes have been saved</strong><br />
+  You can safely exit the app now.
+</wa-callout>
+
+<br />
+
+<wa-callout variant="neutral" appearance="outlined">
+  <wa-icon slot="icon" name="gear" variant="regular"></wa-icon>
+  <strong>Your settings have been updated</strong><br />
+  Settings will take effect on next login.
+</wa-callout>
+
+<br />
+
+<wa-callout variant="warning" appearance="filled">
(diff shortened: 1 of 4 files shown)


src/components/callout/callout.css
```diff
@@ -8,13 +8,17 @@
   align-items: stretch;
   border-radius: var(--wa-panel-border-radius);
   background-color: var(--background-color, var(--wa-color-fill-quiet));
-  border-color: var(--border-color, var(--wa-color-border-quiet));
+  border-color: var(--border-color, transparent);
```
lindsaym-fa, line 11: @LeaVerou Let me know how this change sits with you. A `plain` callout that still had a border felt like a misrepresentation, and I feel that folks would welcome a `filled` callout that is just a subtle background with no border (I've seen this often).

Results:
Screenshot 2025-01-06 at 8 15 39 PM

This also resolves a _very_ small bug where `accent` callouts still had a border.
Screenshot 2025-01-06 at 7 53 18 PM

docs/docs/components/callout.md
```diff
@@ -58,6 +58,71 @@ Set the `variant` attribute to change the callout's variant.
 </wa-callout>
 ```
 
+### Appearance
```
lindsaym-fa, line 61: Because the initial examples were all different `variant`s, I was struggling to understand what was being represented at first. Hoping my changes here help.

lindsaym-fa:
This doesn't quite resolve #431. For callouts with appearances where `--border-color` is not defined (namely anything other than `appearance~="outlined"`), we're inheriting the `--border-color` from the parent card.

The following results show:

1. `appearance="outlined accent"`
2. `appearance="accent"`
3. `appearance="outlined filled"`
4. `appearance="filled"`
5. `appearance="outlined"`
6. `appearance="plain"`

Screenshot 2025-01-06 at 8 32 07 PM

Note all examples without `outlined` are affected.

LeaVerou:
Gotcha. But the fix to #431 was more of a side effect, the primary purpose of this is to close #422 (#433 is a better fix for #431). Do you see any blockers for that?

lindsaym-fa (approved):
100% okay to move forward here.

src/components/callout/callout.css
LeaVerou: Yes, a plain callout should absolutely not have a border (I'm not even sure it should have padding). Interestingly, I thought I had fixed that issue locally but I guess not. Thanks so much! And these examples are much better. Using different variants was an attempt to demonstrate the design space, but not a good one. I guess we need some kind of interactive demo at some point.

LeaVerou:
@lindsaym-fa I wonder if we could use this in the homepage where we have almost all of these styles 😁 

`outlined filled`:
image

`outlined`:
image

`accent`:
image

`filled`:
image

PR #434, opened 2025-01-07
[WIP] CSS properties to change certain attributes via CSS

### LINKED ISSUE #430: Ability to change certain attributes via CSS
This started as a way to set context-dependent defaults for `<wa-icon>` which  would be useful for a bunch of themes, e.g. per @lindsaym-fa it would be good for the Brutalist theme to be able to use FA Sharp Regular instead of the default. 

However, this could be a game-changer for a lot more, e.g. imagine what we could implement if things like `appearance` or `variant` could be set via CSS properties! Probably most of [the presentational attributes](https://github.com/shoelace-style/webawesome/issues/231) would be good candidates, and allow us to have our cake and eat it too.

Since there is a (still undetermined) perf hit from enabling this for a given attribute, the current plan is to make attributes opt-in to being settable via CSS.
 
There is a draft PR open at #447 but is static only.

There are some unknowns we need to figure out first:
- What is the perf hit exactly?
- ~How to monitor changes?~ (edit: Wrote https://observe.style/ as a side project since I opened this issue)

## List of use cases

- Defaults for `<wa-icon>`, either for a whole page or for a specific element (e.g. a danger callout could default to a `circle-exlamation` icon)
- Change icons or icon styles on user interaction (e.g. hover)
- Default color variants for certain states, e.g.:
  - `danger` for `:invalid`
- Themes defining default values for styling attributes, e.g. making all buttons pill or secondary buttons tinted etc.
- Change appearance as a result of user interaction, e.g. a filled outlined button becoming accent on hover
Image
- Set appearance based on a state or attribute, e.g. `checked`:
Image or neutral outlined on disabled

LeaVerou:
- **Static-only** for now, docs excluded from the alpha (intended to be quietly used by themes for now)
- `@property({cssProperty: true | string})` to automatically fall back to reading from a CSS property if the prop is not set (therefore, props with `cssProperty` must not reflect) 
-  Either provide the name of the property or set to `true` for the default `--${tagName}-${name}`
- Don't do any extra work if no property uses `cssProperty` (to minimize perf impact)
- If the JS property is set (`element.foo = "bar"`) it "sticks" and doesn't get overwritten by the CSS property.
- Use in `<wa-icon>`, closes #430 

Look ma, no attributes!
image

TODO:
- [x] Basic support for `--wa-icon-library`, `--wa-icon-variant`, `--wa-icon-family`, `--wa-icon-name`
- [x] Docs
- [x] Do not override explicitly setting the JS property
- [x] Low priority: Create abstraction and move it to base class

Later:
- Observe changes
- `!important` to override attributes
- Unwrap CSS `<string>` values
- Resolve CSS `attr()` values


LeaVerou:
Superseded by #447 (ended up in git hell, recreating was easier)

PR #448, opened 2025-01-08
Fix arrow color on slider tooltips

lindsaym-fa:
Fixes tooltip arrows in `<wa-slider>` using the wrong color.

Before:
Screenshot 2025-01-07 at 7 12 17 PM

After:
Screenshot 2025-01-07 at 7 12 04 PM


claviska (approved):

LeaVerou (approved):
LGTM! Ideally this should not be done with borders at all, there are better ways now. But that’s for the future…

PR #449, opened 2025-01-08
Use theme properties for color picker border

lindsaym-fa:
Updates `<wa-color-picker>` and `<input type='color'>` to use the same theme-derived widths for the trigger's border and whitespace  for better consistency and theme matching.

Before:
Screenshot 2025-01-07 at 7 21 39 PM

After:
Screenshot 2025-01-07 at 7 20 51 PM


LeaVerou (approved):
Thank you! I noticed these too, but wasn't sure what tokens to replace them with.

claviska (approved):

Issue #451, opened 2025-01-08
Fix thick text field borders in Firefox and Safari


lindsaym-fa:
Worst-guys Safari and Firefox have decided the borders for `<wa-input>` and `<wa-textarea>` should be ultra-thick.
<img width="694" alt="Image" src="https://github.com/user-attachments/assets/4b61fd1d-c271-4c34-9b2b-9eade246fc27" />

Affects most themes. Might make sense to fix in #450, which also brings the issue to `<wa-select>`.

PR #458, opened 2025-01-08
Fix #437 regression

### LINKED ISSUE #437: [Regression] `appearance="filled"` on Input, select, textarea now renders as `outlined filled`
Now:

image

[Backers](https://backers.webawesome.com/docs/components/input#appearance):

image

LeaVerou:


src/styles/utilities/appearance.css
```diff
@@ -34,6 +34,10 @@
   --outlined-border-color: var(--wa-color-border-normal);
 
   --text-color: var(--wa-color-on-normal, var(--wa-color-neutral-on-normal));
+
+  :where(&) {
+    --border-color: transparent;
+  }
```
claviska: what is this magic?

LeaVerou: Same selector, 0 specificity.

claviska (approved):
Trusting your judgement, but I'm curious why the selector is necessary. Is it worth a comment to help avoid regressions?

LeaVerou:
Added a comment, though this is literally what `:where()` is for and we use it extensively.

lindsaym-fa (approved):
Thanks, @LeaVerou! ✨

The change didn't take in Safari or Firefox (related: #451). ☹️
Screenshot 2025-01-08 at 11 52 08 AM

I don't think they like something about the combination of `:host()` with nested `:where(&)` because your fix works just fine for native inputs. That said, I've approved this — might need to troubleshoot further in #451.

LeaVerou:
Ah yeah, the stupid Safari bug where `:where(:host)` doesn't match `:host` 🤦🏽‍♀️ Updating and will re-request a review in a couple mins.

claviska:
> Added a comment, though this is literally what :where() is for and we use it extensively.

@LeaVerou So not only does it not have specificity, but it cal also _remove_ specificity? (I'm but on `:has` but haven't realized a use case for `:where` in my travels, so forgive my ignorance.)

Thanks for the comment. It just wasn't obvious to me, so I know it won't be for others.

LeaVerou:
> @LeaVerou So not only does it not have specificity, but it cal also _remove_ specificity? 

Yes! That’s the whole reason it exists 😁
It lets you remove the parts of the selector that just act as a filter from specificity so that specificity becomes meaningful again.
I originally [proposed it](https://lea.verou.me/specs/#where) to kill BEM haha.

LeaVerou:
> Ah yeah, the stupid Safari bug where `:where(:host)` doesn't match `:host` 🤦🏽‍♀️ Updating and will re-request a review in a couple mins.

@claviska btw this is not specific to `:where()`, `:is()` has the same bug.

lindsaym-fa (approved):

PR #460, opened 2025-01-08
Do not reflect `variant` on `<wa-button>`, fixes #454

### LINKED ISSUE #454: Native button inherits variant but `<wa-button>` does not
Not sure which one is the bug, but I'm inclined to think that inheriting variant is the desirable behavior here.

```html
<wa-callout variant="warning">
  <!-- Look ma, no attributes! -->
  <wa-icon slot="icon" name="triangle-exclamation"></wa-icon>
  Here be dragons.
  <button>Button</button>
  <wa-button>Button</wa-button>
</wa-callout>
```

image

Presumably this will be fixed if we simply set `variant` to not reflect. Then we can get a clear picture of the author intent.

LeaVerou:
Fixes #454 

Also a docs fix on callout (used it to make my testing easier)

### DIFF
--- a/docs/docs/components/callout.md
+++ b/docs/docs/components/callout.md
@@ -19,43 +19,40 @@ icon: callout
 Set the `variant` attribute to change the callout's variant.
 
 ```html {.example}
-<wa-callout variant="brand">
-  <wa-icon slot="icon" name="circle-info" variant="regular"></wa-icon>
-  <strong>This is super informative</strong><br />
-  You can tell by how pretty the callout is.
-</wa-callout>
-
-<br />
-
-<wa-callout variant="success">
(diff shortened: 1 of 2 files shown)


LeaVerou:
@claviska I see some tests are now failing because they were written to expect `variant` to be there no matter what. Looking into it, but any help would be welcome!

LeaVerou:
Actually nevermind, this is not quite ready yet.

This would break the JS setter, i.e. `button.variant = "foo"` would stop working. Does Lit provide a way to reflect in general _without_ reflecting the default value if no attribute/property is set?

LeaVerou:
Closing in favor of #512

PR #461, opened 2025-01-08
Fix #451 and other missing styles

### LINKED ISSUE #451: Fix thick text field borders in Firefox and Safari

lindsaym-fa:
Addresses #451 
This fixes a number of `wa-input`, `wa-textarea`, and `wa-select` styling issues in Safari and Firefox without regression in Chrome. `filled` and `outlined` appearances, `wa-filled` and `wa-outlined` classes, and component-level properties like `--border-color` continue to work as expected.

Default appearance before:
Screenshot 2025-01-08 at 1 11 39 PM

Default appearance  after:
Screenshot 2025-01-08 at 1 11 54 PM

Custom properties before:
Screenshot 2025-01-08 at 1 11 28 PM

Custom properties after:
Screenshot 2025-01-08 at 1 11 03 PM


src/styles/native/input.css
```diff
@@ -14,20 +14,18 @@ input:where(:not(
     [type='reset'],
     [type='submit']
   )) {
-  &:not(:host *) {
```
lindsaym-fa, line 17: This selector was preventing these properties from applying in Firefox and Safari. Removing seems to have no ill effects.

LeaVerou: The purpose of this selector was to set these at the "root" of the styling, so they could be overridden. Without it, setting e.g. `--outlined-background-color` (either by the author or the appearance utility) on `<wa-input>` will do nothing because the `.wa-text-field` within `<wa-input>` will *also* have these applied.

Presumably the reason it doesn't work in Safari and Firefox is that they don't match `:host`/`:host()` within `:is()`/`:where()`, and this selector would be rewritten as `:is(.wa-text-field, :host, textarea, input:where(:not(...)))):not(:host *)`. 

I'll do some local testing, I may be missing something and it may turn out that all is fine.

lindsaym-fa: Every styling feature was working as expected in my testing, but let me know what you find out!

lindsaym-fa:
Closing in favor of #464.

PR #446, opened 2025-01-07
Fix mobile theme selector; fixes #440

lindsaym-fa:
- #440 
- Fixes an issue where the label of the theme selector on mobile wasn't updating with the name of the selected theme
- Partially fixes an issue where the color scheme selector label wasn't changing between light and dark for new themes (not fully resolved; created #445 to track)

PR #462, opened 2025-01-08
Fix mobile color picker bug

claviska:
Fixes the issue [mentioned here](https://github.com/shoelace-style/webawesome/pull/446#issuecomment-2578093380).

bug


lindsaym-fa (approved):
Works great! ✨ Thank you!

PR #478, opened 2025-01-09
Attempt at better style isolation between themes

### LINKED ISSUE #475: Multiple themes cannot be used on the same page
While in the docs we claim multiple themes can be included on the same page, in practice this doesn't work, even if CSS layers are used for the ordering you end up with a mix of both themes.

E.g. here is a mix of Awesome & Brutalist:
image

Here is a mix of default & Active:
image

Brutalist.+ Active:
image

Active + Classic:
image

LeaVerou:
Still doesn't quite fix #475 but might be a nudge in the right direction.

### DIFF
--- a/docs/docs/themes.md
+++ b/docs/docs/themes.md
@@ -276,15 +276,19 @@ You can use these classes to apply a specific color scheme to an entire page or 
 In pre-made themes, we use a light color scheme by default.
 
 
-Additionally, styles may be scoped to the `:root` selector to be activated automatically.
-For pre-made themes, *all* custom properties are scoped to `:root`, the theme class, and `wa-light`.
+Additionally, styles may be scope to the `:root` selector to be activated automatically.
+For pre-made themes, *all* custom properties are scoped to `:where(:root)`, the theme class, and `wa-light`. 
+
+:::info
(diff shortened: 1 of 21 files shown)


docs/docs/themes.md
```diff
@@ -276,15 +276,19 @@ You can use these classes to apply a specific color scheme to an entire page or
...
 
 
-Additionally, styles may be scoped to the `:root` selector to be activated automatically.
-For pre-made themes, *all* custom properties are scoped to `:root`, the theme class, and `wa-light`.
+Additionally, styles may be scope to the `:root` selector to be activated automatically.
+For pre-made themes, *all* custom properties are scoped to `:where(:root)`, the theme class, and `wa-light`. 
+
+:::info
+We use `:where(:root)` to give these styles 0 [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) so that they can be easily overridden. If you plan on using multiple themes in your project, we recommend doing the same for your custom themes.
+:::
```
lindsaym-fa, line 284: @LeaVerou, I made this revision to the Themes doc. Mind taking a look and editing as you see fit?

lindsaym-fa (approved):
Approved pending _your_ approval of a revision to the docs! Thank you!

PR #477, opened 2025-01-09
Themes category in docs

LeaVerou:
Depends on #474 #476 

@lindsaym-fa I tried several ideas for the inline theme selector and eventually I realized I was essentially replicating navigation, and ended up adding actual pages for the themes, which contain not just their previews, but also usage instructions, snippets to copy etc. This also gives us a searchable listing page with cards, though we have no icons yet. 

After way longer than I'd care to admit, I eventually gave up on doing inline previews (see #475), they were conflicting with the page theme even in shadow DOM and I ended up just forcing the page theme on these pages. Turbo seems to be having some issues with this, @claviska I'd love your help since I see you've dealt with this before. 

One remaining UX issue is that the theme selector will basically do nothing on theme pages, since they have a forced theme. Ideally it should detect that we're on a theme page and redirect us to the page of the new theme.

PR #481, opened 2025-01-09
Rewrite color scheme logic

LeaVerou:
_(Extracted from another PR to keep it small)_

- Minimize needless swaps
- Centralize logic
- Use an actual `<wa-select>` rather than hacking it with `<wa-dropdown>`


claviska:
This is great! The only thing I see off is this:

CleanShot 2025-01-09 at 17 14 58@2x

It doesn't happen to all the labels. I can only get it to break this one for some reason.

LeaVerou:
I can just not set the width, but then they are a bit too long. Any idea why this happens? It looks like a bug with `<wa-select>`.

lindsaym-fa:
A couple of things I've noticed in testing —

I imagine this was intentionally left out of this logic, but for awareness, the theme name and description are missing from the Themes page:

https://github.com/user-attachments/assets/8564a233-d9c1-430f-af37-561539cbfe45


(I understand this page may be retired, but this is broken behavior while the page continues to exist.)


I can't consistently get into this state to give steps to reproduce, but occasionally, the entirety of the selected theme isn't applied (colors and component-specific overrides seem to still work). Here's a video in Firefox, but this also occurred for me in Safari:

https://github.com/user-attachments/assets/fb94af91-5a93-4e3e-9f25-e2cd031faa51

LeaVerou:
> I imagine this was intentionally left out of this logic, but for awareness, the theme name and description are missing from the Themes page:

Nice catch! Damn, breaking a big change into smaller commits is _really_ hard. Not sure how to best approach this, #477 depends on this, and now this depends on it :( 

> I can't consistently get into this state to give steps to reproduce, but occasionally, the entirety of the selected theme isn't applied (colors and component-specific overrides seem to still work). Here's a video in Firefox, but this also occurred for me in Safari:

Interesting. I wonder if it's a bug with Turbo. Presumably it's gone when you refresh, right?

lindsaym-fa:
> Presumably it's gone when you refresh, right?

A hard refresh usually solves it. A normal refresh does not.

It looks like maybe the last locally stored preset theme is lingering in the `<head>`? Maybe?
Screenshot 2025-01-09 at 5 19 39 PM

ETA: In this attempt, I selected 'Awesome', refreshed the page, then selected 'Brutalist'.
Screenshot 2025-01-09 at 5 23 15 PM

LeaVerou:
Interesting. What is logged when you run `localStorage.presetTheme` in the console? Does this happen with the color scheme as well or only the theme?

lindsaym-fa:
> What is logged when you run localStorage.presetTheme in the console? Does this happen with the color scheme as well or only the theme?

I would love to tell you, but I can't get back into the right state because our options are missing. 😅
Screenshot 2025-01-09 at 5 28 13 PM

LeaVerou:
Whoa, what happened there? How did this happen?

LeaVerou:
@claviska This is how it looks with no explicit widths, no idea why:
image

lindsaym-fa:
> Whoa, what happened there? How did this happen?

Not exactly sure... I'm missing icons, the theme selector, and color scheme selector on the Vercel preview in every browser. Tried to refresh, hard refresh, clear cache... No good. 🤔

LeaVerou:
Alright, I just pushed a bunch of changes!

- Move shared includes to shared template
- Rewrite critical theme JS (from 67 down to 20 loc)
- Refactor of theme picker code
- Utilize view transitions better and use them for color scheme too
- Do not store default value in localStorage
- Removed unused `*-wide` templates
- Removed explicit widths (hopefully one of you could help debug these select widths)
- Fixed theme showcase regression PLUS existing bug #482 

Could you please take another look?

claviska:
look I'm all about minimalism but this is ridiculous 😆 

CleanShot 2025-01-10 at 09 27 40@2x

LeaVerou:
Can you reproduce locally? I think it only happens in the Vercel preview

LeaVerou:
@claviska it looks like the SSR loader is failing to load in the Vercel preview, which seems unrelated to this PR?

claviska (approved):

PR #505, opened 2025-01-13
Refactor: `library.system.ts` to export supported icons

LeaVerou:
I needed a list of icons supported in the system library for the benchmark we were discussing, so I ended up cleaning this up a bit.

As another advantage, if I cannot find the icon in the variant specified, it will now fall back to the same icon in _any_ variant.

I kept the `family == "classic"` check, though not sure it serves a purpose.


claviska:
> As another advantage, if I cannot find the icon in the variant specified, it will now fall back to the same icon in any variant.

Are we sure we want this as a behavior? I foresee a lot of bug reports for "why are my filled/outlined icons showing up as outlined/filled?"

LeaVerou:
> > As another advantage, if I cannot find the icon in the variant specified, it will now fall back to the same icon in any variant.
> 
> Are we sure we want this as a behavior? I foresee a lot of bug reports for "why are my filled/outlined icons showing up as outlined/filled?"

This only applies to the system library.

claviska:
> This only applies to the system library.

We should have the discipline to be intentional about the icons we use :)

LeaVerou:
Haha ok, I'll roll it back

claviska (approved):

PR #508, opened 2025-01-13
Implement `.wa-invert`, closes #497

### LINKED ISSUE #497: `.wa-invert` class that resolves to dark mode in light mode and vice versa
We currently use JS to implement this in our theme showcase, but _I think_ it would be pretty trivial to implement this in our themes and seems like it could be useful.

LeaVerou:


lindsaym-fa (approved):
Looks great! 🎉

LeaVerou:
It's actually not — it doesn't cater for `:host-context()`, but we can deal with that later (this only becomes relevant if themes are used in components without being included in the rest of the page, which is probably rather niche).

PR #512, opened 2025-01-14
Do not reflect default values

### LINKED ISSUE #506: Do not reflect attribute default values when not actually specified on the element
A lot of upcoming improvements (e.g. #430 or making variants inheritable, see #454) need the ability to reflect an attribute without reflecting its default value. While this is primarily needed to pave the way for these improvements, I'd argue that the current situation is also a UX issue, as the cornucopia of sprouted attributes the user did not specify is making it hard to debug things (e.g. see https://github.com/shoelace-style/shoelace/discussions/2339)

The question is: how do we do this? This is a known problem in LIt, see https://github.com/lit/lit/issues/1476 . 

- We could use a custom converter that returns `null` if the value is === the default value. That means we need to figure out how to import the default converters so we can fall back to them once our custom converter is done.
- We could set `reflect: false` for these properties, and reflect them ourselves in `updated()`
- @KonnorRogers might have a better idea

LeaVerou:
Implementation on the base class and sample usage on the `appearance` property of `<wa-button>`.

- Fixes #506
- Fixes https://github.com/shoelace-style/shoelace/discussions/2339
- Paves the way for a ton of improvements

This implements a new `default` descriptor for `@property` and
- wraps the [default converter](https://lit.dev/docs/components/properties/#conversion-type) with a check: if the value is equal to the default value (regardless of how it was set), it does not create an attribute (and removes it if it was already there). Since it imports the default converter and passes the value back to it, the implementation is very minimal, and quite robust.
- wraps the default getter to return the default value if the getter returns `null` or `undefined`.

The only DX "cost" for us is specifying the default value twice — which is actually only needed for the CE manifest, the actual functionality doesn't need it, so perhaps down the line we can just patch the CEM.

### DIFF
--- a/src/components/button/button.ts
+++ b/src/components/button/button.ts
@@ -72,7 +72,8 @@ export default class WaButton extends WebAwesomeFormAssociatedElement {
   @property({ reflect: true }) variant: 'neutral' | 'brand' | 'success' | 'warning' | 'danger' = 'neutral';
 
   /** The button's visual appearance. */
-  @property({ reflect: true }) appearance: 'accent' | 'filled' | 'outlined' | 'plain' = 'accent';
+  @property({ reflect: true, default: 'accent' })
+  appearance: 'accent' | 'filled' | 'outlined' | 'plain' = 'accent';
 
   /** The button's size. */
   @property({ reflect: true }) size: 'small' | 'medium' | 'large' = 'medium';
--- a/src/internal/webawesome-element.ts
+++ b/src/internal/webawesome-element.ts
@@ -1,7 +1,17 @@
-import type { CSSResult, CSSResultGroup, PropertyValues } from 'lit';
-import { LitElement, isServer, unsafeCSS } from 'lit';
+import type { CSSResult, CSSResultGroup, PropertyDeclaration, PropertyValues } from 'lit';
+import { LitElement, defaultConverter, isServer, unsafeCSS } from 'lit';
 import { property } from 'lit/decorators.js';
 import componentStyles from '../styles/shadow/component.css';
+
+// Augment Lit's module
+declare module 'lit' {
+  interface PropertyDeclaration {
+    /**
+     * Specifies the property’s default value
+     */
+    default?: any;
+  }
+}
 
 export default class WebAwesomeElement extends LitElement {
   constructor() {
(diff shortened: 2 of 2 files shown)


KonnorRogers:
Looks pretty good to me. There's one odd edge case where if a user sets the property to the same as the default value, it will not reflect. Don't know if its worth supporting this, but it is worth noting.


https://github.com/user-attachments/assets/7a942f2a-ef7a-4683-86b8-b7e8b448bf72

LeaVerou:
> Looks pretty good to me. There's one odd edge case where if a user sets the property to the same as the default value, it will not reflect. Don't know if its worth supporting this, but it is worth noting.

That is by design. It's a compromise we need to make, otherwise complexity increases very significantly.

I think it's actually a good thing though, it teaches people about default values and it follows the design principle of [preferring the shortest form when serializing](https://github.com/w3ctag/design-principles/issues/284#issuecomment-773611014).

My main gripe with the current implementation is that we need to specify the default values twice, just for the sake of CEM. But oh well.

LeaVerou:
One issue with this approach is specific to these examples: that if we take no attribute to mean "inherit", then there is no way to specify that a button within a callout should have a neutral color. So perhaps the default value for these should be `auto` or something like that.

claviska:
> That is by design. It's a compromise we need to make, otherwise complexity increases very significantly.

To confirm, have all existing component styles been updated to _not_ use reflection?

The main reason reflection was even added was to facilitate styling. At this point, it might be better to update the policy on when we want to reflect and update any remaining components that still do.

So, if styling is handled with classes and states now, what _is_ our reflection policy? 🤔

LeaVerou:
Btw I didn't merge this because I don't think the right approach is making `neutral` the default value, it's having an `auto` default value so you can still explicitly specify a `neutral` that does not reflect. And possibly updating the getter to return resolved values (and not `auto`). See #506 for the design discussion on that.

> > That is by design. It's a compromise we need to make, otherwise complexity increases very significantly.
> 
> To confirm, have all existing component styles been updated to _not_ use reflection?

This does not disable reflection — it only removes the attribute when it's equal to the default. And yes, that's something we can (and in most cases have, but it will need more testing) encode in CSS.

claviska:
> This does not disable reflection — it only removes the attribute when it's equal to the default. And yes, that's something we can (and in most cases have, but it will need more testing) encode in CSS.

I'm aware. My point being if defaults no longer reflect, it would bork a lot of styles if they haven't already been updated.

Thanks for clarifying. I'll keep an eye out for them in the CSS.

claviska:
My other point being: we're probably reflecting things without reason now, so we may want to revisit what we _are_ still reflecting since it's likely arbitrary if not for styling.

LeaVerou:
> > This does not disable reflection — it only removes the attribute when it's equal to the default. And yes, that's something we can (and in most cases have, but it will need more testing) encode in CSS.
> 
> I'm aware. My point being if defaults no longer reflect, it would bork a lot of styles if they haven't already been updated.

Gotcha. I think we might want to be deliberate about this and opt properties in (or at least provide a way to opt them out) rather than applying it to everything at once.

claviska (approved):

PR #518, opened 2025-01-14
Attempt to fix #514

### LINKED ISSUE #514: Theme selector regression with Turbo
To reproduce:

1. Select a theme on the top right
2. Navigate to another page
3. Notice your theme seems to have reverted to default + the colors from the selected theme.

Refresh fixes it.

Looking at the page stylesheets (running `$$("link[rel=stylesheet]").map(l => l.getAttribute("href"))` in the console) shows that the default theme stylesheet is included. `localStorage` reflects the correct theme.

LeaVerou:
This fixes #514 ~but there is still a FOUC I'm trying to resolve~ okay, I think this is now ready.

### DIFF
--- a/docs/_includes/base.njk
+++ b/docs/_includes/base.njk
@@ -83,8 +83,6 @@
           </details>
         </nav>
 
-
-
         {% block header %}
           {% include 'breadcrumbs.njk' %}
           <h1 class="title">{{ title }}</h1>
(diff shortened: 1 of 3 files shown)


lindsaym-fa (approved):
Discovered #520 while testing, but that also exists on `next` and so is unrelated to these changes.

Working well from my POV. 👍 (I only tested the UI; I won't speak to anything code-related.)

Issue #220, opened 2024-11-19
Add "tintless" color variables to hold the key/accent color for each palette (`--color-foo`, not just `--color-foo-XX`)


LeaVerou:
As discussed in Color meeting with Dave. This would include both literal hues (`--wa-color-blue`) as well as semantic colors (e.g. `--wa-color-primary`). 

Then, CSS would need to be updated to use the right hue. E.g. a primary button would use `--wa-color-primary` (with an automatic text color) whereas a callout background would use an actual tint.

LeaVerou:
So, I think the best way to approach this is something maintainable down the line:
- I can write a script that adds tintless variables to one or more CSS files automatically (and updates them when re-run)
- But @lindsaym-fa will need to go through them, see if any adjustments are needed and re-run the script, verify contrast ratios are still met), and use them in the CSS (e.g. ideally `--color-*-on-loud` should use these — if it doesn't, something is off).

LeaVerou:
> So, I think the best way to approach this is something maintainable down the line:
> 
> * I can write a script that adds tintless variables to one or more CSS files automatically (and updates them when re-run)
> * But [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/126139086?s=64&amp;u=0aa393e31139272e91872864cacf1a8dbdfbf303&amp;v=4">@lindsaym-fa](https://github.com/lindsaym-fa) will need to go through them, see if any adjustments are needed and re-run the script, verify contrast ratios are still met), and use them in the CSS (e.g. ideally `--color-*-on-loud` should use these — if it doesn't, something is off).

- [x] Contrast ratio testsuite was added in https://github.com/shoelace-style/webawesome/pull/541
- [x] Tintless script added in https://github.com/shoelace-style/webawesome/pull/596

LeaVerou:
Still needed:
- [ ] Docs about these
- [ ] Using them in themes
- [ ] @lindsaym-fa 's pass

LeaVerou:
This has been done a while ago.

PR #524, opened 2025-01-14
Tintless color variables

LeaVerou:
Related to #220

(definition only for now)


LeaVerou:
I have a better approach in mind, so I'm going to close this.

PR #527, opened 2025-01-15
Finalize playful theme, closes #490

### LINKED ISSUE #490: Finalize Playful theme
- [x] Tidy up `playful.css`
- [x] Verify in theme selector
- [x] Add changelog
- [ ] Testing and tweaks

lindsaym-fa:
Prepares the Playful theme for release


src/styles/themes/playful.css
```diff
@@ -29,53 +29,53 @@
...
   --wa-color-brand-fill-normal: var(--wa-color-violet-80);
   --wa-color-brand-fill-loud: var(--wa-color-violet-50);
   --wa-color-brand-border-quiet: var(--wa-color-violet-90);
   --wa-color-brand-border-normal: var(--wa-color-violet-70);
   --wa-color-brand-border-loud: var(--wa-color-violet-50);
-  --wa-color-brand-on-quiet: var(--wa-color-violet-40);
+  --wa-color-brand-on-quiet: var(--wa-color-violet-30);
   --wa-color-brand-on-normal: var(--wa-color-violet-30);
-  --wa-color-brand-on-loud: var(--wa-color-violet-95);
+  --wa-color-brand-on-loud: white;
```
LeaVerou, line 40: I've been wondering if we want to have these, since they're the same across all themes, including the default (that we are importing). OTOH it could be argued it helps show the full picture.

LeaVerou:
I love this theme. It's such a great option for certain use cases for which there are very few options out there (e.g. apps for children come to mind).

Some design nits, all non-blocking:

Something feels a little off with badges:
image
Maybe they need a slightly bolder font to match the thick outline? Or maybe the outline is too thick? Not sure.

I'm unsure about using a dashed border decoratively, as it tends to often have meaning:
image
image

I love the tactility of the buttons. A little unsure about the hover effect though, especially the small rotation (?):

https://github.com/user-attachments/assets/595ba3ca-06c6-4224-a5ee-df7edbdedb90

Possibly unrelated to this theme, there is a gap in the rounded corners here:
image

Should the card footer have a slightly darker color?
image

The tooltip seems like it should be a little bolder, also the extra rounding looks a bit awkward when the content is narrow, like in this case:
image

The tree checkboxes look a bit too large:
image

Should the handle of an off switch be styled in a "raised" way? 
image

LeaVerou (approved):

LeaVerou:
(I also resolved the conflicts)

src/styles/themes/playful.css
lindsaym-fa: I've been struggling with this myself... I wonder if this would be a good rule of thumb: If 4 or fewer tokens differ from the default, only specify those tokens. If more than 4 differ, specify all of them.

I found that if all but just a few tokens are different, leaving just the redundant tokens out feels misleading — like I forgot to specify those that are missing. But changing just a few tokens has the opposite problem...

lindsaym-fa:
Thanks, @LeaVerou! I made most of the changes you suggested, and things look better!

As for the tooltip and tree checkboxes, I think it's best that we change the defaults... I'll create separate PRs for these.

PR #533, opened 2025-01-15
Checkbox improvements

### LINKED ISSUE #529: `wa-checkbox` missing `part="control"`
We have a `control` part documented for `<wa-checkbox>` (http://localhost:4000/docs/components/checkbox/#css-parts), but the part doesn't seem to exist in the component.

We could either update the documentation or add the expected part. I'd recommend the latter since it's useful for styling.

LeaVerou:
- Added `part="control"`, fixes #529
- Removed wrapper div which is no longer needed
- Removed `form-control--has-hint` class which is no longer used anywhere


claviska (approved):

lindsaym-fa (approved):
Thanks, @LeaVerou!

PR #537, opened 2025-01-15
[Card] Reintroduce `--border-width` and `--border-radius`, fixes #531

### LINKED ISSUE #531: Fix `wa-card` between border and content at corners
Some of our new themes expose a gap between the border and content (especially images and headers and footers with background colors) of `wa-card`. The issue affects all themes; some are more noticeable than others.

Image
Image

LeaVerou:
We _could_ get away with only reintroducing `--border-width` and still use `inherit` for `border-radius`, doing gymnastics with negative margins and transparent borders to account for the difference, but I decided it was not worth it, better to keep the CSS simple and thus easily customizable.

### DIFF
--- a/src/components/card/card.css
+++ b/src/components/card/card.css
@@ -1,29 +1,35 @@
 :host {
   --spacing: var(--wa-space-xl);
+  --border-width: var(--wa-panel-border-width);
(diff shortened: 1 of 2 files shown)


lindsaym-fa (approved):
Thanks, @LeaVerou! ✨

PR #541, opened 2025-01-15
Quick proof of concept contrast tests

### LINKED ISSUE #540: Add tests to ensure the contrast ratios we promise are met
We [currently promise that](http://localhost:4000/docs/tokens/color/):

> - A difference of 40 ensures a minimum 3:1 contrast ratio, suitable for large text and icons (AA)
> - A difference of 50 ensures a minimum 4.5:1 contrast ratio, suitable for normal text (AA) and large text (AAA)
> - A difference of 60 ensures a minimum 7:1 contrast ratio, suitable for all text (AAA)

Since this might be a legal requirement for folks, we should have a testsuite that reads our palettes and validates this.

Given that we have 11 hues, this means testing the following combinations:
- **5** combinations with a difference of **40**: 10/50, 20/60, 30/70, 40/80, 50/90
- **4** combinations with a difference of **50**: 10/60, 20/70, 30/80, 40/90
- **3** combinations with a difference of **60**: 10/70, 20/80, 30/90

I.e. 12 tests per hue.
If we want to also test APCA, this doubles the number of tests, since APCA distinguishes between foreground and background.

LeaVerou:
Closes #540. 
This is completely ad hoc, but it seems to work fine, and uses the right exit codes if we want to use it in CI.

To run: `cd src/styles/color && node contrast.test.js`

Current output for `next`:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,6 +34,7 @@
         "chalk": "^5.3.0",
         "change-case": "^4.1.2",
         "chokidar": "^3.5.3",
+        "colorjs.io": "^0.6.0-alpha.1",
         "command-line-args": "^5.2.1",
         "comment-parser": "^1.4.1",
         "cspell": "^6.18.1",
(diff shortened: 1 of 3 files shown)


claviska:
Interesting! I've never thought to use the test runner this way. This looks good to me as long as other tests continue to run when it fails, which seems like the case here.

claviska (approved):

PR #543, opened 2025-01-16
Improve default tooltip styles

lindsaym-fa:
- Gives tooltips an inverted appearance by default, using `--wa-tooltip-background-color: var(--wa-color-text-normal)` and `--wa-tooltip-content-color: var(--wa-color-surface-default)`
- Removes redundant styles from themes

Example in Playful theme—
Light mode:
Screenshot 2025-01-15 at 8 15 16 PM

Dark mode:
Screenshot 2025-01-15 at 8 16 04 PM


LeaVerou (approved):
Lovely.

PR #547, opened 2025-01-16
Add Matter theme

### LINKED ISSUE #406: Add Material theme
- [x] Add new theme that mimics https://m3.material.io
- [x] Organize into separate files (see `default`)
- [x] Verify via theme selector
- [ ] Add changelog

lindsaym-fa:
Adds Matter theme, based on Material. (Name suggestions welcome.)

I ended up using a Google Sans-esque font for the body and headings rather than Roboto, which I believe is Material's default. It looks nicer and is more immediately reminiscent of Material imo, but I'm torn if we shouldn't just use the default that Material sets.

### DIFF
--- a/src/styles/color/mild.css
+++ b/src/styles/color/mild.css
@@ -1,7 +1,7 @@
 :where(:root),
 :host,
 :where([class^='wa-theme-'], [class*=' wa-theme-']),
-.wa-palette-gentle {
+.wa-palette-mild {
   --wa-color-red-95: #ffeeec;
   --wa-color-red-90: #ffddda;
   --wa-color-red-80: #f9bbb3;
--- a/src/styles/color/mild.css
+++ b/src/styles/color/mild.css
@@ -1,0 +1,100 @@
+:where(:root),
+:host,
+:where([class^='wa-theme-'], [class*=' wa-theme-']),
+.wa-palette-gentle {
+  --wa-color-red-95: #ffeeec;
+  --wa-color-red-90: #ffddda;
+  --wa-color-red-80: #f9bbb3;
+  --wa-color-red-70: #f0978d;
+  --wa-color-red-60: #e47468;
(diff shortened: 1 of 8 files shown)


LeaVerou:
> I ended up using a Google Sans-esque font for the body and headings rather than Roboto, which I believe is Material's default. It looks nicer and is more immediately reminiscent of Material imo, but I'm torn if we shouldn't just use the default that Material sets.

It would help to see screenshots side by side if not too much of a hassle :)

src/styles/color/mild.css
```diff
@@ -0,0 +1,100 @@
+:where(:root),
+:host,
+:where([class^='wa-theme-'], [class*=' wa-theme-']),
+.wa-palette-gentle {
```
LeaVerou: Either the filename should be `gentle.css` or the class should be `wa-palette-mild`

LeaVerou (approved):
LGTM!

Some comments:
- This is another theme that shows that decisions like `pill` need to happen at the design system level. Now we have a `pill` attribute that e.g. does nothing for buttons.
- Nit: the accent purple seems a tad too light. Ours: image Theirs: image

src/styles/color/mild.css
lindsaym-fa: Good catch. I went back and forth on the name, and here's a remnant. 😅 Fixed!

lindsaym-fa:
Thanks, @LeaVerou!

I went a shade darker for the accent buttons.

For interest and posterity, here's the Google Sans-esque font (Wix Madefor Text) vs. Roboto:
Screenshot 2025-01-16 at 11 55 16 AM
Screenshot 2025-01-16 at 11 55 37 AM

LeaVerou:
> For interest and posterity, here's the Google Sans-esque font (Wix Madefor Text) vs. Roboto:

Okay, I see what you mean and I agree with your decision 👍🏼

PR #526, opened 2025-01-14
Separate meaningful theme parts out

LeaVerou:
- Split out color tokens, typography, overrides. Some for mixing & matching (colors, typography), some for maintainability (overrides).
- Implemented a potential fix (or at least improvement) for #475.
- Added `.wa-palette-*` classes to each color palette

Haven't done Gl[ao]ssy and Playful to not conflict with @lindsaym-fa's work.

PR #552, opened 2025-01-16
Theme files housekeeping, closes #519

### LINKED ISSUE #519: Clean up redundant dark mode properties across themes
Many themes specify custom properties for `wa-dark` that exactly match those found in `default/color-semantic.css`. These should be cleaned up so that only those properties that override the defaults are left.

lindsaym-fa:
- Splits `playful.css` and `glossy.css` into separate files, following #526 
- Removes redundant `wa-dark` color tokens across themes, closing #519


LeaVerou (approved):
Haven't reviewed super thoroughly, but from a quick skim this looks great.

PR #555, opened 2025-01-16
Show contrast ratios in contrast pair tables

LeaVerou:
- Show actual WCAG contrast ratios instead of `{tint1} on {tint2}` text.
- Separate code to generate the data structure for palettes and the testing code

Now that we have access to the color tokens in 11ty we could do even more cool things with them later 😁 

image

### DIFF
--- a/docs/_data/palettes.js
+++ b/docs/_data/palettes.js
@@ -1,0 +1,1 @@
+export { default as default } from '../../src/styles/color/palettes.js';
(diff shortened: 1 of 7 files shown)


lindsaym-fa (approved):
Awesome. 🤩

claviska:
I was about to call this out saying it didn't look enough contrast in the text for the darker ones, but Color Slurp proved me wrong. APCA when? 🫠 

CleanShot 2025-01-16 at 13 15 47@2x

claviska (approved):

LeaVerou:
> I was about to call this out saying it didn't look enough contrast in the text for the darker ones, but Color Slurp proved me wrong. APCA when? 🫠

Haha soon. The code is trivial, it's just a matter of doing it.

PR #560, opened 2025-01-16
Basic mixing and matching docs

LeaVerou:
I think we should document this, even if we don't demo it, since the files exist and the mixing and matching actually _does_ work!


lindsaym-fa (approved):
Nice!

PR #565, opened 2025-01-16
Demo for mixing and matching

LeaVerou:
- [x] Add URL params for mixing and matching (`colors`, `palette`, `typography`) to demo page
- [x] Make demo for theme pages to update these

image
image

image
image

Now I just need to take some good screenshots for the Kickstarter announcement…

### DIFF
--- a/docs/_layouts/theme.njk
+++ b/docs/_layouts/theme.njk
@@ -10,7 +10,8 @@
 <p id="mix_and_match" class="wa-gap-m">
   <strong>
     <wa-icon name="merge" slot="prefix"></wa-icon>
-    Remix:
+    Remix
+    <wa-icon-button href="#remixing" name="circle-question" slot="suffix" variant="regular" label="How to use?"></wa-icon-button>
   </strong>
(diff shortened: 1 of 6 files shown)


claviska (approved):

claviska:
Are we sure we want to encourage mixing and matching? Isn't this a major part of what the themer does for them?

LeaVerou:
> Are we sure we want to encourage mixing and matching? Isn't this a major part of what the themer does for them?

Our thinking was:
1. We've already split the files anyway (for maintainability as well)
2. We have no themer yet, and this gets them some flexibility already, so I think the overlap is actually a good thing! Once the themer is there, this functionality can be integrated in it (and I think is still useful, since it gets you something custom very quickly). In fact, it paves an incremental path towards the themer, which I think can be great for shipping it.
3. From a business pov, it increases the value of the Pro themes, since they can now do more things.
4. It _seems_ to actually produce decent results (against our expectations!)
5. Last but not least, we did promise it last week :stuck_out_tongue:

PR #579, opened 2025-01-21
Ensure the `<wa-select hoist>` fix works on the backers website

### LINKED ISSUE #332: Fix <wa-select> examples
Image

claviska:
This was working locally because imports are faster in dev, but it's broken on the backers website.

To repro, visit: https://backers.webawesome.com/docs/components/select

And open any select.

CleanShot 2025-01-21 at 12 29 47@2x

Fixes #332

### DIFF
--- a/docs/docs/components/select.md
+++ b/docs/docs/components/select.md
@@ -420,5 +420,7 @@ This can be hard to conceptualize, so heres a fairly large example showing how l
   //
   // TODO - remove once we switch to the Popover API
   //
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):
This won't help with other elements or elements in shadow DOM, but it's better than the current situation.

claviska:
Yep, quick fix to get the docs updated. The real fix is adding the Popover API to affected components. Thanks!

Issue #580, opened 2025-01-21
Certain attributes (e.g. `variant`, `size`) should inherit when not explicitly set


LeaVerou:
This is a spinoff out of #506 .

- Take a look at [this](https://backers.webawesome.com/docs/components/button-group#button-sizes). It should not take repeating `size` for every button! The author intent can be sufficiently signaled by an attribute on the group.
- Customer request here: https://github.com/shoelace-style/webawesome-alpha/discussions/139
- Same for `pill`: https://backers.webawesome.com/docs/components/button-group#pill-buttons
- Do you really still want a button to default to neutral within a brand callout?
- Etc. Pretty sure I've seen other cases around the library.

Inherited attributes have precedent in HTML too: that's how `lang` and `dir` behave.

Currently, our variant and size utils inherit, but only when used on native styles, e.g. see #454.

At first I tried to solve this in #512, but I think it should be separate, though it's better to tackle this after #506 is solved. Then we can just give them a default value of `inherit` and handle that.

It's not completely trivial to solve, mainly because these utils define default values on `:root`, but we want these to have _different_ default values on certain elements. E.g. the default variant overall is `neutral`, but callouts have a default variant of `brand`. How do we give these properties a default value that is different, without overriding the actual properties is still a known unknown.

lindsaym-fa:
Fun! I was _just_ looking at [a backer idea related to small/large dropdowns](https://github.com/shoelace-style/webawesome-alpha/discussions/139), where they imagined something like this for `<wa-dropdown>`:
```html
<wa-dropdown size="small">
  <wa-button slot="trigger" caret>Dropdown</wa-button><!-- Inherits size attribute from dropdown -->
  <wa-menu>
    <wa-menu-item>Dropdown Item 1</wa-menu-item><!-- Inherits size attribute from dropdown -->
    <wa-menu-item>Dropdown Item 2</wa-menu-item>
    <wa-menu-item>Dropdown Item 3</wa-menu-item>
  </wa-menu>
</wa-dropdown>
```

LeaVerou:
@lindsaym-fa Awesome, thanks! I added it to the first post, and I can now tag this "Request" 😁

LeaVerou:
I’ve _really_ struggled with the CSS for this, so I'm going to write stuff down here for posterity. **Feel free to skip this unless you're interested in actively working on this problem.**

[...]

PR #582, opened 2025-01-21
Infrastructure for inheritable props, make `variant` inherit

### LINKED ISSUE #454: Native button inherits variant but `<wa-button>` does not

LeaVerou:
This PR implements:
- A new `initial` descriptor for `@property`, which sets the initial value for inherited properties (i.e. what is applied if there is no inherited value). When `initial` is used, it automatically applies `default: 'inherit'` if no default value is set.
- Inheritable variants (sizes will be implemented in a future PR) for both components and utility classes
- Makes variants inheritable for the 4 components that support `variant` (button, callout, badge, tag).

It fixes #454 #592 and makes progress on #580.
You can see how it works in http://localhost:4000/docs/experimental/inherit/ (local link)

I also wrote a script to generate `variants.css`, we can decide to keep it or not. 🤷🏼‍♀️  Ideally if we end up generating some of our CSS we'd want something more robust, this was just a quick hack so I could experiment with different ideas without refactoring 90+ declarations.

**Note:** Currently initial variants other than `neutral` need to be specified in `variants.css`. We can reduce coupling by having the base class add a state like `initial-variant-brand` that the CSS can then use. I didn't do it because I figured what's the point if we still need to add classes for the native styles (e.g. `.wa-callout`), but it perhaps could be useful for author components.


LeaVerou:
@lindsaym-fa I pushed a fix for the issues you found, and a bunch more tests:

image
image
image
image

lindsaym-fa:
After some deliberation and separate discussions offline, I just want to reiterate that I think this is a helpful trick to have up our sleeves. Two key reasons:

1. Inheriting `variant` is a great guide rail to help developers implement good UX with minimal effort. E.g., while `brand` and `neutral` are both pretty innocuous, any combination of `success`, `warning`, and `danger` across an ancestor component and its descendants sends mixed signals to end users, and the consequences could be devastating. (Not to mention the poor aesthetics. 😉)
2. As you, @LeaVerou, put it best, "once the user has expressed intent for a variant on an ancestor, that's a stronger indicator of intent than whatever default _we_ have picked for a given element."


Besides the failing tests, the only other thing left to touch up from my POV is the case of a variant-less button within a variant-less callout:
Screenshot 2025-01-24 at 5 41 06 PM
(The rule that applies here, from my understanding, is that if no `variant` is specified on an ancestor, then no `variant` is inherited by its descendants.)

LeaVerou:
@lindsaym-fa 

> Besides the failing tests, the only other thing left to touch up from my POV is the case of a variant-less button within a variant-less callout:

That was.an easy fix! I was simply not outputting any rules for elements with a neutral default variant, so they always inherited.

Working on tests now.

LeaVerou:
@lindsaym-fa Tests now passing, so this should be ready for another round!

src/internal/webawesome-element.ts
```diff
@@ -160,36 +168,69 @@ export default class WebAwesomeElement extends LitElement {
...
+
+              if (value === 'inherit') {
+                // If we've reached this point and we still have `inherit`, we just ran out of parents
+                return options.initial;
+              }
+
+              return value;
+            },
+          });
+        }
```
claviska: Do we know how expensive this traversal will be for complex pages? (People do [crazy things](https://github.com/shoelace-style/shoelace/discussions/2355), and having multiple elements doing it will add up _quickly_.)

What happens when one of the parent nodes changes dynamically? Will the variants go out of sync then?

LeaVerou: Note that this is not used anywhere internally — it is provided solely as a convenience for users. But yeah, I’ve been wondering about making that a method, to communicate better that it’s an expensive operation.

LeaVerou: (or just not including it at all in this MVP and see if anyone needs it — I added it because I thought we might, but maybe not?)

lindsaym-fa (approved):
Everything is working as expected from my testing!

Approved, barring [any performance concerns](https://github.com/shoelace-style/webawesome/pull/582/files#r1930767860).

LeaVerou:
> Approved, barring [any performance concerns](https://github.com/shoelace-style/webawesome/pull/582/files#r1930767860).

I turned the `computedVariant` getter into a `getComputed(propName)` method to better communicate its performance characteristics. To be clear, that's not something we currently use internally, so it doesn't affect performance.

Do note that this method only takes the variant specified via the `variant` attribute into account. Perhaps it should also take the util classes into account before we make it part of the public API?

claviska (approved):

PR #589, opened 2025-01-22
Implement `wide` page setting and add it to Themes and Patterns

### LINKED ISSUE #588: Add `wide` page mode
Needed for both themes and patterns.

I plan to implement it via 11ty data. That way we can set it on entire sections or individual pages.

LeaVerou:
Fixes #588


kelseythejackson (approved):
Perfect

Issue #67, opened 2024-03-13
Remove `closable` from tabs


claviska:
For accessibility, tabs cannot have nested interactives so we need to remove the `closable` attribute and update the docs to show how this can be done with adjacent buttons instead.

PR #590, opened 2025-01-22
Use native events when possible

claviska:
<details>
  <summary>Show original tasks</summary>

- [x] Update all components that use them
  - [x] wa-input
    - [x] checkbox
    - [x] color-picker
    - [x] input
    - [x] radio-group
    - [x] select
    - [x] slider
    - [x] switch
    - [x] textarea
  - [x] wa-change
    - [x] checkbox
    - [x] color picker
    - [x] image comparer
    - [x] input
    - [x] radio group
    - [x] rating
    - [x] select
    - [x] slider
    - [x] switch
    - [x] textarea
  - [x] wa-blur
    - [x] button
    - [x] checkbox
    - [x] color-picker
    - [x] icon-button
    - [x] input
    - [x] radio
    - [x] radio button
    - [x] select
    - [x] slider
    - [x] switch
    - [x] textarea
  - [x] wa-focus
    - [x] button
    - [x] checkbox
    - [x] color-picker
    - [x] icon-button
    - [x] input
    - [x] radio
    - [x] radio-button
    - [x] select
    - [x] slider
    - [x] switch
    - [x] textarea
- [x] Ensure all `@wa- input|change|blur|focus` handlers have been updated
- [x] Ensure all `@event` JSDoc comments have been updated
- [x] Remove the events from `src/events/*`
- [x] Update React wrappers to ignore unprefixed (native) events
- [x] Verify build
- [x] Verify tests
- [x] Update changelog

</details>

### TODO

- [x] Update examples that use the old events
- [x] Update any docs that show/use the old events

---

I _have not_ updated `wa-invalid` yet and there is no `wa-valid`. I'm not 100% sure we should without fully understanding how it affects form validation, so I was hoping to hear from @KonnorRogers since he has the most knowledge of `WebAwesomeFormAssociatedElement`.

@LeaVerou the `wa-close` event was only used by `<wa-tab>`, and the [closable feature](#67) was removed awhile ago so I simply removed the event.

This also doesn't do anything with `toggle` and `beforeToggle`, as Dialog/Drawer use a `wa-show` and `wa-after-show` pattern and native dialogs have `open`, `close` as well as `beforetoggle` and `toggle`, which might merit more discussion.


docs/docs/resources/changelog.md
```diff
@@ -14,6 +14,11 @@ During the alpha period, things might break! We take breaking changes very serio
 
 ## Next
 
+- 🚨 BREAKING: updated all components to use native events instead of `wa-` prefixed events. This will allow components to work more like native elements in your code, frameworks, third-party plugins, etc. To update your code, simply remove the prefix from your event listeners for the following events.
+  - `wa-input` => `input`
+  - `wa-change` => `change`
+  - `wa-blur` => `blur` (this event will no longer bubble)
+  - `wa-focus` => `focus` (this event will no longer bubble)
```
LeaVerou, line 21: Btw we could emit `focusin` and `focusout` which are the bubbling versions (if they don't already leak out anyway)

claviska: Both are composed, so they'll be retargeted. 😄

src/internal/webawesome-element.ts
```diff
@@ -159,6 +159,17 @@ export default class WebAwesomeElement extends LitElement {
...
   }
 
+  /**
+   * Given a native event, this function ensures it's composed and, if not, dispatches it again as a composed event.
+   * This is useful for relaying native events such as `change`, which will otherwise not be retargeted. It is safe,
+   * albeit sloppy, to call this on composed events, as it will no-op.
+   */
+  dispatchComposedEvent(event: Event) {
+    if (!event.composed) {
+      this.dispatchEvent(new Event(event.type, { ...event, composed: true }));
```
LeaVerou: This will cast everything to `Event`, should we do this instead?

```suggestion
      this.dispatchEvent(new event.constructor(event.type, { ...event, composed: true }));
```

src/components/input/input.ts
```diff
@@ -242,20 +234,15 @@ export default class WaInput extends WebAwesomeFormAssociatedElement {
     if (this.value !== '') {
       this.value = '';
       this.dispatchEvent(new WaClearEvent());
-      this.dispatchEvent(new WaInputEvent());
-      this.dispatchEvent(new WaChangeEvent());
+      this.dispatchEvent(new Event('input'));
```
LeaVerou:
```suggestion
      this.dispatchEvent(new InputEvent('input'));
```

LeaVerou (approved):
This is wonderful, thank you so much @claviska! I left some comments.

We can merge it as-is and open separate issues for the other things that remain.

One thing to definitely check before merging is that these actually bubble. I have a faint recollection that you need to specify `bubbles: true` explicitly even for events where it's the default.

docs/docs/resources/changelog.md
LeaVerou: Awesome!

claviska:
> One thing to definitely check before merging is that these actually bubble. I have a faint recollection that you need to specify bubbles: true explicitly even for events where it's the default.

I tested `InputEvent` and it bubbles in all three browsers.

CleanShot 2025-01-22 at 15 54 01@2x

I believe the composed/bubbles/cancelable options are false by default for `Event` and `CustomEvent`.

claviska:
> This will cast everything to Event, should we do this instead?

It's [currently] only used for `change` which is `Event`, but it can't hurt. Had to do some TypeScript-fu for it:

```ts
this.dispatchEvent(new (event.constructor as typeof Event)(event.type, { ...event, composed: true }));
```

KonnorRogers:
> Show original tasks
> 
> 
> ### TODO
> 
>     * [ ]  Update examples that use the old events
> 
>     * [ ]  Update any docs that show/use the old events
> 
> 
> I _have not_ updated `wa-invalid` yet and there is no `wa-valid`. I'm not 100% sure we should without fully understanding how it affects form validation, so I was hoping to hear from @KonnorRogers since he has the most knowledge of `WebAwesomeFormAssociatedElement`.
> 
> @LeaVerou the `wa-close` event was only used by `<wa-tab>`, and the [closable feature](#67) was removed awhile ago so I simply removed the event.
> 
> This also doesn't do anything with `toggle` and `beforeToggle`, as Dialog/Drawer use a `wa-show` and `wa-after-show` pattern and native dialogs have `open`, `close` as well as `beforetoggle` and `toggle`, which might merit more discussion.

I believe `wa-*` components emit both `invalid` and `wa-invalid` 🤔 I'd have to double check that. But iirc all component use `invalid` internally.

claviska:
> I believe wa-* components emit both invalid and wa-invalid 🤔 I'd have to double check that. But iirc all component use invalid internally.

~~@KonnorRogers I don't see a `wa-valid` anywhere 😬~~ (I also misread your message)

The question should have been: will changing `wa-invalid` to `invalid` (which may or may not mean removing ours and letting the native event do its thing) cause problems with the form validation logic?

claviska:
@LeaVerou docs are updated to use the new events now as well. This is ready to merge. If we want to tackle `wa-invalid`, we can track it in a separate issue.

KonnorRogers:
> > I believe wa-* components emit both invalid and wa-invalid 🤔 I'd have to double check that. But iirc all component use invalid internally.
> 
> ~@KonnorRogers I don't see a `wa-valid` anywhere 😬~ (I also misread your message)
> 
> The question should have been: will changing `wa-invalid` to `invalid` (which may or may not mean removing ours and letting the native event do its thing) cause problems with the form validation logic?

I don't believe so. I believe I left it in place because we had `sl-invalid`, the only people it affects is users using `wa-invalid` , but it should not affect the form control itself.

There's no `valid` event because the native platform does not have a `valid` event.

BUT here's where it gets tricky.

`wa-invalid` bubbles.
`invalid` does not bubble.

docs/docs/resources/changelog.md
```diff
@@ -14,6 +14,11 @@ During the alpha period, things might break! We take breaking changes very serio
 
 ## Next
 
+- 🚨 BREAKING: updated all components to use native events instead of `wa-` prefixed events. This will allow components to work more like native elements in your code, frameworks, third-party plugins, etc. To update your code, simply remove the prefix from your event listeners for the following events.
+  - `wa-input` => `input`
+  - `wa-change` => `change`
+  - `wa-blur` => `blur` (this event will no longer bubble)
```
LeaVerou:
```suggestion
  - `wa-blur` => `blur` (this event will no longer bubble, use `focusout` for a bubbling version)
```

docs/docs/resources/changelog.md
LeaVerou:
```suggestion
  - `wa-focus` => `focus` (this event will no longer bubble, use `focusin` for a bubbling version)
```

PR #593, opened 2025-01-23
Inheritable `size`

LeaVerou:
This makes progress on #580 and addresses the customer request in https://github.com/shoelace-style/webawesome-alpha/discussions/139

Note that this also paves the way for differentiating other lengths based on size, e.g. if someone also wanted to switch border radius across sizes, they could do:
```css
--wa-border-radius: calc(
      var(--wa-is-small,  0) * var(--wa-border-radius-s) 
    + var(--wa-is-medium, 1) * var(--wa-border-radius-m) 
    + var(--wa-is-large,  0) * var(--wa-border-radius-l)
  );
```

I'll update the docs once we have consensus on the code.


src/styles/utilities/size.css
```diff
@@ -1,18 +1,29 @@
...
+  --wa-size-large: var(--wa-font-size-l, 1.25rem); /* size=large */
 
-  --space-xs: var(--wa-space-xs, 0.5rem);
-  --space-s: var(--wa-space-s, 0.75rem);
-  --space-m: var(--wa-space-m, 1rem);
-  --space-l: var(--wa-space-l, 1.25rem);
+  --wa-space-xsmall: var(--wa-space-xs, 0.5rem);
+  --wa-space-small: var(--wa-space-s, 0.75rem);
+  --wa-space-medium: var(--wa-space-m, 1rem);
+  --wa-space-large: var(--wa-space-l, 1.25rem);
```
lindsaym-fa: I don't think these names are sufficiently different from our `--wa-space-*` theme tokens. `--wa-space-small` has no recognizable semantics that distinguish it from `--wa-space-s`. It's unclear when I would use one over the other and what effect that might have.

A possible alternative is prefixing each of these in a way that still indicates they're global while scoping them to a specific purpose, like `--wa_size-*`. So, we'd have e.g. `--wa_size-font-size-s` and `--wa_size-space-s`. I don't absolutely love it, so maybe there's a better option out there.


But I wonder: How important is it that these variables are global and public? Is there a clear need for folks to customize these sizing utilities on their own? (I'll admit, it's a bit hard to wrap my head around the 'Customizing Sizes' part of the docs since I'm not sure when I'd use this in practice.)
Or, another question: is renaming these variables crucial for size inheritance, or could it be considered separately?

src/styles/utilities/size.css
```diff
@@ -1,18 +1,29 @@
...
+:host {
   /* Defaults. Components are expected to override these */
-  --size-xs: var(--wa-font-size-xs, 0.75rem); /* Used for hints etc */
-  --size-s: var(--wa-font-size-s, 0.875rem); /* size=small */
-  --size-m: var(--wa-font-size-m, 1rem); /* default */
-  --size-l: var(--wa-font-size-l, 1.25rem); /* size=large */
+  --wa-size-xsmall: var(--wa-font-size-xs, 0.75rem); /* Used for hints etc */
+  --wa-size-small: var(--wa-font-size-s, 0.875rem); /* size=small */
+  --wa-size-medium: var(--wa-font-size-m, 1rem); /* default */
+  --wa-size-large: var(--wa-font-size-l, 1.25rem); /* size=large */
```
lindsaym-fa: Why are these variables named `xsmall`, `small`, `medium`, and `large` while the utility classes — which more directly correspond to the `size="[small | medium | large]"` attribute — use `s`, `m`, and `l`? Wouldn't the reverse be more predictable?

LeaVerou: Because they represent a different thing and `--wa-size-s` etc was taken. `--wa-size-s` is the "literal" token, whereas `--wa-size-small` is the "relative" one, i.e. the breakpoint for sizes. Better naming ideas welcome, I struggled a lot with this one!

LeaVerou: Maybe `--wa-size-breakpoint-s`?

src/styles/utilities/size.css
LeaVerou: > Or, another question: is renaming these variables crucial for size inheritance, or could it be considered separately?

It was at some point, though it may not be required with the current solution I settled on, so I'll take another look. 

> But I wonder: How important is it that these variables are global and public? Is there a clear need for folks to customize these sizing utilities on their own? 

It's probably advanced usage. Basically, it's needed for shifting *all* sizes up or down, so that e.g. `size=medium` resolves to `--wa-space: var(--wa-space-s)` and so on.
Perhaps instead of four breakpoints we just need one property for each thing, e.g. `--wa-space-shift` and`--wa-size-shift` for shifting the scale up/down.

> (I'll admit, it's a bit hard to wrap my head around the 'Customizing Sizes' part of the docs since I'm not sure when I'd use this in practice.)

I think their naming is a big part of why that's so confusing, so better naming could be a win overall.
But also, I _think_ (it's been some time since I looked at the code) part of the reason the docs are so confusing is exactly *because* these breakpoints are not global so you have to work around that.

lindsaym-fa: Ooo, `--wa-[size|space]-shift` could be _really_ interesting. If we don't need to worry about renaming these variables now, we can always pursue this idea for a future improvement — I imagine it'll take some extra thought.

LeaVerou:
@lindsaym-fa Alright, like I suspected with the newer solution these did not actually need to be global, so I just reverted to the previous local breakpoints and this is now a much smaller PR :D 

Agreed we can explore `--wa-size-shift` and `--wa-space-shift` later. They are definitely _possible_, they would just take some slightly awkward CSS in `size.css`. We could start by only supporting `-1`, `0`, `1`, and expanding to other shifts later.

docs/docs/components/card.md
```diff
@@ -143,3 +143,42 @@ If using SSR, you need to also use the `with-image` attribute to add an image to
...
+</wa-card>
+
+<wa-card with-footer size="large">
+  <p>This is a large card.</p>
+
+  <footer slot="footer" class="wa-flank">
+    <wa-button variant="brand" pill>More Info</wa-button>
+    <wa-rating></wa-rating>
+  </footer>
+</wa-card>
```
lindsaym-fa:
```suggestion
<wa-card with-footer size="small">
  This is a small card.

  <footer slot="footer" class="wa-flank">
    <wa-button variant="brand" pill>More Info</wa-button>
    <wa-rating></wa-rating>
  </footer>
</wa-card>

<wa-card with-footer size="medium">
  This is a medium card (default).

  <footer slot="footer" class="wa-flank">
    <wa-button variant="brand" pill>More Info</wa-button>
    <wa-rating></wa-rating>
  </footer>
</wa-card>

<wa-card with-footer size="large">
  This is a large card.

  <footer slot="footer" class="wa-flank">
    <wa-button variant="brand" pill>More Info</wa-button>
    <wa-rating></wa-rating>
  </footer>
</wa-card>
```

Mind if we remove the `<p>` tags to prevent this extraneous bottom margin in the examples? (Probably need to give more thought to the style that applies this default margin between elements...)
Screenshot 2025-01-30 at 6 29 29 PM

lindsaym-fa (approved):
Everything is working great! I left one small suggestion, but otherwise: approved. ✨

PR #594, opened 2025-01-24
Refactor native select caret styles, fixes #571 and #572

### LINKED ISSUE #571: Native `<select>` not inside a `<label>` has no caret
https://github.com/shoelace-style/webawesome-alpha/issues/147

I committed to fixing this in the next release as I think it's pretty high priority. There is no way to work around it (short of not using Native Styles at all), and a `<select>` with no caret is pretty unusable :/ 

However, it will not be an easy fix: we currently use a pseudo-element with a mask so we can give the SVG the right color. A `<select>` cannot have a pseudo-element. The native caret is not accessible in any way.

I see the following avenues:
- I wonder if there is a combination of blending modes that we can use to color the caret. Then we can use `background-blend-mode` and combine it with a solid color layer.
- Failing that, we may end up having to give up on a custom caret and just use the browser default :/

### LINKED ISSUE #572: Native `<select>` caret is blocking interactions
https://github.com/shoelace-style/webawesome-alpha/issues/148

This is an easy fix (see the workaround I posted), though we may want to resolve #571 first, as that will probably automatically fix this too.

lindsaym-fa:
For styling a native `<select>`, the story goes:
- We can't use `::after` on a `<select>`.
- We can't use `::after` on a `<label>` containing the select because it won't appear for selects outside of a label.
- We can't use the technique of `background-color` + `mask` on a `<select>` to get the right shape and color of the caret since...well, the select itself will be masked.

So, in a live discussion of #571, we chose to use the browser's default caret rather than trying to replace it with our own. As it turns out...

We can't do that because each browser's default select styling creates inconsistencies, but most of all..._**Safari**_.

---

### ❌ Fail: Using the browser's default caret
Removing `appearance: none` to use the browser's default caret yielded these results:
Firefox (close!)
Screenshot 2025-01-23 at 4 06 30 PM

Chrome and Edge (they add space around the selected value that we can't seem to target)
Screenshot 2025-01-23 at 4 07 20 PM

Safari (straight-up doesn't care)
Screenshot 2025-01-23 at 4 06 10 PM

---

### ✅ Pass: Using `background-image` with `background-blend-mode`
So, I ended up taking the approach of keeping `appearance: none` and using a `background-image` with `background-blend-mode`.

**The results are not exact, but they are passable.** Here are some examples:
Default theme:
Screenshot 2025-01-23 at 7 04 12 PM
Screenshot 2025-01-23 at 7 04 19 PM
Screenshot 2025-01-23 at 7 04 33 PM
Screenshot 2025-01-23 at 7 04 27 PM

Awesome theme:
Screenshot 2025-01-23 at 7 04 52 PM
Screenshot 2025-01-23 at 7 04 58 PM
Screenshot 2025-01-23 at 7 04 44 PM
Screenshot 2025-01-23 at 7 05 03 PM

The worst offenders are selects in the Premium theme, where `background-blend-mode` is less effective because the theme sets `--wa-form-control-background-color: transparent`. 

A filled select in dark mode is not especially great:
Screenshot 2025-01-23 at 6 43 30 PM

An outlined select in light mode is meh:
Screenshot 2025-01-23 at 6 43 53 PM

There may be a closer match, but I've exceeded the allotment of time I'd like to spend mixing and matching blend modes for now.

### DIFF
--- a/src/styles/native/select.css
+++ b/src/styles/native/select.css
@@ -73,19 +73,11 @@ select:focus,
 /* Replace native select caret */
 select {
   appearance: none;
-}
-label:has(select) {
-  position: relative;
-  --icon-caret: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" height="16" width="16" viewBox="0 0 512 512"><path d="M233.4 406.6c12.5 12.5 32.8 12.5 45.3 0l192-192c12.5-12.5 12.5-32.8 0-45.3s-32.8-12.5-45.3 0L256 338.7 86.6 169.4c-12.5-12.5-32.8-12.5-45.3 0s-12.5 32.8 0 45.3l192 192z"/></svg>');
 
-  &::after {
-    content: '';
-    background-color: var(--wa-color-neutral-on-quiet);
-    mask: var(--icon-caret) 50% 50% no-repeat;
-    width: 1rem;
-    height: var(--wa-form-control-height);
-    position: absolute;
-    bottom: var(--wa-form-control-border-width);
-    right: var(--wa-space);
-  }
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):
Wow, this is great! I'm so glad the blending mode idea actually panned out!! I hadn't considered including the icon twice, that's clever.

Issue #504, opened 2025-01-13
Remove `<wa-radio-button>` in favor of `<wa-radio>` styles


LeaVerou:
We are definitely all in agreement that we need to drop `<wa-radio-button>`, but it's unclear what is the best API to do so. 
- It seems reasonable that `<wa-radio>` should replace it, with a way to specify that the radio should look like a button, but perhaps the signifier could be on `<wa-radio-group>`, since there is no use case where you'd want some radios styled like buttons and others styled like radios. 
- That said, `<wa-checkbox>` should follow the same API and there are use cases for using single button-like toggles without a grouping element.

Ideas discussed:

1. Lea: We already have the `wa-button` class, we can simply ensure it works well on these components. 
	- Problem: how do you specify it on radio-group for all radios?
	- Problem: would this create the expectation that all `wa-*` classes work on all components?
2. Cory: Another idea is a `type="button"` (or e.g. `like="button"` so that it doesn't conflict with existing `type` attributes) attribute, which could be specifiable on either the radio group or the radio (or checkbox)
3. Lindsay: Another idea could be that radio-group accepts _any_ element child, and does the heavy lifting of applying the right a11y attributes etc.
	- Problem: do we really want an "anything can be anything" mentality? Though buttons are not such a bad idea, see https://lea.verou.me/blog/2022/07/button-group/

lindsaym-fa:
Referencing #21, another common styling choice for radios in addition to classic buttons is a sort of "choice card":
<img width="611" alt="Screenshot 2023-12-14 at 5 08 55 PM" src="https://github.com/shoelace-style/webawesome/assets/126139086/35dd5e30-64c1-4562-ba2c-55514389d35b">

Ideally, the API choice we make here is also well-suited for introducing more varied radio item styles like this one.

One reason I prefer approach 2 for this is that there are situations where the appearance you want for `wa-radio` differs from the default styles you get with our element utility classes like `wa-button`. For example, `wa-button` creates a neutral accent button, which isn't likely what you want for a radio group. Outlined would be preferred, so users would have to specify `class="wa-button wa-outlined"` for every `wa-radio`. Compare that to `like="button"` which could impose those outlined styles by default.

Another complicating factor is the need for a 'checked' state. Our element utility classes generally won't have styles for this built-in. So, we either have to impose general checked styles, which may not be compatible with the applied unchecked styles of the control, or the user has to specify the checked styles themselves, which increases their burden.

[...]

PR #598, opened 2025-01-24
Add breaks between radios in Radio Group examples

lindsaym-fa:
Fixes some unsightliness in the Radio Group docs by adding breaks between `<wa-radio>`s in the examples.

Before:
Screenshot 2025-01-24 at 1 57 12 PM

After:
Screenshot 2025-01-24 at 1 56 53 PM

As a more substantial improvement, I'd love to offer some layout options for `<wa-radio-group>` to arrange the controls within either vertically or horizontally with reasonable default gaps that can be overridden, but that might be best considered following #504.

### DIFF
--- a/docs/docs/components/radio-group.md
+++ b/docs/docs/components/radio-group.md
@@ -7,8 +7,8 @@ icon: radio-group
 
 ```html {.example}
 <wa-radio-group label="Select an option" name="a" value="1">
-  <wa-radio value="1">Option 1</wa-radio>
-  <wa-radio value="2">Option 2</wa-radio>
+  <wa-radio value="1">Option 1</wa-radio><br>
+  <wa-radio value="2">Option 2</wa-radio><br>
   <wa-radio value="3">Option 3</wa-radio>
 </wa-radio-group>
 ```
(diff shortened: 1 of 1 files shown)


claviska:
@lindsaym-fa break do you think we should wrap the default slot with a flex container, add a gap, and add an `orientation` attribute that sets `flex-direction`? Or did you have another idea?

This is fine to merge in the meantime :)

claviska (approved):

LeaVerou:
> As a more substantial improvement, I'd love to offer some layout options for `<wa-radio-group>` to arrange the controls within either vertically or horizontally with reasonable default gaps that can be overridden, but that might be best considered following #504.

Ha, I *just* suggested `orientation` for radio-group in #504 without having seen this! And yes, it should probably default to vertical.

lindsaym-fa:
> do you think we should wrap the default slot with a flex container, add a gap, and add an `orientation` attribute that sets `flex-direction`? Or did you have another idea?

@claviska That's exactly what I was thinking. How we allow folks to override the gap was the only thing stalling me, but now that I've stepped away from it, I realize that targeting `::part(form-control-input)` would probably work just fine.

LeaVerou:
Why not just setting `gap` on radio-group? That seems to be the most straightforward option.

PR #599, opened 2025-01-24
Use 100% width for text field and select labels, fixes #578

### LINKED ISSUE #578: Native select is 100% but labels that wrap them aren't
The native select is currently set to 100% width but, when wrapped with a `<label>` [as shown in the docs](https://backers.webawesome.com/docs/native/select), is incorrectly constrained by the label.

Image

This seems like a relatively simply fix…

```css
/* src/styles/native/select.css */
label:has(select) {
  width: 100%;
  position: relative;
}
```

…except this causes clicks in the "whitespace" within the label to focus the select, which feels buggy.

Image

~~Should we consider a grid-style layout where the label fits the content?~~ (This won't work without a container)

Adding a medium priority to this since it shows up in the docs like this.

lindsaym-fa:
This change:

- Allows labels containing a `<select>`, `<textarea>`, or `<input>` of a specific type to use 100% width
- Sets `display: inline-block` on labels for `<wa-input>`, `<wa-select>`, and `<wa-textarea>` to fix an issue where clicking the adjacent whitespace applied focus to the control

---

The issue described for `<select>` in #578 also affected `<textarea>` and `<input>`, so this PR addresses all three.

Before:
Screenshot 2025-01-24 at 4 16 15 PM
Screenshot 2025-01-24 at 4 16 52 PM
Screenshot 2025-01-24 at 4 17 22 PM

After:
Screenshot 2025-01-24 at 4 16 25 PM
Screenshot 2025-01-24 at 4 17 05 PM
Screenshot 2025-01-24 at 4 17 33 PM

I couldn't find a way to prevent the whitespace adjacent to the label from being clickable when using an implicit label. The best solution I could come up with was adding `pointer-events: none` to the label, then `pointer-events: auto` to the control, but that also removes clickability from the label text, which is an important UX feature — so I didn't implement this. Folks who won't find the clickable whitespace advantageous for their users can always use explicit labels with the `for` attribute.

Additionally, the whitespace was also clickable in WA components. This was not the case in Shoelace, and I'm not totally sure this was an intentional change, so this PR fixes that.

### DIFF
--- a/src/styles/native/forms.css
+++ b/src/styles/native/forms.css
@@ -5,11 +5,30 @@ label {
   font-weight: var(--wa-form-control-label-font-weight);
   line-height: var(--wa-form-control-label-line-height);
   position: relative;
-}
 
-label + :is(input:not([type='checkbox'], [type='radio']), textarea, select),
-label > :is(input:not([type='checkbox'], [type='radio']), textarea, select) {
-  margin-block-start: var(--wa-space-xs);
+  &:has(
+      input:not(
+          [type='button'],
+          [type='checkbox'],
+          [type='color'],
+          [type='file'],
+          [type='hidden'],
+          [type='image'],
+          [type='radio'],
+          [type='range'],
+          [type='reset'],
+          [type='submit']
+        ),
+      textarea,
+      select
+    ) {
+    width: 100%;
+  }
+
+  & + :is(input:not([type='checkbox'], [type='radio']), textarea, select),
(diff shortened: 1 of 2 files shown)


claviska:
> Additionally, the whitespace was also clickable in WA components. This was not the case in Shoelace, and I'm not totally sure this was an intentional change, so this PR fixes that.

At least we're consistent now, but this was definitely a regression.

claviska (approved):

PR #601, opened 2025-01-25
Ensure change events bubble

### LINKED ISSUE #597: New native events that should bubble don't bubble
E.g. try theme remixing, which is currently broken. It listens to `change` on an ancestor. Changing that to `input` doesn't fix it, but adding `{capture: true}` does, which means the event isn't bubbling.

Indeed, printing out `event.bubbles` prints out `false`.

I believe specifying `bubbles: true` explicitly when dispatching them would fix this.

claviska:
Fixes #597

### DIFF
--- a/src/internal/webawesome-element.ts
+++ b/src/internal/webawesome-element.ts
@@ -160,14 +160,18 @@ export default class WebAwesomeElement extends LitElement {
   }
 
   /**
-   * Given a native event, this function ensures it's composed and, if not, dispatches it again as a composed event.
-   * This is useful for relaying native events such as `change`, which will otherwise not be retargeted. It is safe,
-   * albeit sloppy, to call this on composed events, as it will no-op.
+   * Given a native event, this function cancels it and dispatches it again from the host element using the desired
+   * event options.
    */
-  dispatchComposedEvent(event: Event) {
-    if (!event.composed) {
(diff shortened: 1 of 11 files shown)


src/internal/webawesome-element.ts
```diff
@@ -160,14 +160,18 @@ export default class WebAwesomeElement extends LitElement {
...
-    }
+  relayNativeEvent(event: Event, eventOptions?: EventInit) {
+    event.stopImmediatePropagation();
+
+    this.dispatchEvent(
+      new (event.constructor as typeof Event)(event.type, {
+        ...event,
+        ...eventOptions,
+      }),
+    );
```
claviska, line 174: I kept this as sugar. I don't love how we need to remember to explicitly add bubbles/composed to change events...it feels like a footgun. Perhaps a long term strategy would bake in default for certain event types, but this works in the meantime.

LeaVerou: Agreed, I think it would be better to encode that here (or anywhere else that is central), i.e. have an object literal with the default `bubbles` state and read it from there. But meanwhile, this works.

LeaVerou (approved):

LeaVerou:
Merging this in the hopes it fixes #605

PR #603, opened 2025-01-25
Fix button appearance docs

LeaVerou:
- List missing appearances
- Harmonize component & native docs (same order of appearances, same button text)


lindsaym-fa (approved):
Looks good!

PR #617, opened 2025-01-27
Add Visual Test Suite

lindsaym-fa:
This PR adds a public-facing docs page with a suite a of visual tests to help us do the following:

- Ensure that our native styles and their component counterparts produce identical results
- Ensure that our CSS utility classes and their attribute counterparts produce identical results
- Evaluate the design consistency and harmony of our entire library and individual themes
- Identify missing selectors in themes with custom styles that cause inconsistencies between native styles vs. components and utility classes vs. attributes

These tests are not automated. All test cases are intended to be reviewed manually as styling features are added or changed.

By allowing all Web Awesome users to access this page, we can empower developers to test their own, custom themes and open up a democratic process for identifying, reporting, and, someday, contributing fixes to visual bugs.


claviska (approved):
Great work. The only things I've noticed are:

1. Phones scroll horizontally (how do you even make such a thing responsive, and who would even want to view this on a mobile phone…)
2. It's available to the public (might be intentional, but let me know if not)

Otherwise, 🚀

lindsaym-fa:
Thanks, @claviska!
1. Good note about the phone UX! I pushed an update to make each table scrollable, which feels a lot nicer than having the entire page scroll horizontally.
2. This is intentionally public, but definitely let me know if that makes you nervous. I think transparency here is a good thing, and it could be a great tool for people who are rolling their own themes and want to make sure they support all manners of styling.

PR #618, opened 2025-01-27
Improve theme remixing UI

### LINKED ISSUE #574: [Docs] Improve theme remixing UI
Currently these add quite a lot of clutter to each theme page:

Image

Additionally, we do not generate copyable code examples for these remixed themes, people just have to figure it out from the general docs, which is suboptimal:

Image

I propose the following:

- A "Remixed Theme" page in the docs that takes URL params and populates these (and changing the selects updates the URL so that you always have a permalink) and shows complete code snippets.
- Only a "Remix" button on the theme pages that takes you to that page, with the base theme pre-populated.
- In the future, we could even let people "Save" these remixed themes into the sidebar (all we need is saving them in `localStorage`)
- The remixed theme has a "Pro" badge if any of its constituent themes are Pro themes.

In the future, as the themer evolves, this could also seamlessly evolve to allow more customization.

LeaVerou:
Closes #574 

- [x] Move theme remixing to separate page and declutter core theme page — now the only mention of remixing on theme pages is a small button in the demo preview
- [x] Implement permalinks so that remixed themes can be shared and bookmarked
- [x] Show code examples for remixed theme (updating these was a b*itch — having that client-side Prism would have helped _immensely_ here)

The code is not great, but we can improve on that later.

Not done in this iteration:
- Remove no-op options from selects:
	- Remove base theme from color & typography selects, 
	- Remove its color palette from the palette select
- Button to set remixed theme as your docs theme
- Should the "Remix" button create a random mix? Currently it feels like you’re taken back to the same page because the theme preview is identical.

### DIFF
--- a/docs/docs/themes/remixed/index.njk
+++ b/docs/docs/themes/remixed/index.njk
@@ -1,0 +1,84 @@
+---
+title: Remixed
+description: TODO
+isPro: true
+tags: pro
+---
+<link rel="stylesheet" href="{{ page.url }}/style.css">
+{% block header %}
+<script>
+globalThis.defaultPalettes = {
+  {% for theme in collections.theme | sort -%}
+    "{{ theme.fileSlug }}": "{{ theme.data.palette }}",
+  {%- endfor %}
+};
+</script>
+<p id="mix_and_match" class="wa-gap-m">
+  {# <strong>
+    <wa-icon name="merge" slot="prefix"></wa-icon>
+    Remix
+    <wa-icon-button href="#remixing" name="circle-question" slot="suffix" variant="regular" label="How to use?"></wa-icon-button>
+  </strong> #}
+  <wa-select name="base" label="Base:" size="small" value="default">
+    <wa-icon name="brush" slot="prefix" variant="regular"></wa-icon>
+    {% for theme in collections.theme | sort %}
+      {% if theme.fileSlug !== page.fileSlug %}
+      <wa-option value="{{ theme.fileSlug }}" data-palette="{{ theme.data.palette }}">{{ theme.data.title }}</wa-option>
+      {% endif %}
+    {% endfor %}
+  </wa-select>
+
+  <wa-select name="palette" label="Palette" size="small">
+    <wa-icon name="swatchbook" slot="prefix" variant="regular"></wa-icon>
+    <wa-option value="">(Theme default)</wa-option>
+    <wa-divider></wa-divider>
+    {% for p in collections.palette | sort %}
+      <wa-option value="{{ p.fileSlug }}">{{ p.data.title }}</wa-option>
+    {% endfor %}
+  </wa-select>
+
+  <wa-select name="color" label="Colors from…" size="small" value="">
+    <wa-icon name="palette" slot="prefix" variant="regular"></wa-icon>
+    <wa-option value="">(Theme default)</wa-option>
+    <wa-divider></wa-divider>
+    {% for theme in collections.theme | sort %}
+      <wa-option value="{{ theme.fileSlug }}">{{ theme.data.title }}</wa-option>
+    {% endfor %}
+  </wa-select>
+
+  <wa-select name="typography" label="Typography from…" size="small" value="">
+    <wa-icon name="font-case" slot="prefix"></wa-icon>
+    <wa-option value="">(Theme default)</wa-option>
+    <wa-divider></wa-divider>
+    {% for theme in collections.theme | sort %}
+      <wa-option value="{{ theme.fileSlug }}">{{ theme.data.title }}</wa-option>
+    {% endfor %}
+  </wa-select>
+</p>
+<script src="{{ page.url }}index.js" type="module"></script>
+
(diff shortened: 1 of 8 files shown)


docs/docs/themes/remixed/index.njk
```diff
@@ -0,0 +1,84 @@
...
+  </wa-select>
+
+  <wa-select name="typography" label="Typography from…" size="small" value="">
+    <wa-icon name="font-case" slot="prefix"></wa-icon>
+    <wa-option value="">(Theme default)</wa-option>
+    <wa-divider></wa-divider>
+    {% for theme in collections.theme | sort %}
+      <wa-option value="{{ theme.fileSlug }}">{{ theme.data.title }}</wa-option>
+    {% endfor %}
+  </wa-select>
```
lindsaym-fa, line 56: These should probably exclude "Remixed" from their options.
Screenshot 2025-01-28 at 5 26 36 PM

docs/docs/themes/remixed/index.njk
```diff
@@ -0,0 +1,84 @@
...
+</p>
+<script src="{{ page.url }}index.js" type="module"></script>
+
+<wa-callout variant="warning">
+  <wa-icon slot="icon" name="triangle-exclamation" variant="regular"></wa-icon>
+  Please note that not all combinations will look good — once you’re mixing and matching, you’re on your own!
+</wa-callout>
```
lindsaym-fa, line 63: What do we want to accomplish with this warning? It seems a bit distracting and too prominent given most combinations actually do look good! 

If we're trying to make sure folks don't assume we've given our stamp of approval to every combination in order to discourage folks from shipping poor ones, then this warning doesn't do much for us. How would they know which combinations are poor?

If we're trying to excuse poor combinations for folks that have the design sense to notice them, I don't necessarily want to absolve ourselves of responsibility — if something doesn't work well, it's worth investigating. Sometimes, of course, the design choices may just not be compatible. Often, though, there are things we can fix.

docs/docs/themes/remixed/index.njk
```diff
@@ -0,0 +1,84 @@
...
+You can import your remixed theme by importing the individual theme files from the Web Awesome CDN.
+{# Note: If you change the order here, you need to update index.js too #}
+{% set stylesheets = [
+  'styles/themes/default.css',
+  'styles/color/default.css',
+  'styles/themes/default/color.css',
+  'styles/themes/default/typography.css'
+] %}
+{% set stylesheet = 'styles/themes/default/color.css' %}
+{% include 'import-stylesheet-code.md.njk' %}
```
lindsaym-fa, line 80: No matter what is selected as the Base theme, the snippet always seems to show `default.css`.
Screenshot 2025-01-28 at 5 55 51 PM

docs/docs/themes/remixed/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+  wa-select::part(label) {
+    margin-block-end: 0;
+  }
+
+  wa-select[value='']::part(display-input),
+  wa-option[value=''] {
+    font-style: italic;
+    color: var(--wa-color-text-quiet);
+  }
```
lindsaym-fa, line 17: I'll file a separate bug for this, but fyi these styles aren't working right now. It looks like `value` doesn't update when a value is selected, so it's always null.
Screenshot 2025-01-28 at 5 59 05 PM

lindsaym-fa (commented):
Offering a code snippet to use the remixed theme is a great improvement and well-worth the effort, I think!

I like the idea of theme remixing living on its own page. As it stands, though, I **completely** missed that this was a different page than the theme I was viewing the first several times I tested this — I thought it merely revealed the fields to remix the theme! That ended up being quite confusing and didn't feel as intuitive as the iteration currently available to backers.

A couple of recommendations that might help:
- **We could use a more appropriate H1 for this page.** 
    - Navigating to a new page that has the exact same H1 and header section as the page we were just on is misleading and disorienting. 
    - Because I see "Remixed" in the sidebar and index right alongside other themes, I assume it's just a standalone theme. A "**Remix a Theme**" H1 or similar with a description of what a remixed theme is and the possibilities it enables would be super helpful. 
    - I'm not sure how straightforward it would be to change the breadcrumbs and H1 in the showcase just for this page, so that's a complicating factor. I do still appreciate having the base theme and its description shown here, if we're just able to present it differently.
- **We could move 'Remix' outside of the showcase.** 
    - Perhaps a link below the showcase that reads, "Remix this theme" with the same `merge` icon. A true link instead of a button-like link is a better signifier that we're navigating to a new page rather than operating within the contained showcase context.
    - This also has the benefit of eliminating the problem where pressing 'Remix' on the Remixed page just returns us to the same page.

docs/docs/themes/remixed/index.njk
LeaVerou: Agreed, this was largely leftover from the current iteration. I'll remove it.

LeaVerou:
Closing this in favor of #724.

PR #635, opened 2025-01-28
Add orientation to radio group

### LINKED ISSUE #613: [Radio-group][Checkbox-group] Add `orientation` attribute
This has been suggested in several places:
- https://github.com/shoelace-style/webawesome/pull/598
- https://github.com/shoelace-style/webawesome/issues/504#issuecomment-2613485607

The default would be `vertical`. The attribute would just set `display: flex` and `flex-direction` accordingly. 

TBD: 
- What is the layout for label and hint in horizontal orientation?  It seems that the hint should definitely be on its own line, but he label could go either way. Slight preference for the label on its own line as well, simply because that’s how other controls work.
- Presumably we want some kind of guardrail to ensure that the horizontal layout won't overflow. `flex-wrap: wrap` seems easiest, but suboptimal, since checkboxes/radios won't be aligned, but may be okay for an initial MVP. We may want an auto-fill grid eventually though.

Setting `gap` on the host should work to override the default gap (which would allow our gap utility classes to work as well).
Looking at the `radio-group` markup, we would need to basically propagate `gap` to the `<fieldset>`. To do that, we can take a leaf out of #207’s book and
1. Set `display: contents` on `:host`.
2. Set `all: inherit; display: inline-flex;` on the `fieldset`

claviska:
Fixes #613. Note that the [suggested approach](https://github.com/shoelace-style/webawesome/issues/613#issue-2812118123) didn't work because the fieldset houses the label and hint, so styles were applied to the slot.

The new orientation attribute works with radio buttons as well. This also fixes a misspelled listener in the radio group docs.

Note: it was a bit daunting to work through the merged component + native CSS. It would be nice if we could get back to a simpler syntax, if at all possible :)

~~During this PR, I noticed a timing issue where Radio Buttons don't always show their selected value on refresh. I'm looking into that now.~~

Update: this was caused by the radio button not being registered when the radio group syncs items. Since we're going to be removing Radio Button, the quick fix was to import it and make it a dependency. This is the pattern we use for parent + child components.

This means, temporarily, we'll have radio and radio buttons imported by radio group automatically BUT this was cleaner than working around the sync logic and, since Radio Button is getting replaced soon, I'm fine with it.

I also removed the `@dependency wa-button-group` and added the `@dependency radio` since that was missing.

### DIFF
--- a/docs/docs/components/radio-button.md
+++ b/docs/docs/components/radio-button.md
@@ -9,7 +9,7 @@ icon: radio-group
 Radio buttons are designed to be used with [radio groups](/docs/components/radio-group). When a radio button has focus, the arrow keys can be used to change the selected option just like standard radio controls.
 
 ```html {.example}
-<wa-radio-group label="Select an option" name="a" value="1">
+<wa-radio-group label="Select an option" orientation="horizontal"  name="a" value="1">
   <wa-radio-button value="1">Option 1</wa-radio-button>
   <wa-radio-button value="2">Option 2</wa-radio-button>
   <wa-radio-button value="3">Option 3</wa-radio-button>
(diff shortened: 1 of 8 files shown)


lindsaym-fa (approved):
What a gift. ✨ Thanks, @claviska!

PR #638, opened 2025-01-29
Revise docs to clarify installation guidance

lindsaym-fa:
Web Awesome has _a lot_ of features. And while many of them are optional, _not_ including these features can yield confusing, seemingly broken results ([like missing `/styles/webawesome.css` did for this user](https://secure.helpscout.net/conversation/2816261097/166579)). 

The goal of this PR is to make our installation guidance more clear so that it's easier to hit the ground running. A few key changes to the docs that I hope accomplish this:

1. I reorganized our [Installation](http://localhost:4000/docs/installation/) page to focus on what's easiest and common, and divert less attention to more advanced setup, which involves:
    1. Including `/styles/webawesome.css` in the autoloader code snippet
    2. Renaming "Autoloading via CDN (Easiest)" to "Quick Start (Autoloading via CDN)" and revising its content
    3. Tucking alternative installation methods or non-standard setup under an "Advanced Setup" heading
2. I revised our [Customizing](http://localhost:4000/docs/customizing/) page and added guidance for [Native Elements](http://localhost:4000/docs/customizing/#native-elements) and [Style Utilities](http://localhost:4000/docs/customizing/#style-utilities)
3. I added links to relevant installation guidance to the intros our index pages (e.g., [Native Styles](http://localhost:4000/docs/native/), [Style Utilities](http://localhost:4000/docs/utilities/), and [Layout](http://localhost:4000/docs/layout/))

If any of these changes **_do not_** help clarify our installation guidance, let me know!

### DIFF
--- a/docs/docs/customizing.md
+++ b/docs/docs/customizing.md
@@ -16,7 +16,7 @@ To use a theme, simply add a link to the theme's stylesheet to the `<head>` of y
 <link rel="stylesheet" href="{% cdnUrl 'styles/themes/awesome.css' %}" />
 ```
 
-You can [customize any theme](/docs/themes/creating) just with CSS — no preprocessor required. All design tokens are prefixed with `--wa-` to avoid collisions with other libraries or your own custom properties. Simply override any design token in your own stylesheet by scoping your styles to `:root`, `:host`, the class for the specific theme you want to override (if needed), and the class for the relevant color scheme (if needed). Here's an example that changes the default brand color to violet in light mode:
+You can [customize any theme](/docs/themes/creating) just with CSS — no preprocessor required. All design tokens are prefixed with `--wa-` to avoid collisions with other libraries or your own custom properties. Simply override any design token in your own stylesheet by scoping your styles to `:where(:root)`, `:host`, the class for the specific theme you want to override (if needed), and the class for the relevant color scheme (if needed). Here's an example that changes the default brand color to violet in light mode:
 
 ```css
 :where(:root),
@@ -35,11 +35,15 @@ You can [customize any theme](/docs/themes/creating) just with CSS — no prepro
 }
 ```
 
+:::info
+Wrapping the `:root` selector in `:where()` gives this selector 0 specificity. This allows us to define our design tokens' default values while ensuring they can be cleanly overridden as needed.
+:::
+
 For a complete list of all custom properties used for theming, refer to `src/styles/themes/default.css` in the project's source code.
 
 ## Components
 
-While a theme offers a high-level way to customize the library, components offer different hooks as a low-level way to customize them individually.
+While themes offer a high-level way to customize the library, components offer different hooks as a low-level way to customize them individually.
 
 Web Awesome components use a [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) to encapsulate their styles and behaviors. As a result, you can't simply target their internals with the usual CSS selectors. Instead, components expose a set of custom properties and CSS parts that can be targeted to customize their appearance.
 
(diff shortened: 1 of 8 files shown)


docs/docs/customizing.md
```diff
@@ -1,24 +1,28 @@
...
 
-To customize a theme, simply override any of these custom properties in your own stylesheet by scoping your styles to `:root`, `:host`, and, if needed, the class for the specific theme you want to override. Here's an example that changes the default brand color (blue) to violet in the light theme using existing [literal colors](/docs/tokens/color/#literal-colors).
+```html
+<link rel="stylesheet" href="{% cdnUrl 'styles/themes/awesome.css' %}" />
+```
+
+You can [customize any theme](/docs/themes/creating) just with CSS — no preprocessor required. All design tokens are prefixed with `--wa-` to avoid collisions with other libraries or your own custom properties. Simply override any design token in your own stylesheet by scoping your styles to `:root`, `:host`, the class for the specific theme you want to override (if needed), and the class for the relevant color scheme (if needed). Here's an example that changes the default brand color to violet in light mode:
 
 ```css
 :where(:root),
```
claviska, line 22: Not part of this PR, but why is this wrapped in `:where()`?

docs/docs/customizing.md
```diff
@@ -31,11 +35,11 @@ To customize a theme, simply override any of these custom properties in your own
 }
 ```
 
-For more examples and further guidance, refer to [Themes](/docs/themes) and the Theming section of this documentation. For a complete list of all custom properties used for theming, refer to `src/styles/themes/default.css` in the project's source code.
+For a complete list of all custom properties used for theming, refer to `src/styles/themes/default.css` in the project's source code.
 
 ## Components
 
-Whereas a theme offers a high-level way to customize the library, components offer different hooks as a low-level way to customize them individually.
+While a theme offers a high-level way to customize the library, components offer different hooks as a low-level way to customize them individually.
```
claviska:
```suggestion
While a themes offer a high-level way to customize the library, components offer different hooks as a low-level way to customize them individually.
```

docs/docs/customizing.md
```diff
@@ -122,3 +126,52 @@ CSS parts have a few important advantages:
 - It encourages us to think more about how components are designed and how customizations should be allowed before users can take advantage of them. Once we opt a part into the component's API, it's guaranteed to be supported and can't be removed until a major version of the library is released.
 
 Most (but not all) components expose parts. You can find them in each component's API documentation under the "CSS Parts" section.
+
+## Native Elements
+
+If you're using [native styles](/docs/native), any custom styles added for a component should also target the corresponding native element. In general, the same styles you declare for components will work just the same to style their native counterparts.
```
claviska, line 136:
```suggestion
If you're using [native styles](/docs/native), any custom styles added for a component will also target the corresponding native element. In general, the same styles you declare for components will work just the same to style their native counterparts.
```

docs/docs/customizing.md
```diff
@@ -122,3 +126,52 @@ CSS parts have a few important advantages:
...
+<style>
+  wa-checkbox,
+  input[type="checkbox"] {
+    --background-color-checked: hotpink;
+    --border-color-checked: hotpink;
+    --border-width: 3px;
+    --checked-icon-color: lavenderblush;
+  }
+</style>
+```
```
claviska: I know we're talking about tokens in this section, but is it worth emphasizing that only customizations made via tokens will be applied to components and native elements (as opposed to parts)?

claviska (approved):
Just a few comments, but nothing blocking!

docs/docs/customizing.md
lindsaym-fa: I added an info callout because I'm sure other folks will have the same question. `:where(:root)` ensures that tokens defined on the root have 0 specificity so they can be overridden cleanly. We don't want tokens on the `:root` to end up taking precedence over specific classes like `.wa-dark`.

docs/docs/customizing.md
lindsaym-fa: I added an example to show using parts to style both WA checkboxes and native checkboxes with standard CSS properties.

docs/docs/customizing.md
lindsaym-fa: Since this is a recommendation, I think this should stay as-is. Custom styles authored by our users _won't_ also target the corresponding native elements _unless_ they specifically write them that way, which is what this sentence is meant to convey.

PR #646, opened 2025-01-31
First stab at `with-*` → `ssr-slots`

### LINKED ISSUE #261: Combine all `with-*` attributes into a single attribute (`ssr-slots`)
Currently we have 8 different `with-*` attributes:

1. `with-label` (7 components)
1. `with-help-text` (7 components)
1. `with-header` (3 components)
1. `with-footer` (3 components)
1. `with-image` (`<wa-card>`)
1. `with-submenu` (`<wa-menu-item>`)
1. `with-prefix` (`<wa-radio-button>`)
1. `with-suffix` (`<wa-radio-button>`)

These are only required for SSR (and possibly React?), to communicate which slots are filled. @KonnorRogers could elaborate on the details of why these are needed and why for these specific slots.

I wonder, could we reduce the API surface of this wart and make it more predictable by combining these into a single `with` attribute that takes a space-separated list of slot names? Then we only need to document this once and authors have to only learn this once. The only variable is which slots need this kind of hint, which we could even include in the Slots table. Additionally, it aids readability before it ensures they are all in one place, rather than scattered among other attributes.

Before & after:
```html
<wa-dialog label="Dialog" with-header with-footer class="dialog-overview">
<wa-dialog label="Dialog" with="header footer" class="dialog-overview">
<wa-dialog label="Dialog" has="header footer" class="dialog-overview"> <!-- alt -->
```

### LINKED ISSUE #641: `with-*` attributes not added automatically even when not in SSR
E.g. example for `<wa-card>` https://codepen.io/leaverou/pen/RNbdZPj/854993e3712d31ba36740ca2cb0112b6?editors=1000

LeaVerou:
This PR converts the existing `with-*` attributes into a single `ssr-slots` attribute and thus fixes #261. It also fixes #641.

- [x] Infrastructure in base class
- [x] Sample implementation in Card
- [x] Docs

To-Do:

- [ ] Color Picker
- [ ] Input
- [ ] Radio button
- [ ] Radio group
- [ ] Select
- [ ] Slider
- [ ] Textarea
- [ ] Dialog
- [ ] Drawer
- [ ] Menu item
- [ ] Radio button

Classes declare a static `SSR_SLOTS` array with the names of the slots to participate in `ssr-slots`, and use `@slotchange={this.slotUpdate}` events to update a `this.hasSlotted` data structure and the `ssrSlots` prop. So most of the logic lives in the base class. 

I do wish we could have mixins for these types of common patterns though, so we didn't have to decide between overloading the base class with something that is not relevant to every component and the noisiness of repeating the pattern on every class (even with utils).

We could make things even easier for the subclasses by listening to the `slotchange` event via delegation, but I was worried about the performance of listening to every single slot change, whereas opting slots in seems more prudent albeit more noisy. I may be overly cautious though.

Before I work on the other components, I was hoping @claviska @KonnorRogers might be able to take a look and let me know what you think about the overall approach.

### DIFF
--- a/docs/docs/components/card.md
+++ b/docs/docs/components/card.md
@@ -6,7 +6,7 @@ icon: card
 ---
 
 ```html {.example}
-<wa-card with-image with-footer class="card-overview">
+<wa-card class="card-overview">
   <img
     slot="image"
     src="https://images.unsplash.com/photo-1559209172-0ff8f6d49ff7?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=500&q=80"
@@ -61,10 +61,10 @@ Basic cards aren't very exciting, but they can display any content you want them
 ### Card with Header
 
 Headers can be used to display titles and more.
-If using SSR, you need to also use the `with-header` attribute to add a header to the card (if not, it is added automatically).
+If using SSR, you need to also use the `ssr-slots` attribute to add a header to the card (if not, it is added automatically).
 
 ```html {.example}
-<wa-card with-header class="card-header">
+<wa-card ssr-slots="header" class="card-header">
   <div slot="header">
     Header Title
     <wa-icon-button name="gear" variant="solid" label="Settings"></wa-icon-button>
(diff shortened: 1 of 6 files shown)


claviska:
I'm doing some PR cleanup :)

This one seems stale and there's a good chunk of work left. I also have some new concerns [I expressed here](https://github.com/shoelace-style/webawesome/issues/261#issuecomment-2864104420).

I'm going to close this one for now to get it off the "PR backlog," but we can revisit it once the npm repos are available and users start asking for SSR. I'd like their feedback as well.

PR #665, opened 2025-02-03
Rename `gray` to `neutral`, fixes #660

### LINKED ISSUE #660: Rename `gray` to `neutral`
Per discussion with @lindsaym-fa

lindsaym-fa:
Renames `gray` to `neutral` to describe a broader range of possible hues and chromas.

### DIFF
--- a/docs/_data/hues.json
+++ b/docs/_data/hues.json
@@ -1,1 +1,1 @@
-["red", "yellow", "green", "teal", "blue", "indigo", "violet", "gray"]
+["red", "yellow", "green", "teal", "blue", "indigo", "violet", "neutral"]
--- a/docs/assets/styles/code-highlighter.css
+++ b/docs/assets/styles/code-highlighter.css
@@ -1,9 +1,9 @@
 pre {
-  background-color: var(--wa-color-gray-20);
+  background-color: var(--wa-color-neutral-20);
   color: white;
 
   /* Ensures a discernible background color in dark mode
-   * Useful for themes that use gray-20 as --wa-color-surface-default  */
+   * Useful for themes that use neutral-20 as --wa-color-surface-default  */
   .wa-dark & {
     background-color: var(--wa-color-surface-lowered);
   }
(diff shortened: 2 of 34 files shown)


LeaVerou:
Before we merge this, I'd love to hear what you think in the question in #669:

> Known unknown around neutrals: In past discussions we've discussed potentially mixing in the accent color or its complementary for warmer and cooler, but this is a property of the **theme**, not the color palette. Should we pick a predefined hue in the palette to mix with (e.g. `blue` and `red`) instead? Or should the exact neutral become part of the theme?

lindsaym-fa:
Ooo! Forgive me if I'm not 100% on the right track here, but this is where the prompt took me:

The role of a _color palette_ is to establish a set of harmonious hues that you can use however you'd like in a theme. The role of a _theme_ is to use those colors in a way that establishes a certain look or communicates a certain message — or, abstractly, gives a certain ✨ vibe ✨. So: Do neutrals pertain more to color harmony or vibe?

I'd say the latter. In the same way we could theoretically choose to use `orange` instead of `yellow` for our `warning` colors, we should also be able to choose to use, say, `pink` instead of `gray` for our `neutral` colors. With that in mind, I'd argue that skewing warmer or cooler is actually a property of the color palette's `gray`, not necessarily the theme's `neutral`. The color palette's job is to offer compatible hues; the theme's job is to decide if/how they're used.

That leads me to think that we ought to keep `gray` in these palettes. And, as we discussed having aliases for `--wa-color-brand-95`, etc. that map to a literal hue, perhaps we should have the same for `--wa-color-neutral-95`, etc.

LeaVerou:
I think I'm now leaning the same way as well, but how are you proposing we generate `gray`? What color do we mix in if we don't know what the brand color will be?
We could also leave it up to user selection…

lindsaym-fa:
If we don't know what the brand color is at the point we're generating grays, I say we make them cooler/mix in `blue` by default. Cooler grays seem to be more common than warmer ones.

LeaVerou:
Sounds good. I’ll close the issue then!

Issue #439, opened 2025-01-07
Utility classes for common container styles


LeaVerou:
In the spirit of "components are for things that need either functionality or additional HTML", people should not have to use a component just to get a certain container style. Additionally, these could be useful for making certain elements look like other elements.

- [x] `.wa-button`
- [x] `.wa-text-field`
- [ ] `.wa-badge`
- [ ] `.wa-tag`
- [x] `.wa-callout`

Unsure about:
- [ ] `.wa-card`
- [ ] `.wa-tooltip`
- [ ] `.wa-dialog`

**Edit:** High Priority now only because now we inadvertently announced it: https://github.com/shoelace-style/webawesome-alpha/issues/153#issuecomment-2607733815 🤦🏽‍♀️

Thankfully, it's pretty low effort.

PR #682, opened 2025-02-04
[WIP] Add `size` to badge + `.wa-badge` native styles

### LINKED ISSUE #676: [Badge] Add `size`
@lindsaym-fa while taking a stab at it (see #682), I noticed several things I wanted to bring to your attention:

Seems like we should be doing something clever here to use a size N levels down from the current one, rather than introducing a cutoff, which also creates issues like #278 

```css
font-size: max(var(--wa-size), 0.75em);
```

Should this be using space tokens?
```css
padding: 0.375em 0.625em;
```

LeaVerou:
Ran out of time but it’s a starting point.

Fixes #676 & makes progress on #439

### DIFF
--- a/docs/docs/components/badge.md
+++ b/docs/docs/components/badge.md
@@ -77,6 +77,16 @@ Use the `pill` attribute to give badges rounded edges.
 <wa-badge variant="danger" pill>Danger</wa-badge>
 ```
 
+### Sizes
+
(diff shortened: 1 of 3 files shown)


claviska:
Closing as not planned. Let's re-evaluate this one to ensure we're all on the same page. I'll notify the user who suggested this, as they can already change badge sizes so it's not a blocker for them.

Issue #486, opened 2025-01-10
RFC: FOUCE solution


claviska:
We got our [first FOUCE complaint](https://github.com/shoelace-style/webawesome-alpha/discussions/156) and we need to have a good answer for it. Here's my suggestion for some helpful utilities to make FOUCE less of a problem when using the autoloader (i.e. not going the painful path of SSR).

It's designed to be as easy as possible for users while still being opt-in. Each part serves a different purpose, so think of this as a collection of FOUCE tools.

### Add an `allDefined()` utility

A common problem is users load components with the autoloader and try to access properties or methods on them right away, often before the element has had a chance to register.

For this, we can add an `allDefined()` utility that waits for all unregistered WA elements on the page to be defined. The simplest usage is as follows. I expect 99% of users to do it this way.

```ts
import { allDefined } from '/dist/webawesome.js';

await allDefined();

// at this point, all WA elements are registered and ready, just like awaiting 
// customElements.whenDefined() for each one, except even better since 
// it's 1) way more convenient and 2) also waits a cycle for the first Lit 
// update to complete
```

For advanced use cases, the function can be configured to work with any custom element, or even a completely differently library, or even multiple libraries!

```ts
// Advanced usage
await allDefined({
  // Wait for all `<wa-*>` and `<my-*>` elements on the page
  match: tagName => tagName.startsWith('wa-') || tagName.startsWith('my-'),

  // Also wait for the following one-offs `<foo-element>` and `<bar-element>`
  additionalElements: ['foo-element', 'bar-element'],

  // Look in the current document (can also be a shadow root)
  root: document
});
```

This function has a number of benefits:

- Users can await it once instead of calling `customElements.whenDefined()` for _every single component_, allowing them to work with methods/properties without worrying about timing issues
- Users can await it to show a loader if they want to (SPA style)
- Users can configure it to wait for third-party custom elements (advanced use case)

Note:`additionalElements` allows for cases where you need to wait for elements that aren't discoverable through DOM traversal at the time `allDefined()` is called. I expect many users won't need it, but it's a convenient hook if you bootstrap your app by importing WA along with your own custom elements manually and you want it to play nice with `allDefined()`.

<details>
  <summary>Show prototype</summary>

```ts
interface AllDefinedOptions {
  /**
   * A callback that accepts a custom element tag name and returns `true` if the custom element should be defined before
   * resolving or `false` to skip it. The tag name is always in lowercase.
   */
  match: (tagName: string) => boolean;

[...]

PR #686, opened 2025-02-04
Add FOUCE utilities

claviska:
This is outside my split work, but I wanted to get it in to have an answer to what I expect to be one of the biggest problems autoloader/CDN users will have.

- Adds support for the `wa-cloak` class when using the autoloader. This eliminates FOUCE on the _initial load_ unless it takes more than two seconds to load.
- Adds the `wa-discovery-complete` class to the autoloader. This is used internally, but it can also be used by consumers to detect when autoloading is complete.
- Adds a utility for Turbo users to eliminate FOUCE when navigating between pages (I disabled SSR and tested this in our docs and it works great — try it out!)
- Adds this info to the appropriate sections of the docs
- Implements the `allDefined()` utility as mentioned in #486

---

- Related issue: #486
- Related support question: https://github.com/shoelace-style/webawesome-alpha/discussions/169


docs/docs/installation.md
```diff
@@ -37,8 +37,30 @@ This snippet includes three parts:
 
 Now you can [start using Web Awesome!](/docs/usage)
 
-:::info
-While convenient, autoloading may lead to a [Flash of Undefined Custom Elements](https://www.abeautifulsite.net/posts/flash-of-undefined-custom-elements/). The linked article describes some ways to alleviate it.
+### Reducing FOUCE
+
+While convenient, autoloading can lead to a [Flash of Undefined Custom Elements](https://www.abeautifulsite.net/posts/flash-of-undefined-custom-elements/). To prevent this, you can add the `wa-reduce-fouce` class to any element on the page. Make sure to include the [FOUCE style utility](/docs/utilities/fouce/#opting-in) to get the corresponding CSS.
```
LeaVerou:
```suggestion
While convenient, autoloading can lead to a [Flash of Undefined Custom Elements](https://www.abeautifulsite.net/posts/flash-of-undefined-custom-elements/).
The [FOUCE style utility](/docs/utilities/fouce/#opting-in) (which is automatically applied if you use the Web Awesome utilities) takes care of hiding custom elements until they and their contents have been registered, up to a maximum of two seconds (configurable).

In many cases, this is not enough, and you may wish to hide a broader wrapper element or even the entire page until all WA elements within it have loaded. To do that, you can add the `wa-reduce-fouce` class to any element on the page or even apply it to the whole page by placing the class on the `<html>` or `<body>` element.
```

src/styles/utilities/fouce.css
```diff
@@ -18,3 +18,8 @@
     animation: var(--wa-fouce-animation);
   }
 }
+
+/* An opt in utility that hides any container with this class until the autoloader removes it */
+.wa-reduce-fouce {
+  opacity: 0;
```
LeaVerou: This means that if the _autoloader itself_ fails to load, the page will be blank. We can piggyback on the CSS above, and then when the autoloader removes it, it's essentially just a performance optimization.

As a bonus, it gets the same customizable timeout.

LeaVerou: Looks like I can't send an edit suggestion outside the diff. Basically this:

```diff
- :state(wa-defined):has(:not(:defined)) {
+ :state(wa-defined):has(:not(:defined)),
+ .wa-reduce-fouce:has(:not(:defined)) {
```

(actually, now that I think about it, this will stop applying once there are no undefined elements within, so do we even need the JS to remove the class?)

LeaVerou:
Nice! I left some comments/suggestions. It seems to me that the CSS version _may_ actually reduce most of this JS, even for Turbo, do you want to give it a shot?

Not a blocker, but I’m wondering if we should avoid `fouce` in the name of this class, which is a bit esoteric. Some ideas:
- `wa-await-defined`
- `wa-cloak` (inspired from Vue, though also a bit cryptic)
- ?

claviska:
@LeaVerou Feel free to update the PR with whatever changes you see fit. I'm going to move on to my split work.

src/styles/utilities/fouce.css
claviska: updated in aa3cd97ddeb1073625965d32b0086fbd1da8e17b

claviska:
I've renamed the class to `wa-cloak` and removed the CSS class, as requested.

[skipping long paragraph about why the themer picker's view transition tripped me up for over an hour]

New since the last review:

- Removed `.wa-cloak` CSS class
- Disabled SSR on Web Awesome
- Enabled Turbo FOUCE middleware

Let's see how the Vercel deployment looks.

web-test-runner.config.js
```diff
@@ -78,15 +76,8 @@ export default {
             ${componentImports.map(str => `"${str}"`).join(',\n')}
           ]
 
-          window.SSR_ONLY = ${process.env['SSR_ONLY'] === 'true'}
           window.CSR_ONLY = ${process.env['CSR_ONLY'] === 'true'}
         </script>
-        <script type="module">
-          ;(async () => {
-            await import('@lit-labs/ssr-client/lit-element-hydrate-support.js')
-            await Promise.allSettled(window.clientComponents.map(str => import(str)));
```
KonnorRogers, line 87: I believe this line needs to stay and is probably why its failing if i had to guess.

If you never want to run SSR, you would just move this back to the old autoloader.

claviska: It doesn't seem to make a difference. I was just hoping to disable the SSR tests altogether since they've been failing and emailing us for weeks now.

docs/docs/installation.md
```diff
@@ -37,8 +37,33 @@ This snippet includes three parts:
 
 Now you can [start using Web Awesome!](/docs/usage)
 
-:::info
-While convenient, autoloading may lead to a [Flash of Undefined Custom Elements](https://www.abeautifulsite.net/posts/flash-of-undefined-custom-elements/). The linked article describes some ways to alleviate it.
+### Reducing FOUCE
```
LeaVerou: Does it make sense to scatter the docs on FOUCE in two places? I would suggest moving most of this to the existing page on FOUCE utilities and just linking to it here.

LeaVerou:
Hi folks, what's the status on this? While not on the critical path for this split, it's really cool to be able to do navigations this fast without SSR, but if this is not ready for review, I could add `wa-cloak` in a separate PR, since that's a one line change (+ docs).

docs/docs/installation.md
claviska: Feel free to make any changes you'd like to see to these docs.

claviska:
@LeaVerou The tests are passing now so if you want docs changes, it looks like that would be the only thing left. Feel free to commit them directly to this branch if you want.

LeaVerou:
> @LeaVerou The tests are passing now so if you want docs changes, it looks like that would be the only thing left. Feel free to commit them directly to this branch if you want.

Happy to look into this once I finish a couple other things I'm doing. Are you sure you're okay with me committing directly? I don't mind opening another PR with this as target.

claviska:
This PR has been updated to include the `allDefined()` utility and simplify FOUCE handling/utils. I removed the customization and hard coded 2s for the timeout. We can add it back if enough users ask for it, but I suspect they won't and this seems like the perfect default. I also removed the example @LeaVerou added because the behaviors were identical after this update.

Ready to merge when approved.

PR #696, opened 2025-02-04
Palette icons, take 2

### LINKED ISSUE #666: Improve color palette icons
Currently, they all look too similar: 

Image

The reasons are multi-fold:
1. Our hue ranges are too narrow (will be fixed by #545)
2. Very light/dark tints are more similar than the accent colors of each hue. The tintless version of each color should be central to the icon and fewer/no tints should be shown (depends on #220)

Perhaps one solution is to reuse the icon from the color tokens page:

Image

But we’ll need to show neutrals separately — perhaps the center of the circle?

LeaVerou:
Closes #666 😈 

This PR presents two ideas for icons. They are not perfect, but showcase the differences across the different palettes a bit better. @lindsaym-fa feel free to tweak their params and let me know which of the two you like best (plus any other feedback).

## Before (current status)

image

## Take 1 (`palette.njk`)

Similar to the existing icons, but Reduced tints to just 2: accent plus one light/dark set (dark mode sensitive) with a stroke inspired by the most common filled outlined styles.

image
image

## Take 2 (`palette-circular.njk`)

Circular icons, inspired from the color design tokens icon with the addition of gray in the middle.
Down the line we'd want to use actual arcs so we could have even gaps, but these were easier to generate for now. 

image


lindsaym-fa:
I'm leaning towards option 1. (The circular palette makes it a bit tougher to see differences between palettes, imo.) But what if we make the swatches circular or perfectly square instead?

Screenshot 2025-02-12 at 11 25 44 AM

Screenshot 2025-02-12 at 11 29 44 AM

Admittedly, these options give you less surface area for the colors, especially once we add orange... Aesthetically, though, it feels a lot nicer to me. And I think it still offers a good bird's-eye view. What do you think, @LeaVerou?

LeaVerou:
I was actually thinking that the current icons look a lot more distinct now with the palette changes, and perhaps all we need is to have 3 rows instead of 4: light tint, core color, dark tint. 🤔

lindsaym-fa:
> I was actually thinking that the current icons look a lot more distinct now with the palette changes, and perhaps all we need is to have 3 rows instead of 4: light tint, core color, dark tint. 🤔

Works for me! Could we still make the swatches square in that case? 😁 And maybe use the sharper border radius you used in option 1?

LeaVerou:
Oof, I don't like any of my experiments. Because the core colors are so different across palettes, using them together with tints creates a very chaotic look (and not in a good way).
Sharing them here in case they give you any ideas:

Here with more square swatches, 80, core, 20:
image

Here with more surface area for the core color:
image

Just the core colors:
image

80, 50, 20:

image

So far, I’m leaning towards the latter, as they are basically the same as the current ones, just simpler.
I also like the ones with _just_ the core colors, but the form factor is quite unwieldy.

lindsaym-fa:
I really like the second one where the core color has more surface area. I think that's the best representation of the differences between the palettes.

That said, the last option works for me, too. I'm happy to defer to your judgement!

LeaVerou:
> I'm happy to defer to your judgement!

Please don't, I trust your judgement on these things way more than mine 😅 
I’ll send a PR with the larger surface core icons shortly. Meanwhile, I'm closing this one.

Issue #673, opened 2025-02-03
Theme remixing UI: Display previews of palettes and typography options


LeaVerou:
Depends on #666 for palettes

Issue #207, opened 2024-11-07
Remove the `base` part from components that don't need it


claviska:
- It provides little benefit
- It makes the styling story more complicated
- The host element can [most of the time] be the container
- It wasn't a great pattern in the first place and it's always felt akward

In components where the `base` part needs to remain, it should be renamed to something more appropriate.

---

Components that currently have a base part:

- [x] avatar (removed `base` part)
- [x] badge (style is now on host)
- [ ] breadcrumb
- [ ] breadcrumb item
- [x] button (used `display: contents`)
- [x] button-group (no styling of its own really)
- [x] callout (removed `base` part)
- [x] card (removed `base` part)
- [ ] carousel
- [x] checkbox (nothing to do here)
- [ ] color-picker
- [x] details (used `display: contents`)
- [ ] dropdown
- [ ] icon-button
- [ ] image-comparer
- [x] input (nothing to do here)
- [ ] menu-item
- [ ] menu-label (remove?)
- [x] option (removed `base` part)
- [ ] page
- [x] progress-bar (used `display: contents`)
- [ ] progress-ring
- [ ] qr-code
- [x] radio (removed `base` part)
- [x] radio-button (used `display: contents`)
- [ ] radio-group (`button-group__base`)
- [ ] rating
- [ ] skeleton
- [x] slider (nothing to do here)
- [x] spinner (nothing to do here)
- [x] switch (nothing to do here)
- [ ] tab
- [ ] tab-group
- [ ] tab-panel
- [x] tag (removed `base` part)
- [x] textarea (nothing to do here)
- [x] tooltip (nothing to do here)
- [ ] tree
- [ ] tree-item

LeaVerou:
For context, this was prompted by [this Discussion](https://github.com/shoelace-style/webawesome/discussions/192):

> Simply applying CSS properties to the component itself rarely works, because the actual rendered component is in a Shadow DOM `base` part, and the custom element itself is just a shell. This means that often to do simple things, one has to write rules with `::part(base)` all over the place, often even multi-level exported parts!
> 
[...]

Issue #129, opened 2024-06-17
Identify and document custom states for each component


claviska:
With the introduction of custom states,

- [x] Add support for `@cssstate` to the manifest and the docs
- [ ] Add CSS States section to `docs/usage` and/or `docs/customizing`
- [ ] Move `internals` from `WebAwesomeFormAssociatedElement` to `WebAwesomeElement`
  - [ ] Move custom validity functions over too
- [ ] Establish and document a convention for when a custom state should be applied
- [ ] Identify custom states we should apply to each component
- [ ] Implement and document custom states for each component

claviska:
From #168:

> Mainly WebAwesomeFormControl which uses :state(user-valid), :state(user-invalid), :invalid, :disabled, and :valid

LeaVerou:
I believe most of these are now done. Should we close this?

claviska:
Closing as this work has been completed.

PR #699, opened 2025-02-05
[Option] Drop `base` part, rel #207

LeaVerou:
I was working on #673 and at some point it seemed easier to fix `<wa-option>` than to work around it.

This PR:
- Drops the `base` part (related to #207)
- Adds states `current`, `selected`, `hover` (rel #129)
- Removes `hasHover` state, as it’s no longer needed
- Uses both `:hover` and `:state(:hover)` for SSR 
- Adds link to testcase, so we can remove this wart later
- Refactors the watchers to use `updated()` per @claviska’s plan for them

I kept it in case, but it appears like `current` is not being set anywhere, either here or in the [original code](https://github.com/shoelace-style/webawesome/blob/next/src/components/option/option.ts). Possibly leftover? It doesn't seem to be the same as `:state(hover)`.

### DIFF
--- a/src/components/option/option.css
+++ b/src/components/option/option.css
@@ -8,13 +8,7 @@
   color: var(--wa-color-text-normal);
   -webkit-user-select: none;
   user-select: none;
-}
 
-:host(:focus) {
-  outline: none;
-}
-
-.option {
   position: relative;
   display: flex;
   align-items: center;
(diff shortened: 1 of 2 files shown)


claviska (approved):
While reviewing this, I noticed the select is reading as invalid in VoiceOver. It doesn't appear to be related to this PR because it also happens on backers.webawesome.com. However, Shoelace's select still reads correctly

https://github.com/user-attachments/assets/fe7c2966-8d73-40b4-afdd-1666c27f65c5

claviska:
It's impossible to screen capture VoiceOver without doing a video. Here's are screenshots of the screenshot, for clarity.

Incorrect (WA)

CleanShot 2025-02-05 at 09 27 06@2x

Correct (SL)

CleanShot 2025-02-05 at 09 28 02@2x

LeaVerou:
Thanks for the review approval!

Interesting. We can run a `git bisect` at some point to see when this broke, and that should shed some light on what broke it. Can you file an issue?

PR #704, opened 2025-02-05
Add `label` attribute to option & menu-item

### LINKED ISSUE #698: [Option] `label` attribute
When working on #673, I was using rich HTML in `<wa-option>` and noticed the inferred label was …not great:

Image

Fair, it was some pretty complex HTML (cards etc). So then I looked for a `label` attribute on `<wa-option>` (The native `<option>` also has one, though it’s not used much) to specify the label displayed when it is selected. However the only thing I found was [`getTextLabel()`](http://localhost:4000/docs/components/option/#methods).

An imperative method seems useful for certain cases (e.g. lots of options that are not templated) but it seems like a `label` attribute would cover the vast majority of cases that need to override the default, and is much simpler to use — there is a steep learning curve for using any method on components: specifying attributes just requires being comfortable with regular HTML, calling methods needs not just JS, but also timing issues (`whenDefined()` etc), Lit-specific API stuff like `updateComplete` etc.

Once this exists, I’m not even sure `getTextLabel()` is even needed — for the niche use cases that need imperative logic, it seems people can just update the attribute dynamically?

This is not high priority for the themer as I can work around it with `getTextLabel()`, but ideally we should address at some point before the public launch. Happy to do it if there’s consensus for the change.

Edit: Turns out this also affects the menu-item component.

LeaVerou:
I was doing some work with selects for the themer UI and ended up addressing this anyway (mainly because it _seemed_ like a quick fix — boy was I wrong 😅 ).

- Add `label` attribute to option and menu-item, closes #698
- Improve default label generation (ignore contents of `<style>` and `<script>` elements)
- Add `defaultLabel` cached property to both
- Remove `getTextLabel()` method from both (people can just set `el.label`)
- Refactor: Merge slot-aware logic for getting text content into a single utility function
- Fixed typo in `slot.ts`
- Added initial value example in select docs (only had example with `multiple`)
- Update tests and add tests for the new props
- Refactor some watchers into `updated()` per @claviska's plan for them


src/internal/getText.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * Like textContent, but better:
+ * - Uses assignedNodes to get text content from slots (and falls back to content if nothing is slotted)
+ * - Ignores script and style elements
+ * @param root - One or more nodes to get text content from.
+ * @param depth - By default, will just return element.textContent for any child elements instead of calling the function recursively.
+ *                Set to a positive integer to recurse that many levels. Generally a tradeoff between performance and accuracy.
+ * @returns
+ */
+export default function getText(root: Node | Iterable<Node>, depth = 0): string {
```
claviska: Nit: can we call the file `get-text.ts` to align with other files in the project?

claviska (approved):

src/internal/getText.ts
LeaVerou: Ah right, will do!

LeaVerou: Done!

PR #708, opened 2025-02-06
Fix #622

### LINKED ISSUE #622: `.wa-outlined` doesn't fully override default `appearance`
When an appearance utility is set on a component instead of the `appearance` attribute, the component still sets `appearance=""` to whatever its default is. 

Usually, this is innocuous, but in cases of `.wa-outlined`, some of the styles from the default `appearance` leak in.

Such is the case for the following components with`.wa-outlined`:
- `<wa-badge>` – uses `--outlined-border-color` set for `appearance="accent"`
- `<wa-callout>` – uses `--outlined-border-color` set for `appearance="filled"`
- `<wa-tag>` – uses `--outlined-border-color` set for `appearance="filled"`

The effects aren't seen if `.wa-outlined` is combined with `.wa-filled` or `.wa-accent`.

### Screenshots

Image
Image
Image

LeaVerou:


lindsaym-fa (approved):
All seems to be working! Thanks! ✨

PR #716, opened 2025-02-06
Revise `classic` palette, closes #707

### LINKED ISSUE #707: Revise `classic` palette
Revisit the colors in `classic.css` so that the intended core colors (`--wa-color-red`, etc.) have the highest or near-highest chroma.

Key principle for `classic` palette: Mimics the colors used by Shoelace.

lindsaym-fa:
`classic.css` is designed to mimic [the colors offered in Shoelace](https://shoelace.style/tokens/color). This PR brings all of the core colors closer to `50` wherever possible. It also brightens some of the colors (namely `green` and `red`), which importantly brings this palette closer to the spirit of Shoelace's colors.

---

### Before
Screenshot 2025-02-06 at 4 54 59 PM

### After
Screenshot 2025-02-06 at 4 54 55 PM

### Reference
Corresponding colors in Shoelace.
Screenshot 2025-02-06 at 4 33 00 PM

---

### Observations

**Cyan:** `cyan` in this palette skews more green, less blue than Shoelace in order to distinguish it further from the sky blue used for Shoelace's branding, which I feel is important to keep as close to the inspiration as possible. I think I've struck a good balance here — distinguishable without being _too_ green.

**Core colors:** Each of the core colors are a pretty close match to Shoelace's semantic button colors¹, which I feel is a great improvement. That said, to be able to use white button labels and still meet WCAG 2.1 contrast requirements, we'll still want to use `50` for button background colors in the _Classic_ theme.

¹ Buttons in Shoelace
Screenshot 2025-02-06 at 4 48 21 PM


LeaVerou (approved):
This is great work @lindsaym-fa!

My only reservation is the dark accent color for yellow. I wonder if we may want to do this at the theme level. I guess it depends on which one we'd like to show up when people use the accent color for e.g. icons, charts, things like that. Anyhow, not a strong opinion either way.

lindsaym-fa:
Good note. ✔️ We'll give that some thought — implementing the tintless variables in themes in #220 might sway us one way or the other. I'll merge this in the meantime.

PR #731, opened 2025-02-07
Revisit `default` palette, closes #706

### LINKED ISSUE #706: Revisit `default` palette
Revisit the colors in `default.css` so that the intended core colors (`--wa-color-red`, etc.) have the highest or near-highest chroma.

Key principle for `default` palette: Hues are relatively un-opinionated. Colors with the highest chroma tend towards the mid-point (`50`).

lindsaym-fa:
This PR tweaks all hues (`gray` excluded) to bring each core color closer to `50`. `yellow` is a key exception, where the core color is now a truer representation of yellow.

Other key changes:

- Cool down `cyan` and `purple` to be truer representations of their respective hues
- Reduce the chroma of darker `indigo` tints

---

### Before
Screenshot 2025-02-10 at 5 26 05 PM

### After
Screenshot 2025-02-10 at 5 26 01 PM

---

### Observations

**Gray:** I'm wondering if `gray` should have the same chroma clamp as the other colors. I'd love to continue using a _very_ dark gray for `--wa-color-neutral-fill-loud`. Is it important that the `gray` tint with the highest chroma is a midtone, given all tints in the `gray` scale intentionally lack chroma? I'd argue that darker grays (assuming light mode) are naturally "louder" because of this.


LeaVerou:
I actually think the previous green that peaked at 60 looked a lot better. Green is a naturally lighter color (like yellow, but to a lesser extent), it does not need to peak at the same tint as e.g. blue or indigo.

> Gray: I'm wondering if gray should have the same chroma clamp as the other colors. I'd love to continue using a very dark gray for --wa-color-neutral-fill-loud. Is it important that the gray tint with the highest chroma is a midtone, given all tints in the gray scale intentionally lack chroma? I'd argue that darker grays (assuming light mode) are naturally "louder" because of this.

We'll figure it out :) What tint would you like to use for it ideally?

lindsaym-fa:
I agree the lighter green was nicer, but in the spirit of matching to core colors to the intended `loud` colors of the Default theme, `50` is a better fit. 

e.g., the Default theme is meant to have a particular focus on accessibility. For `accent` buttons, we want sufficient contrast between both the surface + the button's background color, and the button's background color + the button's text color. Using `50` as the background color allows us to use white text with a minimum 4.5:1 contrast ratio and also provides a minimum 3:1 contrast ratio between the background color and the surfaces in both light and dark modes.

To that end, I'm actually inclined to bring the core color for `yellow` back up to `70`. I can't really justify using `50` as the core color for `yellow` since it's not as true as the lighter tints, but it may help with visibility against surfaces if it were a bit darker.

Or, perhaps the better path is that we don't use the core colors for `loud` fills in the Default theme and instead prioritize contrast. What would you recommend, @LeaVerou?

LeaVerou:
> Or, perhaps the better path is that we don't use the core colors for `loud` fills in the Default theme and instead prioritize contrast. What would you recommend, @LeaVerou?

What about using the core color for `loud` fills but clamping it to a certain tint (e.g. `50` or `60`)? With the new `-key` variables, we can do that. I could spend some time writing these expressions today.

lindsaym-fa:
As discussed in Slack, we'll explore creating variables for different clamping conditions we may need (e.g., ≤`50`, ≥`60`, etc.) and use `color-mix()` to clamp values to a specific range as-needed in each theme. That will allow us to ensure the contrast requirements of the Default theme in the theme itself rather than influencing the color palette.

With that in mind, I've updated `green` to keep its original core color.

LeaVerou (approved):

PR #732, opened 2025-02-07
Analyze color components

LeaVerou:
Script to print out component ranges for all our colors. 

- Analyze all palettes or a single palette
- Prints out hue ranges, chroma ranges, L ranges, core hue ranges (when analyzing multiple palettes)

@lindsaym-fa if any other stats would be useful for your work let me know!

Sample output for all palettes:
image

Sample output for specific palette:
image

Also, refactored palette scripts:
- Move to `scripts` directory
- Move utils to separate file
- Centralize dir name


lindsaym-fa (approved):
The hue, chroma, and lightness ranges should be helpful! Thanks!

PR #737, opened 2025-02-09
Revisit `mild` palette, closes #720

### LINKED ISSUE #720: Revisit `mild` palette
Revisit the colors in `mild.css` so that the intended core colors (`--wa-color-red`, etc.) have the highest or near-highest chroma.

Key principle for `mild` palette: Reflects the tonal scales of Material, which feature softer colors with low-to-medium chroma.

lindsaym-fa:
Mimics Material's tonal palettes. These tend to, as much as possible, maintain similar levels of chroma across tints.

Key changes:

- Skew `cyan` far more towards blue
- Change previous `indigo` to `purple`
- Add new `indigo`

---

### Before
Screenshot 2025-02-09 at 5 09 31 PM

### After
Screenshot 2025-02-09 at 5 09 28 PM


LeaVerou (approved):
Previous indigo seems definitely more of a purple, but indigo and purple still seem very close (although their hue ranges are squarely within the purple & indigo ranges). 🤔 

One rule of thumb is: would you be ok with a user using these together? (not a rhetorical question, entirely possible the answer is just "yes")

Approving anyway, since this is overall a clear improvement over the current palette.

lindsaym-fa:
> One rule of thumb is: would you be ok with a user using these together?

Yes, though I agree `blue`, `indigo`, and `purple` are pretty close here.

PR #738, opened 2025-02-10
Revisit `vogue` palette, closes #723

### LINKED ISSUE #723: Revisit `vogue` palette
Revisit the colors in `vogue.css` so that the intended core colors (`--wa-color-red`, etc.) have the highest or near-highest chroma.

Key principle for `vogue` palette: Mimics Tailwind's colors. Watch out for overlap with the `classic` palette since Shoelace adapted Tailwind colors.

lindsaym-fa:
Adjusts the colors in the `vogue` palette to further differentiate it from `classic`, both of which are inspired by Tailwind colors. Whereas `classic` uses the hues of Tailwind's red, amber, green, sky, indigo, purple, and gray, `vogue` offers closer matches with neighboring hues rose, yellow, lime, sky, indigo, violet, and slate. (`blue` and `indigo` are therefore very near matches between palettes.)

Key changes:

- `red` matches the hue of TW's rose
- `green` matches the hue of TW's lime
- `blue` matches the hue of TW's sky (needed to distinguish it further from TW's blue-leaning indigo)
- `purple` matches the hue of TW's violet

---

### Before
Screenshot 2025-02-09 at 10 36 37 PM

### After 
Screenshot 2025-02-09 at 10 36 32 PM

---

### Observations

I actually quite like the balance of hues in this palette compared to the original iteration. This lime-leaning `green` is a particularly fun pairing with the _Tailspin_ theme's brand color, `indigo`.

**Cyan:** This definitely skews into teal territory. However, given we don't have a teal, I figure we'd be comfortable splitting the typical hue range of teal between cyan and green.

**Gray:** Here's another example where `gray` chroma peaks darker than `40` (`10`, in this case), and we do actually want that darker tint to serve as the core/loud color for `gray`.


LeaVerou:
> Gray: Here's another example where gray chroma peaks darker than 40 (10, in this case), and we do actually want that darker tint to serve as the core/loud color for gray.

Do you think we should remove the cutoff or special-case it for `gray`?

Also, in those cases, how does that translate to dark mode?

LeaVerou (approved):
Love the green in this palette!

lindsaym-fa:
> Do you think we should remove the cutoff or special-case it for `gray`?

I think `gray` is a special case. Chroma isn't as reliable an indicator of the core color or "loudness" as it is for other colors since `gray` itself isn't colorful. Plus...


> Also, in those cases, how does that translate to dark mode?

Given `gray` will typically be used for surfaces in addition to semantic neutral, I don't think chroma is a sufficient indicator of volume/loudness. i.e., `40` against a white background or an `05` background doesn't accurately reflect peak loudness — on white, `05` would be loudest, and on `05`, white would be loudest. Contrast is more reliable indicator of loudness than chroma when it comes to `gray`, I think.

That said, I'm not sure that `--wa-color-gray` should reflect the loudest tint, nor am I sure that `--wa-color-neutral-loud` should use the loudest possible tint in either light or dark mode (the contrast would be exceptionally stark, and that's not typically preferred). So, what _should_ `--wa-color-gray` represent? The most passive tint that still reads as gray, given grays typically represent something inert/inconspicuous, i.e. always `50`?

LeaVerou:
Agreed. The tintless variable should work well when people use it for e.g. icon colors, so we probably don't want to encode this specific design choice in it.

One question to consider is, given that we decided not to hardcode the `gray` -> `neutral` mapping, to allow for folks to use different colors for all or part of the neutral scale, how would it affect that design choice if people are using something other than gray as their neutral?

PR #739, opened 2025-02-10
`<wa-isolate>` as a wrapper for DSD

LeaVerou:
Needed to use Vue for themer, as DSD is dropped (see https://github.com/vuejs/core/issues/12813 ).

Currently not documented and more of an MVP, but could add docs and polish later if we decide it's more broadly useful.

### DIFF
--- a/docs/_includes/svgs/theme-color.njk
+++ b/docs/_includes/svgs/theme-color.njk
@@ -1,7 +1,7 @@
 {% set themeId = theme.fileSlug %}
 
-<div>
-  <template shadowrootmode="open">
+<wa-isolate>
+  <template>
     <link rel="stylesheet" href="/dist/styles/utilities.css">
     <link rel="stylesheet" href="/dist/styles/themes/{{ page.fileSlug or 'default' }}.css">
     <link rel="stylesheet" href="/dist/styles/themes/{{ themeId }}/color.css">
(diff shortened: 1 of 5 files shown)


claviska:
It's not clear what this does or why it's needed. The use of Vue for the themer is questionable. Why has it gotten so complex that it requires a framework?

LeaVerou:
> It's not clear what this does or why it's needed. 

It's a utility that renders markup in shadow DOM to support encapsulation (e.g. rendering markup with certain stylesheets without including these stylesheets on the whole page). Here we use it for theme icons.

While this helps with Vue, it's not Vue-specific. Many scripts that process the DOM cannot handle DSD because it leaves no trace in the markup, so the only way to handle it properly is to traverse the whole DOM and check each element's `shadowRoot` every time you want to clone a piece of markup. Understandably, most scripts decide not to take that hit.

This component provides an alternative way to isolate a piece of markup, without the pains of `<iframe>`s or the incompatibility of DSD.

LeaVerou:
While I still think something like this can be a lifesaver for certain scenarios (especially given its potential to work around certain bugs in bare shadow DOM, e.g. `@font-face` not working), I'm gonna close this as I think this is a tad too complex in ways that I'm not convinced are worth the complexity. 

Also, I think the name `<wa-isolate>` is confusing, if I were defining it now I'd go for something around scoping, e.g. `<wa-scoped>`.

PR #741, opened 2025-02-11
Clamped brand color tokens

LeaVerou:
Per discussion with @lindsaym-fa, this PR adds:

- `--wa-color-brand-if-lt-N` and `--wa-color-brand-if-gte-N` tokens for use in `color-mix()`. Notes:
  - I only defined these for 40, 50, 60, 80 since it seems these are the most common branching cutoffs and we have no accent colors on other levels, but it's trivial to add variables for all levels.
  - I can define `-lte-` and `-gt-` versions if you think fixing the assummetry is worth the extra weight.
- `--wa-color-brand-N-min` and `--wa-color-brand-N-max` color tokens for the most common cutoffs (40, 50, 60)
  - After working with these a bit, I found `lt(e)`/`gt(e)` easy(ish) to reason about for the conditional tokens, but I found they were making me think a bit too hard when used in color names. I'm still not super happy with `N-(min|max)` but I think they're a little better. I also tried `N-or-(more|less)` but I found it too long and not necessarily more understandable than min/max. The jury is still out on what is a good naming scheme for these.

For convenience, I also added:
- `--wa-color-brand-on` for the common case of using `brand-10` for light accent colors (key >= 70) and `white` otherwise. Happy to add more contrasting versions too (though naming is hard).
- `.wa-brand-*` classes to change the brand color of an element. This was originally added for testing, but I think it can be more broadly useful.

I created a testing page in http://localhost:4000/docs/experimental/clamped-colors/ which we could move to visual tests once these are more mature and we're more confident about exposing them.
Note the text color in the core column below is using `--wa-color-brand-on` 😉 

image

### DIFF
--- a/src/styles/brand/base.css
+++ b/src/styles/brand/base.css
@@ -1,0 +1,63 @@
+:where(:root),
+:host,
+:where([class^='wa-theme-'], [class*=' wa-theme-']),
+:where([class^='wa-palette-'], [class*=' wa-palette-']),
+:where([class^='wa-brand-'], [class*=' wa-brand-']) {
+  /**
+   * Conditional tokens for use in color-mix()
+   * --wa-color-brand-if-lt-N ➡️ 100% if key < N, 0% otherwise
+   * --wa-color-brand-if-gte-N ➡️ 100% if key >= N, 0% otherwise
+   */
+  --wa-color-brand-if-lt-40: calc(clamp(0, 40 - var(--wa-color-brand-key), 1) * 100%);
+  --wa-color-brand-if-lt-50: calc(clamp(0, 50 - var(--wa-color-brand-key), 1) * 100%);
+  --wa-color-brand-if-lt-60: calc(clamp(0, 60 - var(--wa-color-brand-key), 1) * 100%);
+  --wa-color-brand-if-lt-70: calc(clamp(0, 70 - var(--wa-color-brand-key), 1) * 100%);
+  --wa-color-brand-if-lt-80: calc(clamp(0, 80 - var(--wa-color-brand-key), 1) * 100%);
+
+  --wa-color-brand-if-gte-40: calc(100% - var(--wa-color-brand-if-lt-40));
+  --wa-color-brand-if-gte-50: calc(100% - var(--wa-color-brand-if-lt-50));
+  --wa-color-brand-if-gte-60: calc(100% - var(--wa-color-brand-if-lt-60));
+  --wa-color-brand-if-gte-70: calc(100% - var(--wa-color-brand-if-lt-70));
+  --wa-color-brand-if-gte-80: calc(100% - var(--wa-color-brand-if-lt-80));
+
+  /*
+   * Convenience tokens for common tint cutoffs
+   * --wa-color-brand-N-max ➡️ var(--color-brand) if key <= N, var(--color-brand-N) otherwise
+   * --wa-color-brand-N-min ➡️ var(--color-brand) if key >= N, var(--color-brand-N) otherwise
+   */
+  --wa-color-brand-40-max: color-mix(
+    in oklab,
+    var(--wa-color-brand) var(--wa-color-brand-if-lt-40),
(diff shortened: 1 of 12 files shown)


src/styles/brand/base.css
```diff
@@ -0,0 +1,63 @@
...
+    var(--wa-color-brand-60)
+  );
+  --wa-color-brand-60-min: color-mix(
+    in oklab,
+    var(--wa-color-brand) var(--wa-color-brand-if-gte-60),
+    var(--wa-color-brand-60)
+  );
+
+  /* Text color: white if key < 70, brand-10 otherwise */
+  --wa-color-brand-on: color-mix(in oklab, var(--wa-color-brand-10) var(--wa-color-brand-if-gte-70), white);
```
lindsaym-fa: Should this be white if key < 60? As much as I think white on `60` looks better than `10` on `60`, we can only use white on `50` or below for WCAG 2.1 AA compliance. 😢

lindsaym-fa (approved):
Approved with one comment for consideration. Thanks, @LeaVerou!

src/styles/brand/base.css
LeaVerou: Assuming `white` is tint 100, doesn't `60` give us a difference of `40`, i.e. AA?

lindsaym-fa: A difference of `40` ensures a 3:1 contrast ratio, which is only suitable for icons and [large text](https://www.w3.org/TR/WCAG21/#dfn-large-scale), described as an 18pt font, roughly equivalent to 1.5em or 24px.

LeaVerou: Ah bummer, I forgot about that. Maybe we should just remove this token then and set it in our own stylesheet for our own use. With a cutoff that produces 10 on 60, I think we're not going to want to use it, and neither will our users. What do you think?

lindsaym-fa: FWIW, pure black looks just as hairy on `60`, so it's not even that we'd benefit from a special case for `60`; _no_ accessible text color on `60` looks particularly good/readable.

As an accessibility guide rail, I still think `--wa-color-brand-on` is valuable. Because, if you care about WCAG 2.1 compliance _and_ aesthetic usability, the best case scenario is that getting the supposedly accessible combination of `60` + `10` reveals that you _probably_ want to choose a different accent color that looks better.  Worst case scenario, you _really_ want to use `60` and _really_ need to care about compliance, so you go with the technically accessible combination over one that might look better. `--wa-color-brand-on` is a valuable tool in both cases.

I'd hate to remove something useful because the contrast algorithm doesn't accurately represent human perception. 🥲

LeaVerou: Okay, I'll update it to 60 :)

LeaVerou:
```suggestion
  --wa-color-brand-on: color-mix(in oklab, var(--wa-color-brand-10) var(--wa-color-brand-if-gte-60), white);
```

lindsaym-fa: A side-by-side example of `60` with `10` vs black, for context:
Screenshot 2025-02-12 at 1 41 14 PM

Black is better, but not by much.

LeaVerou: Yes, perceptually the best, most accessible color here is white, neither black, nor brand-10. White on bright backgrounds are some of the well known problematic areas of the WCAG algorithm. The fact that it doesn't pass doesn't mean the pairing is not accessible, it's just exposing a bug in the algorithm. I believe APCA passes with flying colors on these pairings.

So I don't think resolving to `white` is _actually_ anti-accessibility, but it is very much anti-conformance to certain a11y legal requirements (that many of our users need to conform to), hence why I originally suggested removing it. It is not steering people the right way, but we also _can't_ steer them the right way without breaking conformance. One of the few cases where we _can't_ have our cake and eat it too. 😕

LeaVerou:
Alright, let's merge this and we can figure out the fate of this variable later — until we document it, it's the tree that fell in the woods and made no sound 😁

Issue #746, opened 2025-02-12
Palette tweaking: Save tweaked palette and have it displayed in sidebar


LeaVerou:
(The palette counterpart of #712 / #714)

- [x] UI to trigger saving
- [x] Save to localStorage
- [x] On all pages with a sidebar: Read localStorage and add items to sidebar
- [x] UI to delete item from localStorage
- [x] Detect that a palette is saved and update its title on its page
- [x] Detect that a palette is saved and show update UI (rename, remove from sidebar)

Wishlist:
- Show tags about what is tweaked (e.g. "More muted", "Greener cyans" etc) rather than just "This palette has been tweaked"
- Better default names based on the actual tweaks

Issue #744, opened 2025-02-11
Make it possible to tweak exact chroma of gray and which hue it’s mixed with


LeaVerou:
This is similar to #669, but requires a completely different solution, as tweaking the hue range of gray is not that meaningful.

The assumption being that grays are tinted with one of the accent colors of the rest of the palette, in varying intensities. So we need to give the user control over both of these:
- Which color is mixed in (`blue` by default)
- How much of it is mixed in (with a reasonable maximum so that the color remains perceptibly gray, and is visually distinct from the color it is mixed with)

For tweaking existing palettes, my current plan is to read the hue of the gray accent, figure out which hue bucket it fits in, and from that assume it is mixed with that color. We'll see how well that works in practice.

LeaVerou:
This was done in #761

Issue #670, opened 2025-02-03
Tweak overall chroma of color palette


LeaVerou:
This is another step towards the themer.

A slider on each palette page, serving double duty: 
1. **Feedback** for how muted the palette is
2. Supporting tweaking for making a whole palette brighter/more muted.

LeaVerou:
Done in #747

PR #747, opened 2025-02-12
Hue tweaking & chroma scaling, closes #669 #670

### LINKED ISSUE #669: Make it possible to tweak exact hue of accent color
This is a first step towards evolving our current theme & palette docs towards becoming the themer.

The accent color of each palette (depends on #668) should display editing affordances. Once it is clicked, there is a popup with a gradient slider letting the user tweak the exact OKLCh hue within the range of acceptable hues for that hue. E.g. clicking on green would slow a slider between more lime-y greens to more teal-y greens.

Some R&D is needed to identify the exact hue range for each of these, but I can use placeholder numbers until @lindsaym-fa's work with the color palettes is finished.

This is for non-grays. The UI and logic for grays will need to be significantly different, so I’ll open a separate issue for that (edit: #744).

- [x] UI for tweaking
- [x] Generate copyable code snippets
- [x] Permalinks
- [x] Highlight tweaked scales
- [x] Affordance to signify that tweaking is possible
- [x] Update contrast ratios
- [x] Make sure it works with Turbo

Later:
- [ ] Ensure hues have at least N difference

LeaVerou:
What this PR does:
- [x] UI for tweaking hue and general infrastructure for palette tweaking
- [x] Generate copyable code snippets
- [x] Permalinks (tweaking updates your URL, which you can use to share or save your tweaks)
- [x] Highlight tweaked scales
- [x] Affordance to signify that tweaking is possible
- [x] Update contrast ratio tables (styling for contrast up/down/fail could use improvement, but it's a starting point)
- [x] Make sure it works with Turbo (i.e. things don't break when we navigate to another page)
- [x] Adds a slider for the palette's overall chroma, which scales the chroma of all tints up/down

What is out of scope for this PR:
- [ ] Ensure hues have at least N difference or steer users away from too close differences (will open new issue for this before merging this PR)
- [ ] Tweaking grays is a separate issue (#744)
- [ ] Storing the tweaked palette locally and displaying it in the sidebar (#746)
- [ ] Using the palette in theme remixing or the docs (that would depend on #746)

## Screenshots

Anodized palette with chroma taken down:
image

Natural palette with chroma taken up:
image

## Notes

The affordance for tweaking and the signifier that a hue has been edited is not great, but it's a starting point. Here green is tweaked and cyan is hovered:
image

I also have not yet figured out a better way to display the old/new colors, so it's still using the swatch top/bottom, which is also suboptimal:
image

The slider now displays hue deltas instead of hues. I'm unsure about this, but thought it may be helpful to have a number to quantify the magnitude of the edit, and be able to transfer it to other hues/palettes.

The code generated looks like this:
image

It's not ideal that colors have to be repeated, but there is no way (that I can think of) to reference a color AND set the very same color on the very same element, so if we wanted to reference the original colors, we would have needed all our palettes to define them twice, which is a bit eek.


docs/assets/scripts/tweak/data.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const hueRanges = {
+  red: { min: 5, max: 35 }, // 30
+  orange: { min: 35, max: 60 }, // 25
+  yellow: { min: 60, max: 120 }, // 60
+  green: { min: 120, max: 170 }, // 50
+  cyan: { min: 170, max: 220 }, // 50
+  blue: { min: 220, max: 260 }, // 40
+  indigo: { min: 260, max: 290 }, // 30
```
lindsaym-fa: I know we should be wary about choking this range too much, but the lower end of `indigo` feels distinctly blue to me. What it we pinch it _just_ a little bit by moving `blue`'s `max` and `indigo`'s `min` to `265`?

lindsaym-fa (commented):
Really cool, @LeaVerou!

Just one bug I'm seeing: The top half of some of the lighter tints seem to be showing as pink. E.g., check out the `95` tint of blue, indigo, and purple in the Anodized palette:
Screenshot 2025-02-12 at 5 31 07 PM

And the `95` and `90` tints of various hues in the Natural palette:
Screenshot 2025-02-12 at 5 33 43 PM

Once the colors are tweaked, the entire affected swatch goes pink:
Screenshot 2025-02-12 at 5 34 28 PM


Every palette seems to have this issue, though it doesn't affect the same tints and hues.

lindsaym-fa:
Oh, one more: Should the colors in the contrast tables update to the new hue?

E.g., when I update the red in the Bright palette like this:
Screenshot 2025-02-12 at 6 00 01 PM

The original red still appears in the contrast tables:
Screenshot 2025-02-12 at 6 00 16 PM

docs/assets/scripts/tweak/data.js
```diff
@@ -0,0 +1,34 @@
...
+
+export const selectors = {
+  palette: id =>
+    [':where(:root)', ':host', ":where([class^='wa-theme-'], [class*=' wa-theme-'])", `.wa-palette-${id}`].join(',\n'),
+};
+
+export const hueRanges = {
+  red: { min: 5, max: 35 }, // 30
+  orange: { min: 35, max: 60 }, // 25
+  yellow: { min: 60, max: 120 }, // 60
```
lindsaym-fa: I think we could stand to reduce `yellow`'s `max`. Here's what the `max` looks like on the Elegant palette (similar results for the other palettes):
Screenshot 2025-02-12 at 6 04 26 PM

LeaVerou:
> Oh, one more: Should the colors in the contrast tables update to the new hue?

Whoops, yes, absolutely they should!!

docs/assets/scripts/tweak/data.js
LeaVerou: Agreed!

docs/assets/scripts/tweak/data.js
LeaVerou: Fine by me!

LeaVerou:
Just pushed fixes for the contrast table colors and the hue range tweaks!

> Just one bug I'm seeing: The top half of some of the lighter tints seem to be showing as pink. E.g., check out the 95 tint of blue, indigo, and purple in the Anodized palette:

I cannot seem to reproduce it. Does it happen when you just visit the page or after you interact with it? Which browser/version?

LeaVerou:
Okay, I've reproduced it. It's our good friend Safari!

LeaVerou:
So, the issue is that Safari incorrectly converts hue to 0 when certain colors are converted to oklch. More details here: https://bugs.webkit.org/show_bug.cgi?id=287637

So this messes with our colors because we implement the hue shift like this:
```css
--color-tweaked: oklch(from var(--color) l c calc(h + var(--hue-shift)));
```

Since it only affects low chromas (< 0.027) and mid to high lightnesses (L > 60) it seems to mainly affect the top half of our grays (but we are not hue shifting them) most of our 95 tints, and in some of the more muted palettes our 90 tints.

So, it's clear we'd need to use different code for tweaking the hue of these tints. For our UI, we can detect the bug and only do it if the user is using Safari with the bug present, but for the code we give to users, we would need to do it for all colors where it might possibly apply.
Here are some options for how to do that.

## Option 1: Just YOLO-ing it with `lch()`

What if we just use `lch()` to hue shift those tints? 
Since it's light colors, low chroma, and relatively small hue shifts, I don't think it will make much of a difference.

- Pros: Easy fix, smallest code difference (some tints just use `lch` instead of `oklch` and aside from that everything is the same across tints)
- Cons: Theoretically it can lead to the difference being visible, what happens with future tweaks (e.g. overall palette chroma)?

## Option 2: Darken in oklab and then lighten 

Since this only happens for L > 60, we _could_ convert to another space first (even `oklab`) and reduce L, then bring it up again (and we can do the latter in the same step as the hue shift, since oklch's L is the same as oklab's L):

```css
--color-oklch-tweaked: oklch(from oklab(from var(--color) calc(l - 0.5) a b) calc(l + 0.5) c calc(h + var(--hue-shift)));
```

- Pros: No risk of slightly different result. Operation still happens in oklch, so it doesn't complicate future tweaks.
- Cons: Longer, more convoluted code, for every affected tint.

## Option 3: Do the hue shift in oklab

Since oklch is just a polar form of oklab and CSS now supports trig functions, we _could_ do the math in oklab, it would just be more convoluted. It would look something like this:

```css
/* Only once per scale */
--color-tweak: l calc(sqrt(a*a + b*b) * cos(atan2(b, a) + var(--hue-shift))) calc(sqrt(a*a + b*b) * sin(atan2(b, a) + var(--hue-shift));

/* Actual tweak */
--color-tweaked: oklab(from var(--color) var(--color-tweak));
```

or, since we can store calculations in variables:
```css
/* Only once: */
--oklch-c: sqrt(a * a + b * b);
--oklch-h: atan2(b, a);

/* Only once per scale */
--color-tweak: l calc(var(--oklch-c) * cos(var(--oklch-h) + var(--hue-shift))) calc(var(--oklch-c) * sin(var(--oklch-h) + var(--hue-shift));

/* Actual tweak */
--color-tweaked: oklab(from var(--color) var(--color-tweak));
```

- Pros: If we use oklab for all tints, the CSS will be _almost_ as nice as today, the wart localized to one variable per scale.
- Cons: This wart has now forced us to change how we do hue shifts across everything, and future palette tweaks (e.g. chroma tweak etc) could complicate it even further 

----

I’m currently leaning towards Option 2 (mainly thinking ahead to future tweaks), so that it doesn't get in the way of future tweaks, but would love some input.

lindsaym-fa:
@LeaVerou, you're a saint for dealing with this. 🫶

If there's any chance the Safari bug could be fixed in a timely manner, I'd vote for option 1 since it seems the most straightforward to bring back to oklch later.

That said, given we can't be sure of the Safari team's responsiveness to this issue, option 2 seems like a good, more future-proof alternative.

LeaVerou:
Even if the Safari bug is fixed tomorrow, it would take at least a year before we can reliably ignore it, and I'm not sure what other tweaks like chroma scaling would look like with Option 1, so I'm inclined to go with Option 2, assuming it works (I have not tested it yet).

LeaVerou:
The saga continues.

Turns out that low L values do not reliably avoid the bug, so Option 2 (in its present form) is out.

Then I wondered: high chromas _do_ seem to reliably not trigger the bug, so perhaps a variation of Option 2 could be multiplying `a` and `b` by a constant factor, and then dividing `c` by the that factor! Hue is not affected by scaling, since it depends on the ratio between a and b. 

Looking at the range of chromas in our palettes (thanks to #752):
image

A multiplier of `6` seems like it would reliably bring the chroma above the "danger zone" of 0 - 0.027.
Putting it together:

```css
--color-tweaked: oklch(
  from oklab(from var(--color) l calc(6 * a) calc(6 * b)) 
  l calc(c / 6) calc(h + var(--hue-shift))
);
```

That actually seems to work, but gosh I’ll need a vacation to recover from this ordeal 🤣 

-----

One issue is that the current code stores the tweak in a single variable. To implement this fix, we'd need to either use two, or restore the chroma in another pass (ending up with **three** nested functions!).

So I think I’ll actually use Option 1 for now, and if that causes issues for chroma scaling, we’ll cross that bridge when we get to it.

docs/docs/palettes/tweak.css
```diff
@@ -0,0 +1,120 @@
...
+    --color-tweaked: oklch(from var(--color) l c calc(h + var(--hue-shift)));
+
+    &.safari-oklch-bug {
+      /* Multiplying and then dividing by 6 to work around  https://bugs.webkit.org/show_bug.cgi?id=287637 */
+      --color-tweaked: oklch(
+        from oklab(from var(--color) l calc(6 * a) calc(6 * b)) l calc(c / 6) calc(h + var(--hue-shift))
+      );
+
+      /* outline: 1px dashed red; */
+    }
```
lindsaym-fa: Should this be removed? Or are we keeping this here for posterity?

docs/docs/palettes/tweak.js
```diff
@@ -0,0 +1,175 @@
...
+
+// Detect https://bugs.webkit.org/show_bug.cgi?id=287637
+const SAFARI_OKLCH_BUG = (() => {
+  let dummy = document.createElement('div');
+  document.body.appendChild(dummy);
+  dummy.style.color = 'oklch(from #d5e0e6 l c h)';
+  let computedColor = getComputedStyle(dummy).color;
+  dummy.remove();
+  return computedColor.endsWith(' 0)');
+})();
```
lindsaym-fa: Do we still want to keep this now that we're using `lch` for the affected tints across the board?

docs/_layouts/palette.njk
```diff
@@ -21,18 +30,57 @@
...
       {{ palettes[paletteId][hue].maxChromaTint }}
-      <wa-copy-button value="--wa-color-{{ hue }}" copy-label="--wa-color-{{ hue }}"></wa-copy-button>
     </div>
+    {% endif %}
   </td>
   {% for tint in tints -%}
-    <td>
-      <div class="color swatch" style="background-color: var(--wa-color-{{ hue }}-{{ tint }})">
+    {%- set color = palettes[paletteId][hue][tint] -%}
+    <td style="--color: var(--wa-color-{{ hue }}-{{ tint }})" :class="{'safari-oklch-bug': SAFARI_OKLCH_BUG && {{ color.c < 0.027 }} }">
```
lindsaym-fa: Just calling out this line as the only usage of `safari-oklch-bug`/`SAFARI_OKLCH_BUG`. Still necessary? Does this ever take effect?

lindsaym-fa (commented):
Looks good! I noted a few places we're still detecting Safari's OKLCH bug since I figured this code might not be necessary now. (I could be wrong, so feel free to ignore!)

Additionally, it looks like there's an extra closing parenthesis in the CSS generated for tweaked colors:
Screenshot 2025-02-13 at 4 42 53 PM
 
Otherwise, all looks good!

LeaVerou:
I fixed the extra closing paren!

Re: Safari bug detection: 

Right now, we apply the `lch()` tweak to the copyable code, but the tweaked Option 2 in the one we use in our UI, only in the places that are actually affected (hence the Safari bug detection).

I can see the argument for displaying the same result our users would get from the tokens we give them though. Would you like me to change it to do that?

lindsaym-fa:
Ah, I see! I don't feel too strongly about this, but if you're on the fence, here's my nudge:

My inclination would be to show the same results that we're giving to users so that nothing comes as a surprise if there are visible differences (i.e., "Why is this happening in my project but not on the docs?"). It'll also make it easier for us to see under what conditions unsightly differences pop up.

lindsaym-fa (commented):
This is really great! Unfortunately, Safari is **_not_**. 💢

We have some rogue pink swatches again — now, they're also in unexpected places! Check out `gray-80` and `gray-70` in the Vogue palette:
Screenshot 2025-02-13 at 6 49 02 PM

And the entire `gray` scale in Classic:
Screenshot 2025-02-13 at 6 49 21 PM

lindsaym-fa (approved):
Approved, pending any further boxing matches with Safari. 🥊

PR #750, opened 2025-02-13
Palette icons, take 4

### LINKED ISSUE #666: Improve color palette icons

LeaVerou:
Per discussion in https://github.com/shoelace-style/webawesome/pull/696

@lindsaym-fa feel free to tweak the params as you please!
Current status:

image


lindsaym-fa (approved):
Looks good to me!

PR #754, opened 2025-02-14
Use tintless and clamped `brand` colors in themes

### LINKED ISSUE #663: Use tintless variables in themes
Per discussion with @lindsaym-fa. This is part of the general arc around regularizing theme CSS so it can be programmatically generated by the themer.

There is some R&D work needed here: once we embark on this we'll realize there are _categories_ of usage, depending on the **contrast** the theme wants to convey, which is a separate "variable". We'll likely need to implement this via separate CSS files for now.

We _may_ end up having to add `-lighter`/`-darker` suffixes to the tintless variables to support this. I’m hoping one level will suffice, but we'll know more once we tackle this.

lindsaym-fa:
Updates our themes to use tintless or clamped `brand` colors where applicable.

The following themes now use whatever the core color is for any hue assigned to `brand`:

- Active
- Glossy
- Playful
- Premium

The following themes use the core color clamped to a specific range:
- Awesome (light: `70-min`, dark: `50-max` — limits contrast with surfaces and ensures button borders stand out)
- Brutalist (light: `50-max`, dark: `70-min` — ensures high contrast with surfaces)
- Default (light and dark: `50-max` — ensures compliant contrast with surfaces)
- Mellow (light and dark: `50-max` — ensures compliant contrast with surfaces)

The following themes are left unchanged because they intentionally target a specific tint to match their sources of inspiration:
- Classic (Shoelace)
- Matter (Material)
- Tailspin (Tailwind)

Additionally, I've added convenience tokens for `--wa-color-brand-70-max` and `--wa-color-brand-70-min`. The latter was immediately useful!

---

Comments on scope:

- `--wa-color-brand-on` is only useful when paired with `--wa-color-brand`. We can't use this token with a clamped value, like `--wa-color-brand-60-max`, because `brand-on` is based on the core color, not the clamped value that's taking effect. So, if the core color is e.g. `70`, `brand-on` will always be `10`, whether the resulting background color is `60`, `50`, or `40`.
- We can only use these tintless/clamped tokens on other `brand` tokens. I wonder if we ought to support `success`, `neutral`, etc. with this system as well. Worth exploring further!

### DIFF
--- a/docs/docs/resources/changelog.md
+++ b/docs/docs/resources/changelog.md
@@ -27,8 +27,9 @@ You can find them in the first column of each color palette.
 
 ### Themes
 
+- Updated Active, Glossy, Playful, and Premium themes so that `--wa-color-brand-fill-loud` uses the core color of the chosen brand color, regardless of tint.
 - You can now override the brand color of any theme with any of the 9 hues supported.
-- Improved UI for theme remixing, with previews and generated copyable code snippets
+- Improved UI for theme remixing, with previews and generated copyable code snippets.
 
 ### Components
 
--- a/src/styles/brand/base.css
+++ b/src/styles/brand/base.css
@@ -30,6 +30,11 @@
     var(--wa-color-brand) var(--wa-color-brand-if-lt-40),
     var(--wa-color-brand-40)
   );
+  --wa-color-brand-40-min: color-mix(
+    in oklab,
+    var(--wa-color-brand) var(--wa-color-brand-if-gte-40),
+    var(--wa-color-brand-40)
+  );
 
   --wa-color-brand-50-max: color-mix(
     in oklab,
@@ -53,6 +58,11 @@
     var(--wa-color-brand-60)
   );
 
+  --wa-color-brand-70-max: color-mix(
+    in oklab,
+    var(--wa-color-brand) var(--wa-color-brand-if-lt-70),
+    var(--wa-color-brand-70)
+  );
   --wa-color-brand-70-min: color-mix(
     in oklab,
     var(--wa-color-brand) var(--wa-color-brand-if-gte-70),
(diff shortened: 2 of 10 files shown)


docs/docs/resources/changelog.md
```diff
@@ -27,8 +27,9 @@ You can find them in the first column of each color palette.
 
 ### Themes
 
+- Updated Active, Glossy, Playful, and Premium themes so that `--wa-color-brand-fill-loud` uses the core color of the chosen brand color, regardless of tint.
```
lindsaym-fa, line 30: How in-depth should we go with this? Do we want to mention the addition of `--wa-color-brand`, `--wa-color-brand-on`, and the clamped tokens?

LeaVerou: I’m in two minds about whether to mention the clamped tokens. Happy to defer to you on that!

LeaVerou (approved):

LeaVerou:
Great work @lindsaym-fa! Approved.

> * `--wa-color-brand-on` is only useful when paired with `--wa-color-brand`. We can't use this token with a clamped value, like `--wa-color-brand-60-max`, because `brand-on` is based on the core color, not the clamped value that's taking effect. So, if the core color is e.g. `70`, `brand-on` will always be `10`, whether the resulting background color is `60`, `50`, or `40`.

Yup, the idea is that with a clamped token, you can use a specific tint over it and it ensures a certain minimum contrast so you don't need something like `--wa-color-brand-on`.

> * We can only use these tintless/clamped tokens on other `brand` tokens. I wonder if we ought to support `success`, `neutral`, etc. with this system as well. Worth exploring further!

Yup, I think that would make a lot of sense. Do you have a sense of whether they would mirror the brand usage or if they would be used differently?

lindsaym-fa:
> Do you have a sense of whether they would mirror the brand usage or if they would be used differently?

In general, I think other semantic tokens would mirror the brand usage so that all e.g. accent buttons have the same visual intensity, regardless of semantics.

PR #755, opened 2025-02-14
Save palette MVP, fixes #746

### LINKED ISSUE #746: Palette tweaking: Save tweaked palette and have it displayed in sidebar

LeaVerou:
Hue tweaking, chroma scaling, and permalinks were implemented in #747.

What this PR adds:
- [x] UI to save a tweaked palette to the sidebar
- [x] Save palette to localStorage
- [x] On all pages with a sidebar: Read localStorage and add saved palettes to sidebar
- [x] UI to delete palette from localStorage on both its own page and its sidebar item
- [x] UI to rename a palette
- [x] Detect that a palette is saved when its page is visited, update its title on the page and show update UI (rename, delete) next to it
- [x] When further tweaks are applied to a saved palette, update it in place 

Screencast of the current behavior:
https://github.com/user-attachments/assets/8ff1c2bb-0954-4de9-895e-2e18c9115f34

Tweaks made after the screencast:

* Nicer description of tweaks with removable tags
image
* Change the generated code to not encode the tweaks in the CSS (this is not necessary, and in some ways not even desirable).

### Save, rename, delete UI

- Save button should not show when the palette is not tweaked or has already been saved
- Save button should show when the palette has been tweaked and not saved
- When navigating to a saved palette, the rename and trash icon buttons should show next to its title
- When renaming a saved palette, the title should be updated in both the page title and the sidebar
- When deleting a saved palette, it should disappear from the sidebar and the title should go back to the default
- When tweaking a saved palette further, the saved palette should be updated with the new tweaks

### Current link in sidebar

- When saving a palette, it should be highlighted as current in the sidebar
- When visiting the URL of a saved palette (regardless of how), the saved palette should be highlighted as current in the sidebar
- When deleting a palette, its parent should be highlighted as current

### Ensure no regressions 

Also good to make sure that things that worked previously still work:
- When tweaking chroma and hue, the URL should update
- When clearing a tweak (by clicking the clear button), the corresponding URL param should be removed
- When visiting a URL with tweaks, the palette should update to reflect the tweaks (including the contrast tables and the code)

### Navigation

- For anything that should happen on page load, make sure it happens when the page loads normally, as well as when navigating to it by clicking a sidebar link (i.e. when Turbo does the navigation).


lindsaym-fa (commented):
Very cool, @LeaVerou! 🤩

I'm seeing odd behavior in any browser when you initially manipulate the chroma or hues. For example, when I increase the overall chroma of the e.g. Elegant palette, it looks like this:
Screenshot 2025-02-18 at 10 17 15 AM

But the contrast tables look like this:
Screenshot 2025-02-18 at 10 17 23 AM

Similarly, if I manipulate hues in the e.g. Mild palette, like this:
Screenshot 2025-02-18 at 10 31 12 AM

The contrast tables look like this:
Screenshot 2025-02-18 at 10 31 20 AM

Refreshing or navigating directly to the URL seems to resolve it, but the problem persists if I navigate to the saved palette via the sidebar.

Additionally, as you can see in the last screenshot, the descriptions don't always populate. Tweaked palettes (whether saved or unsaved) occasionally don't load at all, or the page loads without any of the saved data applied, like in this screenshot:
Screenshot 2025-02-18 at 10 38 49 AM

Here's a console error I'm receiving in Chrome:
Screenshot 2025-02-18 at 10 38 00 AM

LeaVerou:
Okay, I broke it by trying to push some changes today but not others :P I'll fix that asap and re-request a review.

LeaVerou:
Yup, it was a silly thing I broke. Should be ready for another look!

lindsaym-fa (approved):
The removable tags are a neat addition! Approved, though I noticed a minor bug. It's not major enough to block merging this PR, so let me know if you'd prefer to tackle this in a separate issue:

When I reset the edits on a saved palette, then return to the palette, the previous hue modifications persist, but the chroma modification correctly does not.

https://github.com/user-attachments/assets/031268bb-9144-488f-985f-4ca0906a757e

LeaVerou:
Alright, another go. The palette will now stick around even after removing its edits, and no edits stick around after being removed.

lindsaym-fa (commented):
I'm getting into some weird states with saved palettes.

---

1) When I saved a tweaked palette, navigate to a different page, then navigate back to it, a duplicate sidebar item is created. (This only seems to happen the first time navigating back to the saved palette.)
Screenshot 2025-02-18 at 3 29 08 PM

---

2) When navigating back to a saved palette, the heading no longer shows the saved palette's name. 
Screenshot 2025-02-18 at 3 34 24 PM


---

3) On the subsequent views of the saved palette, I can't delete the palette from the sidebar, and the delete button doesn't appear on the page. (It seems to be loading the original palette, given the above issue.)

LeaVerou:
Okay, I _think_ I just fixed these issues. I still sometimes see a duplicate item for a split second, but it disappears.

lindsaym-fa (approved):
No further issues of note! 🎉

PR #762, opened 2025-02-18
Fix bounding rect issue for elements with `display: contents`, fixes #763

### LINKED ISSUE #763: [Radio-button] Cannot be a trigger for `<wa-tooltip>`
As discussed on Slack.

https://github.com/user-attachments/assets/fc5e2eba-9c97-44db-b0df-cf4c51d961a4

This is actually a bug that all elements using `display: contents` will have. The fix is simple: add a `getBoundingClientRect()` like we've done for `<wa-button>`. Perhaps time to do so on the base class, with a per-component-type opt-in (for performance).

LeaVerou:
Classes opt in using `static rectProxy = "propName";` so that no extra work happens on elements that don't need it.

Fixes #763

### DIFF
--- a/src/components/button/button.ts
+++ b/src/components/button/button.ts
@@ -52,6 +52,7 @@ import styles from './button.css';
 @customElement('wa-button')
 export default class WaButton extends WebAwesomeFormAssociatedElement {
   static shadowStyle = [variantStyles, appearanceStyles, sizeStyles, nativeStyles, styles];
+  static rectProxy = 'button';
 
   static get validators() {
     return [...super.validators, MirrorValidator()];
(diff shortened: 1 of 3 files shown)


claviska (approved):
I don't agree with this approach, but I'll approve it to unblock today's release.

This will need to be cleaned up and the change that originally prompted this fix (passthrough CSS styles, let's call them) will need to be reverted to fix this properly.

LeaVerou:
> I don't agree with this approach, but I'll approve it to unblock today's release.

Thank you.

> This will need to be cleaned up and the change that originally prompted this fix (passthrough CSS styles, let's call them) will need to be reverted to fix this properly.

I’m hoping we can make this decision as a team, when the time comes.

PR #766, opened 2025-02-20
[Tiny PR, 8 loc diff] Reduce build process noise

LeaVerou:
- Only run 11ty when something has changed within `docs/`
- Only run esbuild if a JS file has changed within `src/`

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -379,7 +379,9 @@ if (isDeveloping) {
         return;
       }
 
-      await regenerateBundle();
+      if (filename.includes('src/') && /\.(js|ts)$/.test(filename)) {
+        await regenerateBundle();
+      }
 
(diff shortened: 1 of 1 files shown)


scripts/build.js
```diff
@@ -379,7 +379,9 @@ if (isDeveloping) {
         return;
       }
 
-      await regenerateBundle();
+      if (filename.includes('src/') && /\.(js|ts)$/.test(filename)) {
+        await regenerateBundle();
+      }
```
KonnorRogers: I think this will cause stale components on `.css` files

```suggestion
      if (filename.includes('src/') && /\.(js|ts|css)$/.test(filename)) {
        await regenerateBundle();
      }
```

KonnorRogers (dismissed):
Approved, just one tiny fix.

PR #767, opened 2025-02-20
Make CSS great again 😁

LeaVerou:
A while back we discussed selectively using PostCSS to preprocess some of our more repetitive stylesheets. But today, as I was thinking about the clamped tokens stuff, it dawned on me: we don't need to learn any new thing, we can use the templating engine we *already* have and know: Nunjucks!!

So I wrote a CSS preprocessing pipeline based on Nunjucks that is fully integrated with our build process.
 
- Any file ending in `.css.njk` will automatically be preprocessed and spit out a corresponding `.css` file
- Various CSS-related data is available in this context, e.g. hues, tints (see `data.js` for a full list). We can easily add more.
- Each of the arrays in `data.js` can be referenced in **filenames as well** in their singular form. E.g. `{{ hue }}.css` will produce a CSS file per hue.

To demonstrate the possibilities, I've converted the following CSS files to use it:
- `variants.css` is now generated by `variants.css.njk` (and I deleted the existing `variants.css.js` file)
- `brands/base.css` is now generated by `brands/base.css.njk`
- All other stylesheets in `brands/` are generated from `brands/{{ hue }}.css`

Once this is merged, it becomes trivial to add clamped tokens for all hues, which I'll do in a separate PR.

----

Q: Currently the generated files live in the repo. It is typically good practice to gitignore build artifacts. Should the tool add/update `.gitignore` files in these directories to automatically gitignore the build artifacts? Doing it manually is messy, since there is no way to say "Gitignore .css files IFF there is a corresponding .css.njk file in the same directory", so the only way is to do it manually, which is destined to get out of sync.


claviska:
Some of the team has expressed concerns with this PR, so I'll voice mine here as well. I don't think a bespoke CSS preprocessor is the right thing for this project.

I understand the result makes things DRY. I understand there's a subjective belief that this is easier to maintain. I don't think the rest of the team shares that opinion.

Cleverness often translates to complexity which results in bugs and maintenance burdens. Where we used to have boundaries, we continue to color outside the lines and, when the lines become invisible, chaos ensues.

Web Awesome's codebase — even just its CSS — has become far less simple and more difficult to maintain from changes such as this that tend to snowball due to convenience.

If we want to introduce a preprocessor, we need to consider _who_ will be maintaining the source and how comfortable they are with it. I, for one, am not comfortable trying to parse these files and I like to think I'm decent at CSS. I'll let others on the team speak for themselves. My honest opinion is that it feels brittle and hacked together.

A better solution would be, as we previously discussed as a team, to consider PostCSS with a very specific subset of modules, as PostCSS is a platform-leaning superset of CSS that's well-tested and documented. It seems like that would be easier for all of us to write and maintain. However, that would seem to be out-of-scope for this split anyways.

I know this isn't what you want to hear, but I don't want to leave this PR in limbo while others such as #768 build on top of it.

LeaVerou:
I’ve been collaborating with @lindsaym-fa closely, so I’m familiar with her concerns, which were a lot more nuanced:
1. Timing (which I agreed with her on) 
2. A specific file ([`variants.css.njk`](https://github.com/shoelace-style/webawesome/blob/preprocss/src/styles/utilities/variants.css.njk)), which has since been rewritten, but even in its initial form was an improvement over the current state (check out [`variants.css.js`](https://github.com/shoelace-style/webawesome/blob/next/src/styles/utilities/variants.css.js) which this PR deletes). 

> I don't think a bespoke CSS preprocessor is the right thing for this project.

**I agree! This PR *removes* the bespoke JS-based CSS preprocessor** we are _currently_ using for `variants.css` in favor of a templating language that:
* we are _already_ using elsewhere in the project, 
* while not the pinnacle of an active open source project, is far more widely used and maintained than any of the PostCSS plugins available for this 
* provides syntax that is comparable to the alternatives (see below)

In the spirit of evaluating PRs for _improvement_ rather than _perfection_, is [`variants.css.js`](https://github.com/shoelace-style/webawesome/blob/next/src/styles/utilities/variants.css.js) clearer than [`variants.css.njk`](https://github.com/shoelace-style/webawesome/blob/preprocss/src/styles/utilities/variants.css.njk)? Is it really a better option to edit [`variants.css`](https://github.com/shoelace-style/webawesome/blob/next/src/styles/utilities/variants.css) manually?

> If we want to introduce a preprocessor, [...] as we previously discussed as a team, to consider PostCSS

Indeed, we _had_ already resolved to preprocess some of the more repetitive CSS files, right? In that case, the question is not _whether_ to preprocess, but _how_. Back then, _I_ proposed PostCSS with a minimal set of plugins.
I have since researched the options around this, and no longer think it's a good direction. 

### The PostCSS alternatives

Our primary need around CSS preprocessing is **looping** over arrays of predefined strings. That’s it. There is no CSS proposal or standard for this, so the only PostCSS plugins that support it implement ad hoc syntaxes that (IMO) do not look significantly better than this one:

1. [`postcss-each`](https://www.npmjs.com/package/postcss-each): 39k downloads per week, last publish 4 years ago

Example snippet:
```css
@each $icon in foo, bar, baz {
  .icon-$(icon) {
    background: url('icons/$(icon).png');
  }
}
```

4. [`postcss-for`](https://www.npmjs.com/package/postcss-for): 128k downloads per week, last publish 9 years ago

Example snippet:
```css
@for $i from 1 to 3 {
    .b-$i { width: $(i)px; }
}
```

(unclear if this does anything other than numbers)

5. [`postcss-foreach`](https://www.npmjs.com/package/postcss-foreach): 3 downloads per week, last publish 10 years ago

Example snippet:
```css
$array: (foo, bar, baz);
 
@each $val, $i in s, m, l  {
  .icon_size_$(val) {
    background: url('icons/$(array[$i]).png');
  }
}
```

Do you think any of these are sufficiently better or more CSS-y that we want to introduce a whole new tool and syntax? Is there a better PostCSS plugin that I missed?

My thinking was that if none of these provide a **significant** benefit over njk, we may as well use what we already have, so folks don't need to learn a new syntax, and we don't need to add a new dependency.

If you're still unconvinced, it may be helpful to point me to some code in this PR you find particularly hard to process, and I'd be happy to rewrite it with any of these syntaxes. Then we can compare and contrast the code snippets.

### So, what to do?

There are other options., but we should be clear what the primary goal is.
* If it's using a standards-based syntax, we're out of luck, because there isn't one.
* If it's maximizing portability of knowledge, I think Nunjucks is the way to go here, since we're already using it (and it sounds like we will continue to use it in the foreseeable future)
* If it is having syntax that is easy to understand and looks as CSS-y as possible, I’m happy to explore that direction (heck, even as a side project!), but any new syntax would — by definition — be far more bespoke.

Or, we may choose to go back on our previous resolution and not preprocess after all. Given the repetitiveness of these files (which stuff like #768 would increase further), I think that's the worst of all worlds. I’m not in favor of DRY as a goal in itself. Too far on the DRY side and code becomes too clever and hard to process (I've called this the [_peak of drought_](https://lea.verou.me/blog/2020/09/developer-priorities-throughout-their-career/) in the past). But the problems with overly WET code are well established, and we've already hit them (e.g. changing stuff in one place and forgetting to update it in another).

Additionally, exactly because repetition is hard to manage, not having a project-wide process for managing it breeds ad hoc preprocessors like the one in `variants.css.js`.

claviska:
Closing as part of cleanup. If there are any critical pieces of this that need to move forward, let's discuss and separate them into separate PRs.

PR #768, opened 2025-02-20
Semantic scales and bounded color tokens for every hue

LeaVerou:
Depends on #767 

This PR:
- [x] Adds clamped color tokens for every literal hue and aliases brand clamped tokens to these. This eliminates the need for `brand/base.css` though it introduces a `palettes/base.css`.
- [x] Adds "semantic color scales", allowing us to reference colors like `--wa-color-warning-90` or `--wa-color-danger`, including clamped versions. 
- [x] I've made an attempt to use these scales in `default/color.css` but I'm unsure if all grays should be mapped to the corresponding neutral. @lindsaym-fa feel free to tweak to your heart's content!

Also some renaming:
- [x] Renamed `-if-{gte|lt}-{tint}` tokens to just `-{gte|lt}-{tint}`
- [x] Renamd `{tint}-{min|max}` tokens to `{min|max}-{tint}` as both @lindsaym-fa and I felt that was more natural.


claviska:
Closing as part of cleanup, since it's quiet stale. If there's anything we want to bring in, let's make a note and get it assigned in a future split :)

PR #774, opened 2025-02-24
[Tooltip] Specify inherited CSS properties on host, fixes #773

### LINKED ISSUE #773: [Tooltip] Inheritable properties like `text-align` have no effect
Related customer request: https://github.com/shoelace-style/webawesome-alpha/discussions/203#discussioncomment-12304678

Indeed, applying `text-align: center` to the tooltip does nothing: https://codepen.io/leaverou/pen/emYdmLX/6659ac633aa05544d2049b9dee336e18?editors=1100
That’s ...not great. :/ 

At first I thought it was a #207-related issue, but tooltip is marked as "nothing to do here", presumably because it’s wrapping a `<wa-popup>`. However, there is no reason for inherited CSS properties to not inherit through. 

Glancing at the code, it looks like an easy fix.

LeaVerou:
I also removed `--text-color` which wasn't used anywhere.

`--hide-delay` and `--show-delay` also don't appear to be used anywhere, but that’s for another day 😁

### DIFF
--- a/src/components/tooltip/tooltip.css
+++ b/src/components/tooltip/tooltip.css
@@ -4,10 +4,16 @@
   --max-width: 30ch;
   --padding: var(--wa-space-2xs) var(--wa-space-xs);
 
+  /** These styles are added so we don't interfere in the DOM. */
   display: inline-block;
   position: absolute;
 
-  /** These styles are added so we dont interfere in the DOM. */
+  /** Defaults for inherited CSS properties */
+  color: var(--wa-tooltip-content-color);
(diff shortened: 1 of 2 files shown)


claviska:
It looks like `--hide-delay` and `--show-delay` were replaced by properties of the same name so they're safe to remove from the docs.

claviska (approved):

LeaVerou:
> It looks like `--hide-delay` and `--show-delay` were replaced by properties of the same name so they're safe to remove from the docs.

Thanks! I’ll remove them then. Which properties replaced them?

PR #776, opened 2025-02-25
Fix bugs in style utilities

lindsaym-fa:
Fixes two key bugs in our style utilities.

---

### 1) `filled` used `on-quiet` with `fill-normal`

Before this fix, `appearance="filled"` / `.wa-filled` specified `--wa-color-fill-normal` for `--background-color` and `--wa-color-on-quiet` for `--text-color`. This yields unexpected results when the `normal` and `quiet` colors in a theme are significantly difference.

For example, the following styles —
```css
--wa-color-brand-fill-quiet: var(--wa-color-brand-95);
--wa-color-brand-fill-normal: var(--wa-color-brand-50);
--wa-color-brand-on-quiet: var(--wa-color-brand-40);
--wa-color-brand-on-normal: white;
```

— yields the following results:
Filled tags with light blue backgrounds and white text

After this fix, the same styles yield the following results:
Filled tags with light blue backgrounds and dark blue text

---

### 2) `size` based on space, not font size

Before this fix, the `--size-*` variables in our `size` utilities mapped to spacing tokens, e.g. `--size-m: var(--wa-space-m)`. This yields unexpected results in all themes, often resulting in a components that are smaller than intended since the spacing scale is not intended for sizes. Instead, we want these to map to font size tokens, e.g. e.g. `--size-m: var(--wa-font-size-m)`.

For example, our `size` utilities yielded larger than intended components for the Awesome theme, which has an increased spacing scale —
Before:
Awesome buttons larger than the default font size

After:
Awesome buttons matching the default font size

And, our `size` utilities yielded smaller than intended `small` components for the Default theme (and many others), given the default spacing scale —
Before:
Pair of small and medium dropdowns, small dropdown smaller than expected

After:
Pair of small and medium dropdowns, small dropdown not too small

### DIFF
--- a/src/styles/utilities/appearance.css
+++ b/src/styles/utilities/appearance.css
@@ -45,7 +45,8 @@
 
   --outlined-border-color: var(--wa-color-border-normal);
 
-  --text-color: var(--wa-color-on-normal, var(--wa-color-neutral-on-normal));
+  --text-color: var(--wa-color-on-quiet, var(--wa-color-neutral-on-quiet));
+  --text-color-hover: var(--wa-color-on-normal, var(--wa-color-neutral-on-normal));
 }
 
 .wa-plain,
--- a/src/styles/utilities/size.css
+++ b/src/styles/utilities/size.css
@@ -9,10 +9,10 @@
 
 :host {
   /* Components are meant to override these */
-  --size-xs: var(--wa-space-xs);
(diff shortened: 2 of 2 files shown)


LeaVerou (approved):
Oh wow, nice catch! No idea why I used spacing tokens for size 🤦🏽‍♀️

PR #796, opened 2025-03-12
[Select] More robust dynamic value / options handling, fixes #789

### LINKED ISSUE #789: [Select] Values set via JS are sometimes quietly dropped / overwritten
While most of our components work fine with a their `value` being set in JS before they initialize, `<wa-select>` seems to fail and clears it out (edit: sometimes).
Among other things, this means it doesn't work well with Vue. 

Testcase: https://codepen.io/leaverou/pen/GgRMayP/b522aeefab7fd00b780df35935110c30?editors=1010 
There is some commented out Vue code in the testcase as at first I thought it was a Vue + WA interop bug, but eventually I realized the vanilla JS version doesn't work either.

I've investigated a little bit, and it appears that while the `value` property gets preserved just fine, it does not get converted to `selectedOptions`, and thus [this line](https://github.com/shoelace-style/webawesome/blob/next/src/components/select/select.ts#L634) overwrites it.

I might send a PR if I can figure it out soon, as it seems it might actually be easier to fix this than work around it in my code.

Edit: Actually, it seems to be a race condition — in some cases the value _is_ correctly ported to `selectedOptions` and in others it is not. While I can reliably reproduce it with Vue, I cannot reliably reproduce it with vanilla JS.

(related: #650)

LeaVerou:
Fixes #789.

This PR implements a private property (`_value`) that keeps the full value of `this.value` around, whether its values exist in the DOM or not. Instead of subsetting it in place, which is lossy, it subsets it in a custom getter, so that these values are still preserved.

To avoid DOM I/O every time `el.value` is read, it also caches the values of DOM options in a set, stored in the private `optionValues` property. This is lazily evaluated when it is actually needed, using `undefined` as a dirty flag.

I discussed the general plan with @KonnorRogers earlier today and he agreed it was the way to go. Not my most elegant piece of code, but it does fix the issue, passes all tests (including new ones I added) and I cannot spot any regressions. 🤞🏼

### DIFF
--- a/src/components/select/select.test.ts
+++ b/src/components/select/select.test.ts
@@ -647,6 +647,7 @@ describe('<wa-select>', () => {
             );
             const el = form.querySelector<WaSelect>('wa-select')!;
 
+            expect(el.defaultValue).to.equal('option-1');
             expect(el.value).to.equal('');
             expect(new FormData(form).get('select')).equal('');
 
@@ -657,6 +658,7 @@ describe('<wa-select>', () => {
 
             await aTimeout(10);
             await el.updateComplete;
+            expect(el.optionValues ? [...el.optionValues] : []).to.have.members(['option-1']);
             expect(el.value).to.equal('option-1');
             expect(new FormData(form).get('select')).equal('option-1');
           });
@@ -745,6 +747,8 @@ describe('<wa-select>', () => {
             );
 
             const el = form.querySelector<WaSelect>('wa-select')!;
+            expect(el.optionValues ? [...el.optionValues] : []).to.have.members(['bar', 'baz']);
+            expect(el.optionValues?.size).to.equal(2);
             expect(el.value).to.have.members(['bar', 'baz']);
             expect(el.value!.length).to.equal(2);
             expect(new FormData(form).getAll('select')).to.have.members(['bar', 'baz']);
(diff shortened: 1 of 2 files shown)


claviska (approved):
My only nit is that we don’t tend to use the `_value` convention for properties but not a blocker.

LeaVerou:
> My only nit is that we don’t tend to use the `_value` convention for properties but not a blocker.

I followed it because `_defaultValue` in the same file followed the same convention. Is there a different convention you'd prefer?

KonnorRogers:
> My only nit is that we don’t tend to use the _value convention for properties but not a blocker.

Its a `private` property. We could use actual private properties like `this.#value`, but to my knowledge we dont actually use that in the codebase anywhere.

claviska:
git blame

\*stares at @KonnorRogers*

😅

KonnorRogers:
> git blame
> 
> *stares at @KonnorRogers*
> 
> 😅

Blame Lit. I stole it from them. https://lit.dev/docs/components/properties/#accessors-custom

```js
private _prop = 0;


@property()

set prop(val: number) {

  this._prop = Math.floor(val);

}


get prop() { return this._prop; }
```

LeaVerou:
> > My only nit is that we don’t tend to use the _value convention for properties but not a blocker.
> 
> Its a `private` property. We could use actual private properties like `this.#value`, but to my knowledge we dont actually use that in the codebase anywhere.

I have no strong opinion either way, but some issues I've found from using these extensively in other projects are:
- Instances can no longer be proxied (without hacks), see https://lea.verou.me/blog/2023/04/private-fields-considered-harmful/ Though this is less of an issue with elements, as they are rarely proxied anyway.
- Because they are static, the compiler often doesn't understand that you're dealing with instances of the same class 
- Many use cases actually need `protected` rather than `private` access, for which a naming convention is still the only way.
- A naming convention is easier to debug.

None of these are super strong reasons not to use private fields, though the benefit is also pretty marginal. 🤷🏽‍♀️

PR #797, opened 2025-03-13
Add `--wa-color-{role}-N` variables, closes #785

### LINKED ISSUE #785: Use `--wa-color-[role]` variables for all roles, not just brand
This is necessary to support setting default roles in palettes (see #782), as currently themes refer to literal tokens for anything other than brand.

Some or all of the CSS changes can probably be cherry picked out of #768, leaving only the automation in there.

lindsaym-fa:
Selectively adds the following changes from #768:
- Adds clamped color tokens for every literal hue and reframes the corresponding brand clamped tokens as aliases for these. This eliminates `brand/base.css` and introduces `color/base.css`.
- Adds role-based color aliases for neutral, success, warning, and danger, allowing us to reference colors like `--wa-color-warning-90` or `--wa-color-danger`.
- Renames `-if-{gte|lt}-{tint}` tokens to just `-{gte|lt}-{tint}`.
- Renames `{tint}-{min|max}` tokens to `{min|max}-{tint}`.

Additionally:
- Leverages the role-based tokens in every theme.
- Removes unused `--wa-color-brand-key` tokens.
- Removes `--wa-color-brand-{lt|gte}-40`, `--wa-color-brand-{lt|gte}-80`, and `--wa-color-brand-{max|min}-40`, which have limited value. Details:
  - Because core colors are clamped to `40` or higher on generation:
    - `--wa-color-brand-max-40` will never result in any tint less than `40`.
    - `--wa-color-brand-min-40` has the same results as `--wa-color-brand`.
    - `--wa-color-brand-lt-40` and `--wa-color-brand-gte-40` have limited potential use outside of `--wa-color-brand-{max|min}-40`.
  - `--wa-color-brand-{lt|gte}-80` is currently unused and has limited potential use compared to e.g. `--wa-color-brand-{lt|gte}-70`.


docs/docs/experimental/clamped-colors.njk
```diff
@@ -1,47 +0,0 @@
----
-title: Clamped brand tokens
```
LeaVerou, line 2: #768 removed this file because it also added docs for these tokens, so we could use the docs to test that they were working properly. I'm fine either way, but I think we should _either_ have this, or have _docs_. Otherwise we have no way to test these.

LeaVerou:
Glad #768 was useful!

> Removes unused `--wa-color-brand-key` tokens.

My weak preference would be to maintain the expectation that everything that exists in `--wa-color-[hue]-*` will also exist in the corresponding `--wa-color-[role]-*` though I understand that with 5 roles and 10 hues, that would add 50 declarations, so I'm also ok with deferring it until it is needed.

Your reasoning for the removed tokens sounds very sound to me 👍🏼 (yes, the terrible pun was intended 😅 )

My only actual pushback would be to either re-add the removed file or the docs (if we want to expose these) so we have a way to visually test. Aside from that, everything LGTM! Thanks for your hard work on this!

If you get a moment after the PR is merged, it would be good to update #768 so that the changes already incorporated from it are not also present there (perhaps just updating it from the main branch will do that anyway) but that can wait until cooldown.

LeaVerou (approved):

lindsaym-fa:
Thanks for that clarification, @LeaVerou! I've reverted the removal of the clamped tokens test page and refactored it to use the clamped hue tokens instead of the clamped `brand` tokens.

LeaVerou (approved):
LGTM!

PR #798, opened 2025-03-13
Add curated `orange` to all palettes, closes #657

### LINKED ISSUE #657: Add `orange`

lindsaym-fa:
These changes more carefully curate `orange` values in each color palette. In the Classic palette, `red` and `yellow` are also adjusted.


LeaVerou (approved):

LeaVerou:
Wonderful!

PR #801, opened 2025-03-13
[Icon-button] Make `--background-color-hover` work + remaining 3 interaction properties

### LINKED ISSUE #800: [Icon-button] `--background-color-hover` no longer works
Testcase: https://codepen.io/leaverou/pen/WbNXayV?editors=1100

LeaVerou:
- Fixes #800 
- Introduces `--text-color-hover`, `--background-color-active`, `--text-color-active`

### DIFF
--- a/src/components/icon-button/icon-button.css
+++ b/src/components/icon-button/icon-button.css
@@ -22,8 +22,8 @@
 
 :host(:not([disabled])) .icon-button:hover,
 :host(:not([disabled])) .icon-button:focus-visible {
-  background-color: var(--wa-color-neutral-fill-quiet);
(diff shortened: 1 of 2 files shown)


claviska (approved):
Are we still planning on merging this with button?

LeaVerou:
Thanks for the approval!

> Are we still planning on merging this with button?

I’d love to, but IMO other changes should happen first for the ergonomics to be comparable. To achieve the same result as:

```html
<wa-icon-button name=“foo” label=“Foo”></wa-icon-button>
```

with button today it would be:

```html
<wa-button appearance=“plain” variant=“neutral”><wa-icon name=“foo” label=“Foo”></wa-icon></wa-button>
```
 which feels a lot clunkier.

PR #803, opened 2025-03-17
11ty for webawesome-app

KonnorRogers:
This PR is what makes it possible to run pre-compile pages with 11ty and run an app on top to inject additional information.


claviska (approved):
Prettier :D

PR #819, opened 2025-03-20
fix remove event and return null when empty

claviska:
Fixes the [backer issue described here](https://github.com/shoelace-style/webawesome-alpha/discussions/215) by removing the `<div>` (which was used to listen for tag removal) and using event delegation on the tag wrapper. Also added an explicit check to ensure the tag isn't falsey before returning.

### DIFF
--- a/src/components/select/select.ts
+++ b/src/components/select/select.ts
@@ -596,13 +596,7 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
 
     // If no direct option was provided, find the option from the event path
     if (!option) {
-      const path = event.composedPath();
-      const tagElement = path.find(
-        el =>
-          el instanceof HTMLElement &&
-          (el.tagName.toLowerCase() === 'wa-tag' ||
-            (el instanceof HTMLElement && el.hasAttribute('part') && el.getAttribute('part')?.includes('tag'))),
-      );
+      const tagElement = (event.target as Element).closest('wa-tag[part~=tag]');
 
       if (tagElement) {
         // Find the index of this tag among all tags
--- a/src/components/select/select.ts
+++ b/src/components/select/select.ts
@@ -291,7 +291,6 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
           ?pill=${this.pill}
           size=${this.size}
           removable
-          @wa-remove=${(event: WaRemoveEvent) => this.handleTagRemove(event, option)}
         >
           ${option.label}
         </wa-tag>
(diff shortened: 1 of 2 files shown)


src/components/select/select.ts
```diff
@@ -587,10 +586,39 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
...
+
+    // If no direct option was provided, find the option from the event path
+    if (!option) {
+      const path = event.composedPath();
+      const tagElement = path.find(
+        el =>
+          el instanceof HTMLElement &&
+          (el.tagName.toLowerCase() === 'wa-tag' ||
+            (el instanceof HTMLElement && el.hasAttribute('part') && el.getAttribute('part')?.includes('tag'))),
+      );
```
LeaVerou: Could `event.target.closest("wa-tag[part~=tag]")` potenitally do the same thing in a simpler way?

src/components/select/select.ts
```diff
@@ -587,10 +586,39 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
...
+      if (tagElement) {
+        // Find the index of this tag among all tags
+        const tagsContainer = this.shadowRoot?.querySelector('[part="tags"]');
+        if (tagsContainer) {
+          const allTags = Array.from(tagsContainer.children);
+          const index = allTags.indexOf(tagElement as HTMLElement);
+
+          if (index >= 0 && index < this.selectedOptions.length) {
+            option = this.selectedOptions[index];
+          }
```
LeaVerou, line 610: Getting the option by index seems fragile, but I may not have all the details.

I had assumed tags are entered in the order they are clicked, which would not necessarily match the order of the options.

If what we're trying to do is get from the tag to the option, some ways might be:
- a `WeakMap` with the tag elements as keys and the options as values. Downside: needs to be updated if the tag reference changes.
- setting data- attributes on the tags with the option values they are for (or `value-${value}` parts, so they can be used for styling!). Downside: what if there are multiple options with the same value? Edge case, but still. Also, needs to be updated if option value mutates.
- A map of tag indices to selectedOptions indices
- Or just making sure `selectedOptions` are in the same order (which may already be the case 😅 )

LeaVerou (approved):
Nice!

Non-blocking: The code to find the option seems a little too complex for what it does. I left some comments with ideas, in case any of them are useful.

src/components/select/select.ts
claviska: Good catch. In this case it doesn't matter so going with `.closest()`.

src/components/select/select.ts
claviska: They're in order by design. Whether that design is correct or not is arguable of course 🙃

src/components/select/select.ts
```diff
@@ -596,13 +596,7 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
...
     // If no direct option was provided, find the option from the event path
     if (!option) {
-      const path = event.composedPath();
-      const tagElement = path.find(
-        el =>
-          el instanceof HTMLElement &&
-          (el.tagName.toLowerCase() === 'wa-tag' ||
-            (el instanceof HTMLElement && el.hasAttribute('part') && el.getAttribute('part')?.includes('tag'))),
-      );
+      const tagElement = (event.target as Element).closest('wa-tag[part~=tag]');
```
LeaVerou, line 599: I can't remember, is there any chance `event.target` can be a text node here? If so, we may want to do:

```suggestion
let targetElement = event.target.closest ? event.target : event.target.parentElement;
const tagElement = targetElement.closest('wa-tag[part~=tag]');
```

PR #821, opened 2025-03-21
Content hierarchy bugfixes & improvements

LeaVerou:
This PR fixes several issues around sidebar listings and overview pages, and makes it easier to maintain the website's content hierarchy with less manual work.

- Sidebar, overview listings, breadcrumbs now based on actual parent-child relationships, rather than increasingly outdated heuristics 
   - Previously: if a child had a parent, it was excluded from the top level of the sidebar, which is why patterns weren't displaying when proper `parent` properties were used
  - Previously: if a child had a parent, it was excluded from overview listings, unless `listChildren` was used (this option is now removed, as it is no longer needed)
  - Previously: The breadcrumbs were generated from the URL structure, regardless of any `parent` properties, with some hacks for the existing parent-child relationships in components
- `parent` properties are now generated automatically from the URL structure, and need only be specified to override that default
- Ability to group by page hierarchy in overview pages, where pages that have >= 2 children (can easily become configurable) become categories

Other smaller improvements:
- More flexible syntax for specifying the params of overview pages (e.g. title overrides are now specified separately, which means you don't need to list all tags just to override one title) 
- In overview pages, hide group heading if only one group is present (e.g. now we no longer have a weird "Tokens" group in the [Tokens](https://backers.webawesome.com/docs/tokens) page)
- `parentItem` and `parentUrl` properties that can be used on any page
- Alias a collection as the children of a page (useful for "virtual" parents like Layout)
- Do not error if a page card icon is missing

With these changes, the overview pages for patterns should work with almost no config. I will send an additional small PR to `patterns-main` once this one is merged to update the listing pages.


docs/docs/components/index.njk
```diff
@@ -2,13 +2,10 @@
...
-  - inputs
-  - navigation
-  - organization
-  - helpers: 'Utilities'
 override:tags: []
+categories:
+  tags: [actions, feedback, imagery, inputs, navigation, organization, helpers]
+  titles:
+    - feedback: 'Feedback & Status'
+    - helpers: 'Utilities'
```
lindsaym-fa: It doesn't look like this is doing what it should. (I kind of think that's fine, though. I don't mind "Feedback" and "Helpers" as the section headings instead.)
Screenshot 2025-03-21 at 2 22 13 PM

lindsaym-fa (approved):
Working well on my end! Left a comment on one specific piece that doesn't seem to be working, and I also noticed that the order changed slightly under Components > Actions —

Before (Button comes before Button Group, which is technically correct):
Screenshot 2025-03-21 at 2 41 20 PM

After (Button Group comes before Button):
Screenshot 2025-03-21 at 2 41 16 PM

Hardly worth mentioning and certainly not a blocker.

docs/docs/components/index.njk
LeaVerou: Turns out that was a very easy fix: The hyphens in front were turning this into an array. Removing the hyphens made it work fine. I've always been confused by this in YAML.

LeaVerou:
> I also noticed that the order changed slightly under Components > Actions —
> 
> Before (Button comes before Button Group, which is technically correct): Screenshot 2025-03-21 at 2 41 20 PM
> 
> After (Button Group comes before Button): Screenshot 2025-03-21 at 2 41 16 PM
> 
> Hardly worth mentioning and certainly not a blocker.

Interesting, I see the correct order on my end:

image

Either I inadvertently fixed it, or it's a race condition 👀

PR #823, opened 2025-03-21
Scrub `:host-context()` from everywhere

LeaVerou:
As discussed on Slack yesterday, since [the CSS WG resolved to remove `:host-context()` this week](https://github.com/w3c/csswg-drafts/issues/1914#issuecomment-2737310093) and since [it’s not supported outside Chromium based browsers](https://caniuse.com/mdn-css_selectors_host-context), let’s kill it with 🔥


claviska (approved):
🧽🫧

PR #777, opened 2025-02-26
Generate custom color palette prototype

LeaVerou:
This PR contains work on generating custom color palettes from one or more core colors. We decided to focus on the themer for now, but may return to it later.
Details in https://github.com/orgs/shoelace-style/projects/9/views/1

PR #828, opened 2025-03-24
Pull in some changes from #777 

LeaVerou:
Since we're not merging #777 for a while, this PR pulls in some of the various bugfixes, improvements, and refactors from it that are _not_ tied to custom palettes, a lot of which is useful for the themer work.

I've scrubbed all the code related to custom palettes and roles from it, but it's entirely possible I've forgotten something.

In terms of user-facing changes, this PR:

Adds popups to copy the color value for any tint (UI adjusted to show a copy icon instead of a tweaking icon)
image

Makes saving explicit and moves Save button to the title. Saving no longer requires a name.
image

Saved variations section at the bottom of palette pages (no icon)
image

Hue wheel showing palette key colors (made more discreet when not interacted with)
image 
image

Updated color popup that shows a tick for the default value and supports copying the CSS color too:
image

Overall colorfulness slider now shows the brightest color in the palette instead of being hardcoded to blue:
image

If there are any objections for the user-facing changes, I can move any of them back to #777, the main goal is to port the underlying changes.


claviska:
@LeaVerou is this stale or should we keep open?

LeaVerou:
> @LeaVerou is this stale or should we keep open?

This one is definitely stale. We might want to keep #777 open though, as it includes all the past work on generating custom color palettes.

claviska:
> We might want to keep https://github.com/shoelace-style/webawesome/pull/777 open though, as it includes all the past work on generating custom color palettes.

Would you mind updating the description (the linked card doesn't show me anything)? I'll re-open and mark it draft.

PR #830, opened 2025-03-25
Add `appearance` to details, closes #569

### LINKED ISSUE #569: [Details] Add `appearance`
Seeing this on the Google Fi website reminded me that `appearance` (and possibly `variant`?) would be useful for Details too (especially for `filled`)

Image

And looking at `<wa-tree>` provides an example of how `appearance=plain` can be useful too:

Image

LeaVerou:
I needed a plain disclosure widget for some of my themer work and it was actually quicker to do it properly. 😁 

- Includes both `<details>` and `<wa-details>`
- I excluded the `accent` value as that's a) far less useful here and b) trickier due to the icon color

image

image

Also fixes a broken link on two other pages.

### DIFF
--- a/docs/docs/components/details.md
+++ b/docs/docs/components/details.md
@@ -77,6 +77,31 @@ The details component automatically adapts to right-to-left languages:
 </wa-details>
 ```
 
+### Appearance
+
+Use the `appearance` attribute to change the element’s visual appearance.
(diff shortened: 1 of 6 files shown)


lindsaym-fa (approved):
Looks great!

PR #835, opened 2025-03-26
Remove unused SSR module and remove first load fade

claviska:
- Removes the unused SSR module from the 11ty config
- Improves first load and page refreshes by disabling the view transition (handled by wa-cloak now)


LeaVerou (approved):
LGTM, but why are we commenting it out instead of deleting these lines? Do we plan to re-add later?

claviska:
> LGTM, but why are we commenting it out instead of deleting these lines? Do we plan to re-add later?

We still consider SSR to be experimental, so this let's us quickly re-enable it for testing.

PR #836, opened 2025-03-26
Theme icons, take 1

LeaVerou:
This is a prototype of theme icons based on @lindsaym-fa’s sketch: 

image

> Current thinking is something like this —
> 
> - Overall background uses the default surface color. Maybe we also outline this with the surface border color.
> - Text sample uses the heading font
> - Swatches show the loud fills for brand, success, warning, and danger
> - Brand swatch is larger and has icon that uses the corresponding “on” color
> - Alternatively, we could focus in on the brand color and show loud, normal, and quiet fills as swatches
> - Second row includes a small input with placeholder text and a small brand button
> 
> That seems like it would be the minimum to get an impression of fonts, colors, rounding, borders, tactility, etc.
> 
> Open to suggestions or other explorations!

Screenshot:
image

With `--wa-color-surface-default`:
image

The icons are using the actual CSS files (rendered in shadow DOM) so they can easily work with custom themes.
I did not include the icon on the brand swatch since it will just be noise until we can have different icons per theme (but I’m not opposed to adding them now if you feel that's best).

I feel they look a bit too noisy all together, but it gives us a starting point we can improve on. Of course it goes without saying that if you have the cycles @lindsaym-fa feel free to go in this PR and tweak them to your heart's content.

### DIFF
--- a/docs/_includes/head.njk
+++ b/docs/_includes/head.njk
@@ -24,6 +24,9 @@
 <link rel="stylesheet" href="/assets/styles/hydration-errors.css">
 <link rel="preconnect" href="https://cdn.jsdelivr.net">
 
+{# Internal components #}
+<script type="module" src="/assets/components/scoped.js"></script>
+
 {# Web Awesome #}
 <script type="module" src="/dist/webawesome.loader.js"></script>
 
@@ -46,7 +49,6 @@
 <link rel="stylesheet" href="/dist/styles/webawesome.css" />
 <link id="color-stylesheet" rel="stylesheet" href="/dist/styles/utilities.css" />
 <link rel="stylesheet" href="/dist/styles/forms.css" />
-
 
 {# Used by Web Awesome App to inject other assets into the head. #}
 {% server "head" %}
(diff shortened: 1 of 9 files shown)


lindsaym-fa (approved):
I think this works well overall! I did tweak the color swatches a bit to reduce their height and the whitespace between. To me, this helps the swatches draw less attention, which might have been contributing to the optical noise.

Screenshot 2025-03-26 at 11 33 52 AM
 
I did notice that the sample text doesn't always render with the theme's heading font family, like in the screenshot above. Note the Awesome theme icon isn't rendering Quicksand, Brutalist isn't rendering IBM Plex Sans, Playful isn't rendering Fredoka, etc. Approved in the meantime if we want to circle back to fix that later.

LeaVerou:
> I did tweak the color swatches a bit to reduce their height and the whitespace between. To me, this helps the swatches draw less attention, which might have been contributing to the optical noise.

I love it! Much, much better.

> I did notice that the sample text doesn't always render with the theme's heading font family, like in the screenshot above. Note the Awesome theme icon isn't rendering Quicksand, Brutalist isn't rendering IBM Plex Sans, Playful isn't rendering Fredoka, etc. 

It did feel that something was off about the typography, glad you were able to identify it! It seems to me that in all of these cases, the issue is that a web font is not loading, can you confirm?

PR #837, opened 2025-03-26
Improve search lists

claviska:
- Adds a debounce to the search list so it feels more natural
- Improves search grid styles so we don't end up with things like this

Before
CleanShot 2025-03-26 at 11 59 52@2x

After
CleanShot 2025-03-26 at 12 00 21@2x


LeaVerou:
Nice! This thing with the overly long cards bugged me for a while.

PR #839, opened 2025-03-26
Various bug fixes

claviska:
Fixes numerous bugs. See commits for links to each issue.


claviska:
Note: I took a few stabs at [fixing the padding](https://github.com/shoelace-style/webawesome-alpha/issues/177) in `<details>` and `<wa-details>` but wasn't successful.

It looks like this got broken when native styles were introduced and, while I was able to get it visually correct in the following test cases, the animation began jumping and it became a rabbit hole I had to get out of. All changes to details in this PR have been reverted.

I would suggest reverting the styles to what we have in Shoelace and styling native ones differently. In fact, I don't believe the perceived benefit of sharing as many styles between native and non-native components is worth the bugs and disadvantages it causes.

#### Test cases

- `<details>` with just a text node for content
- `<details>` with a `<p>` for content
- `<wa-details>` with just a text node for content
- `<wa-details>` with a `<p>` for content

PR #841, opened 2025-03-27
Fix for theme icons + easier to generate palette icons

LeaVerou:
Work extracted from my themer branch since this can be merged as a separate PR (and also to get @lindsaym-fa's eyes on some of this).

**Bugfix:** Make sure any components used within theme icons (or any other page icons we render with this kind of encapsulation) get properly detected and registered. For context, I ran into issues where the `<wa-input>` in our theme icons was not being registered because the page was not using another `<wa-input>` and our autoloader does not look inside shadow roots.

This also updates palette icons to use the same mechanism as theme icons, which makes them easier to generate dynamically as they can now just use a regular CSS grid, rather than having to reimplement grids with SVG + Nunjucks. This makes it a lot easier to generate icons dynamically, which is necessary for tweaked (and later custom) palettes. As an additional benefit, this also makes them more easily adapt to having variable scales, which looks like it will be needed for custom palettes.

I’ve tried to style them as closely as the SVG ones we had, but using our design tokens meant slightly different results. E.g. this is the smallest border-radius we have, and it still looks a lot bigger than the current icons. Of course we could always just not use tokens here. @lindsaym-fa what do you think?

Current:
image

New:
image

@lindsaym-fa feel free to make any changes you see fit (in fact, one of the reasons I extracted this out as a separate PR is to get your help with the design details of these palette icons)

### DIFF
--- a/docs/_includes/svgs/palette.njk
+++ b/docs/_includes/svgs/palette.njk
@@ -1,31 +1,20 @@
 {% set paletteId = palette.fileSlug or page.fileSlug %}
 {% set suffixes = ['-80', '', '-20'] %}
-{% set width = 20 %}
-{% set height = 12 %}
-{% set height_core = 20 %}
-{% set gap_x = 4 %}
-{% set gap_y = 4 %}
 
-{% set total_width = (width + gap_x) * hues|length %}
-{% set total_height = (height + gap_y) * suffixes|length + (height_core - height) %}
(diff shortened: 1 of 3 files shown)


lindsaym-fa (approved):
Looks good! Pushed a small change to reduce the whitespace between swatches. With that in place, I think the corners work just fine.

PR #869, opened 2025-04-14
update patterns to isPro

KonnorRogers:


claviska (approved):

LeaVerou:
I’m surprised this was needed, since https://github.com/shoelace-style/webawesome/blob/next/docs/docs/patterns/patterns.json has `isPro: true`. Is this so that the middleware can pick it up? (asking to see if I need to update other stuff too)

claviska:
@LeaVerou Yes, it was for middleware.

@KonnorRogers is it worth it to update the path finder utility to include all Eleventy data? Seems like that might be tricky. 🤔

KonnorRogers:
> I’m surprised this was needed, since https://github.com/shoelace-style/webawesome/blob/next/docs/docs/patterns/patterns.json has `isPro: true`. Is this so that the middleware can pick it up? (asking to see if I need to update other stuff too)

admittedly, i just did something quick, it probably isn't needed. I didn't see the `patterns.json` 🤦‍♂️

LeaVerou:
> @KonnorRogers is it worth it to update the path finder utility to include all Eleventy data? Seems like that might be tricky. 🤔

Where is the path finder utility? I could take a look too.

> @LeaVerou Yes, it was for middleware.

Gotcha, thanks. If updating the pathfinder utility is too tricky, then we'd need to manually add `isPro` to other pro assets as well, not just patterns.

KonnorRogers:
> > @KonnorRogers is it worth it to update the path finder utility to include all Eleventy data? Seems like that might be tricky. 🤔
> 
> Where is the path finder utility? I could take a look too.
> 
> > @LeaVerou Yes, it was for middleware.
> 
> Gotcha, thanks. If updating the pathfinder utility is too tricky, then we'd need to manually add `isPro` to other pro assets as well, not just patterns.

The path finder utility isn't do anything to propagate the data. Its all 11ty.

https://github.com/shoelace-style/webawesome-app/blob/21d7e3ff4dba1c05310167d855cc532571ee9e6f/.eleventy.js#L22-L24

claviska:
> admittedly, i just did something quick, it probably isn't needed. I didn't see the patterns.json 🤦‍♂️

Not to be _that guy_, but I recently argued [unsuccessfully] _in favor or_ using 11ty as an "engine," which meant limiting our usage to specific features for consistency (since 11ty does sooo much).

Setting global data in `eleventy.config.js` was one of those points, and the reasoning was, well, things like this get overlooked. <3

LeaVerou:
> > admittedly, i just did something quick, it probably isn't needed. I didn't see the patterns.json 🤦‍♂️
> 
> Not to be _that guy_, but I recently argued [unsuccessfully] _in favor or_ using 11ty as an "engine," which meant limiting our usage to specific features for consistency (since 11ty does sooo much).
> 
> Setting global data in `eleventy.config.js` was one of those points, and the reasoning was, well, things like this get overlooked. <3

The thing is, it's much harder to set global data that way that only apply to specific directories, that's why 11ty has directory data. I do agree that 11ty's data cascade can be very complicated to grok though. What about a middle ground? E.g. we could agree on a subset of 11ty's data features to use so that we don't need to duplicate data across many pages (and risk forgetting in some down the line) but also don't have to look _everywhere_ to figure out what data a page has?

PR #874, opened 2025-04-17
Add horizontal orientation to `<wa-card>`

### LINKED ISSUE #534: Horizontal cards
Horizontal cards are pretty common, yet we currently only do vertical cards. Seems like we need an `orientation="horizontal"` perhaps. 

Low priority for now, but is needed for patterns in Week 6.

image

Image

Image

Image

kelseythejackson:
Screenshot 2025-04-17 at 6 43 43 PM
Still experimenting. There are lots of questions to answer and scenarios to solve for.

Closes #534


claviska:
@kelseythejackson FYI — I merged `next` into this and cleaned up the SSR-only attributes, e.g. `with-header`, `with-footer`, etc. per #917.

claviska:
Great initiative and start!

> There are lots of questions to answer and scenarios to solve for.

What questions need to be answered to help move this one forward? :)

claviska:
@lindsaym-fa what else do we need for this one?

claviska:
Let's talk about adding an actions slot that moves to the "end" of the card in horizontal.

paulGeoghegan:
I personally would love this. Definitely let me know if I can help contribute or test.

claviska (commented):
Nice!

CleanShot 2025-09-03 at 15 39 04@2x

nit: we should probably remove the padding on the paragraph to visually center the text.

---


> Set the orientation attribute to horizontal to create a card with a horizontal, side-by-side layout. Make sure to set a width or maximum width for the media slot. Horizontal cards do not currently contain the header and footer slots.

Even though we mention this in the Slots section, we should probably add a note about the `actions` slot here as well.

---

It looks like the `actions` slot is only available in horizontal mode. If that's the case _and if we don't have plans to add support for it in vertical cards_, does it make sense to reuse the `footer` slot? (Feels weird, I'll admit, but both are technically rendered at the logical end of the container.)

Or, we _could_ consider adding support for it in vertical cards as well. (Prepare for some wild ASCII art.)

```
Vertical

|--------------------|
|       Media        |
----------------------
| Header  |  Actions |
|--------------------|
|       Body         |
|--------------------|
|      Footer        |
----------------------

- The `media` slot is optional and will render 
  above the header when provided
- The actions slot is optional but requires the 
  header slot to be present

Horizontal

----------------------------------
|         |          |           |
|  Media  |   Body   |  Actions  |
|         |          |           |
----------------------------------

- Media is optional
- Actions are optional (no need to provide 
  header in horizontal mode)
- There is no footer
```

Although that does make the API a bit more conditional, which can also be confusing 🤔 

I'd be curious what @kelseythejackson and @lindsaym-fa think would be best from a UX perspective.

lindsaym-fa:
☝️ Put that ASCII in the MoMA.

I like the idea of having `actions` available for vertical/default cards too. fwiw, this is similar to `wa-dialog` and `wa-drawer`, both of which support a `header-actions` slot that is only visible when the header is present. (Of course, since it's not _always_ dependent on the header being present, it makes sense to use the more generic `actions` name.)

packages/webawesome/src/components/card/card.css
```diff
@@ -102,3 +102,27 @@
...
+    border-start-end-radius: 0;
+  }
+}
+
+:host([orientation='horizontal']) ::slotted(*) {
+  display: block;
+  height: 100%;
+  margin: 0;
+  object-fit: cover;
+}
```
lindsaym-fa:
```suggestion
:host([orientation='horizontal']) ::slotted(*) {
  display: block;
  height: 100%;
  margin: 0;
  object-fit: cover;
}
```

I'm not sure we want these styles to target all slotted elements. That results in issues like this when there are multiple elements in the default slot:
Screenshot 2025-09-04 at 4 57 05 PM

Seems like `display`, `height`, and `margin` should apply to `.body` and `object-fit` should apply to only the elements slotted into `.media`, right?

kelseythejackson: Yeah, the plan is to have both header and footer actions. Problem is, you can't nest slots, so it would structurally change the card. That seems like a major change, but if we can explain it, it should be fine.

kelseythejackson: Oooh, thanks for catching that. that's been updated

claviska:
Looks good to me. The only thing left is adding a description here, e.g. "Set the `orientation` attribute to `horizontal` to...".

CleanShot 2025-09-09 at 15 23 33@2x

It's probably also a good idea to add a separate example showing how the actions slot works with vertical _and_ horizontal cards, especially since there's a bit of variability in that API.

kelseythejackson:
weird, that description is there on my branch, and I've updated the example with the different action slots

KonnorRogers (approved):

PR #883, opened 2025-04-23
[Image-comparer] Several fixes + rename to `comparer`

### LINKED ISSUE #495: [Image-comparer] Drop `image-` from name
The image-comparer component can be used for non-image comparisons. E.g. in our theme pages, we use it to compare a theme's light & dark modes. The naming should reflect this, which would also make it shorter.

Ideas:
- `<wa-compare>`
- `<wa-comparison>`

### LINKED ISSUE #563: `<wa-image-comparer>` should apply `isolation: isolate` to each element
See #561 for an example of the breakage not having this can cause.

### LINKED ISSUE #882: Image-comparer: Arrow keys inside interactive content move the divider
To reproduce:
- Visit [a theme page on the deployed website](https://backers.webawesome.com/docs/themes/default/)
- Focus on the input
- Press the left or right arrow keys. 

The divider moves, but it shouldn't.

This is an easy fix: This happens because the component listens to keydown events on the wrapper, instead of the divider, which is the actual focusable element. Moving the `keydown` listener to the divider fixes it.

LeaVerou:
1. Rename `<wa-image-comparer>` to `<wa-comparer>`, closes #495 
2. Apply `isolation: isolate` to non-image slotted content, closes #563.
3. Move `keydown` listener to the divider, fixes #882
4. Drop `base` part (part of #207). Pretty straightforward, since the base part was just a div, so the styling just moved to the host.

Note that the diff is large due to the renaming, but each fix has been implemented as a separate commit, so just view individual commits to review them separately.


docs/docs/resources/changelog.md
```diff
@@ -12,6 +12,10 @@ Components with the <wa-badge variant="warning" pill>Experimental</wa-badge> bad
 During the alpha period, things might break! We take breaking changes very seriously, but sometimes they're necessary to make the final product that much better. We appreciate your patience!
 :::
 
+## Next
+
+- Renamed `<image-comparer>` to `<wa-comparer>` and improved compatibility for non-image content.
```
claviska:
```suggestion
- 🚨 BREAKING: Renamed `<image-comparer>` to `<wa-comparer>` and improved compatibility for non-image content.
```

claviska (approved):
Thanks for updating this one. Looks good!

PR #885, opened 2025-04-23
[Skeleton] Remove `base` part, rel #207

LeaVerou:
Completely straightforward, as the base was just a div.


claviska (commented):
Thanks for tackling this. Everything looks good, but the avatar example is no longer rendering in this branch for some reason.

CleanShot 2025-04-28 at 12 30 09@2x

PR #892, opened 2025-04-29
fix select

KonnorRogers:
i quickly poked at https://github.com/shoelace-style/webawesome-alpha/issues/230

and found there's a divergent code path for manually set value properties, and values set via selecting an option.

### DIFF
--- a/src/components/select/select.ts
+++ b/src/components/select/select.ts
@@ -679,7 +679,7 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
     // Toggle values present in the DOM from this.value, while preserving options NOT present in the DOM (for lazy loading)
     // Note that options NOT present in the DOM will be moved to the end after this
     if (selectedValues.size > 0 || this._value) {
-      const oldValue = this._value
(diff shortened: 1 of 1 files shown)


src/components/select/select.ts
```diff
@@ -173,11 +173,12 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
       this._value = val;
       let newValue = this.value;
 
-      if (newValue != oldValue) {
+      if (newValue !== oldValue) {
```
KonnorRogers, line 176: @LeaVerou any reason this used coercive equals?

claviska (approved):

src/components/select/select.ts
LeaVerou: No idea. I suspect it's an oversight, otherwise I'd have left a comment explaining why the non-strict equals.

LeaVerou (approved):

src/components/select/select.ts
KonnorRogers: 🚢

PR #894, opened 2025-04-29
Add font metadata to themes + a couple `typography.css` fixes

LeaVerou:
This PR adds metadata about fonts used on each theme page (just like we do for `brand` and `palette`). 
Metadata that is the same as the default theme is omitted. Same with metadata we don't want to expose (e.g. the typography of Classic).

Also, a couple `typography.css` fixes:
- Add `:where([class^='wa-theme-'], [class*=' wa-theme-'])` to default `typography.css` since we declare mappings that need to be inherited by other themes (such as `--wa-font-family-heading: var(--wa-font-family-body);`).
- Remove declarations that redeclare defaults
- Ensure consistent order (`body` before `heading`)
- [Awesome] Switch code font to default (`ui-monospace`)

### DIFF
--- a/src/styles/themes/playful/typography.css
+++ b/src/styles/themes/playful/typography.css
@@ -6,8 +6,7 @@
 .wa-theme-playful {
   --wa-font-family-body: 'Nunito', sans-serif;
   --wa-font-family-heading: 'Fredoka', sans-serif;
-  --wa-font-family-code: 'Azeret Mono', monospace;
-  --wa-font-family-longform: ui-serif, serif;
+  --wa-font-family-code: 'Azeret Mono', ui-monospace, monospace;
 
   --wa-font-size-scale: 1.125;
 
--- a/docs/docs/themes/active.md
+++ b/docs/docs/themes/active.md
@@ -5,4 +5,6 @@ isPro: true
 tags: pro
 palette: rudimentary
 brand: green
+fonts:
+  body: Inter
 ---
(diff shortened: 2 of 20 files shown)


src/styles/themes/playful/typography.css
```diff
@@ -6,8 +6,7 @@
 .wa-theme-playful {
   --wa-font-family-body: 'Nunito', sans-serif;
   --wa-font-family-heading: 'Fredoka', sans-serif;
-  --wa-font-family-code: 'Azeret Mono', monospace;
-  --wa-font-family-longform: ui-serif, serif;
+  --wa-font-family-code: 'Azeret Mono', ui-monospace, monospace;
```
lindsaym-fa, line 9: I'm curious why we're changing this particular value to fall back to `ui-monospace, monospace` whereas we only have a single generic fallback (`sans-serif`, `serif`, or `monospace`) for all other values. Is there a reason we need both the `ui-*` and generic fallbacks here but not elsewhere?

lindsaym-fa (approved):
Looks good! Left just one minor comment.

src/styles/themes/playful/typography.css
LeaVerou: My thinking was that it's a better fallback if the web font doesn't load (I'd argue the same for `sans-serif`) but it's not a strong opinion at all and I don't mind removing it. Should I?

lindsaym-fa: Totally fine with that logic! Just curious why we wouldn't implement the same choice everywhere we're using web fonts. This is the only instance I've seen.

PR #899, opened 2025-04-29
Fix `<wa-dialog>` width example

claviska:
We were using `width` instead of `--width`. This is a quick docs fix _unless_ `width` is intentional and is supposed to work. I don't think this is the case, though, as the `--width` property is still documented.

Fixes https://github.com/shoelace-style/webawesome-alpha/issues/258


LeaVerou:
We reverted the dialog to not use `display: contents` in a previous PR, so this must be leftover.

PR #904, opened 2025-05-01
Improve native radio alignment

claviska:
This was [reported back in Feburary](https://github.com/shoelace-style/webawesome-alpha/issues/171). This PR improves, but doesn't fully solve, the problem shown below.

However, instead of being broken in all sizes (not pictured), it's now broken only in the small size.

To get this close, I had to switch from `lh` to `em` for sizing it. I also switched to using top/left 50% + translate left -50% instead of `scale`.

I believe the bigger problem has something to do with calculations + rounding `--wa-size-small`, but I didn't want to go that deep into the rabbit hole today.

This doesn't close the original issue, but it does improve it but I'd like to get another pair of eyes before moving this out of draft.

/cc @lindsaym-fa @LeaVerou 

### Before

CleanShot 2025-05-01 at 15 26 23@2x

### After

Chrome
CleanShot 2025-05-01 at 15 26 53@2x

Firefox
CleanShot 2025-05-01 at 15 27 52@2x


lindsaym-fa:
Certainly much better for the default theme!

One issue that I can look into: We have a number of themes that set `--checked-icon-scale` for WA and native radios. Without using this custom property in the native radio styles, the results aren't identical between `<wa-radio>` and `<input type="radio">`. A handful of examples below, though most themes are affected. I'll see if there's some way we can incorporate `--checked-icon-scale` or some alternative without negatively affecting alignment.

**Classic**
Screenshot 2025-05-05 at 12 28 13 PM

**Active**
Screenshot 2025-05-05 at 12 28 29 PM

**Premium**
Screenshot 2025-05-05 at 12 29 22 PM

src/styles/native/radio.css
```diff
@@ -35,9 +35,11 @@ input[type='radio'] {
     border-radius: 50%;
     content: '';
     position: absolute;
```
lindsaym-fa:
```suggestion
```

src/styles/native/radio.css
```diff
@@ -35,9 +35,11 @@ input[type='radio'] {
...
-    width: calc(var(--toggle-size) - var(--border-width));
-    height: calc(var(--toggle-size) - var(--border-width));
-    scale: var(--checked-icon-scale);
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    width: calc(100% - 0.25em);
+    aspect-ratio: 1;
```
lindsaym-fa:
```suggestion
    width: round(calc(100% * var(--checked-icon-scale)), 1px);
    aspect-ratio: 1;
```

src/styles/native/radio.css
```diff
@@ -10,7 +10,7 @@ label:has(input[type='radio']),
   --box-shadow: none;
   --checked-icon-color: var(--wa-form-control-activated-color);
   --checked-icon-scale: 0.75;
-  --toggle-size: 1lh;
+  --toggle-size: 1.25em;
```
lindsaym-fa:
```suggestion
  --toggle-size: round(1lh, 1px);
```

lindsaym-fa (commented):
@claviska, how do we feel about using `round()` ([which is well-supported](https://caniuse.com/mdn-css_types_round)) to help browsers avoid splitting pixels? Plus, I think we can rely on the flex properties on the parent to position the checked icon rather than absolute positioning. (I've added suggestions for both of these changes.)

Test results for native radios in each browser for the Default theme —

**Chrome**
Screenshot 2025-05-05 at 5 21 41 PM

**Edge**
Screenshot 2025-05-05 at 5 21 54 PM

**Firefox**
Screenshot 2025-05-05 at 5 22 03 PM

**Safari**
Screenshot 2025-05-05 at 5 22 13 PM

claviska:
Looks great. I’ll get this in tomorrow and see how it works!

claviska:
Beautiful! Thanks, @lindsaym-fa. Tested in all browsers and I can verify this resolves it.

PR #907, opened 2025-05-01
Scroller component

claviska:
This PR adds a scroller component to facilitate scrolling horizontal and vertical containers with overflow. It provides a tab stop around the content, an accessible label, and keyboard support for scrolling.

[Preview this component](https://webawesome-git-scroller-font-awesome.vercel.app/docs/components/scroller)

This component _does not_ use scroll animations or anything fancy, but it works in all browsers. It can absolutely be adapted in the future as browser support improves.

This can be one of the 14 free stretch goal components if we want it to be. It would be a quick win if we're comfortable bringing it in and using it.

If so, I can finish up the docs for it and get this PR wrapped up.

### TODO

- [x] Initial port
- [x] Update translations
- [x] Add examples to docs
- [x] Design review / approval

### Preview

Light mode
CleanShot 2025-05-01 at 17 38 43@2x

Dark mode
CleanShot 2025-05-01 at 17 39 03@2x


lindsaym-fa:
Agreed that we definitely want some affordance of scrollability in dark mode and that white shadows _probably_ aren't it. 😅 I'll see what else I can think of.

Edit with my own notes: 
Shadows in dark mode are a bit tricky. To maintain realism, they still need to be dark; white shadows don't exist IRL and, because of this, yield a glow effect on screen. 

The reason why we bake opacity into `--wa-color-shadow` is because shadows need to become more opaque in dark mode to have a noticeable effect against the darker backgrounds. Nevertheless, it might be wise to have `--wa-color-shadow` handle *only* the actual color (as the name implies) and introduce another token for `--wa-shadow-opacity` to handle...well, opacity, and set this differently for light vs. dark modes. 

In that case, we could use `--shadow color: var(--wa-color-shadow);` and `--shadow-opacity: var(--wa-shadow-opacity);` in the scroller component and fully leverage our themes. I'll explore this!

src/components/scroller/scroller.styles.ts
```diff
@@ -0,0 +1,209 @@
+import { css } from 'lit';
```
LeaVerou: This is using the old `*.styles.ts` format, we should convert to a `scroller.css` file.

src/components/scroller/scroller.styles.ts
```diff
@@ -0,0 +1,209 @@
...
+      );
+      pointer-events: none;
+    }
+
+    #start-shadow {
+      &:dir(ltr) {
+        left: 0;
+        border-left: solid var(--edge-width)
+          color-mix(in oklab, var(--edge-color) calc(var(--start-shadow-opacity) * 100%), transparent);
+        background: linear-gradient(
```
LeaVerou: A radial gradient tends to look better as it goes in smoothly at the edges, @lindsaym-fa could perhaps help define it.

src/components/scroller/scroller.styles.ts
```diff
@@ -0,0 +1,209 @@
+import { css } from 'lit';
+
+export default css`
+  :host {
+    --edge-color: var(--wa-color-neutral-border-soft);
+    --edge-width: var(--wa-border-width-s);
+    --shadow-color: var(--wa-color-text-normal);
```
LeaVerou:
```suggestion
    --shadow-color: var(--wa-color-shadow);
```

docs/docs/components/scroller.md
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+</style>
+```
+
+## Examples
+
+### Orientation
+
+Set the `orientation` attribute to `vertical` and provide a height to create a vertical scroller.
```
LeaVerou, line 111: Would it make sense to instead have the component figure this out automatically (and possibly keep `orientation` as an override)?
What happens if the scroller is scrollable in both directions? Do we ever want shadows on both axes or is it always just one?

src/components/scroller/scroller.styles.ts
claviska: fixed

src/components/scroller/scroller.styles.ts
claviska: This completely removes the shadow. I agree that the text token isn't the best here, but I imagine the design will change.

docs/docs/components/scroller.md
claviska: That might be too much magic, but could be a later add should users want it. This component took inspiration from [HeroUI](https://www.heroui.com/docs/components/scroll-shadow), which uses the `orientation` property the same way.

claviska:
Looking at [Hero UI's scroller design](https://www.heroui.com/docs/components/scroll-shadow), maybe it's best to remove the shadow and do a background fade, which works better on dark mode. We can expose the shadow parts (not currently done) to allow custom shadows and show an example of doing that in the docs.

Any thoughts, @lindsaym-fa?

CleanShot 2025-05-12 at 11 11 24@2x

CleanShot 2025-05-12 at 11 11 13@2x

lindsaym-fa:
@claviska Yes! A fade is definitely the right call. Shadows are really challenging to work with any text + background color in dark mode.

claviska:
Thanks, @lindsaym-fa. This is ready for review now!

**A nice to have:** a new illustration for this component 😄

CleanShot 2025-05-14 at 13 08 18@2x

### Inspiration from AI

<details>
<summary>Show me those hideous renditions</summary>

image
image

</details>

lindsaym-fa (commented):
Works great! 🎉

There's an annoyingly minor issue in Safari only where a single pixel peaks out from behind the starting shadow in `orientation="horizontal"`. 🙄
Screenshot 2025-05-14 at 6 02 50 PM

Adding `overflow: hidden;` to the `:host` element fixes it. Is that style worth adding?

claviska:
@lindsaym-fa Safari pixels fixed in [5e52af5](https://github.com/shoelace-style/webawesome/pull/907/commits/5e52af54f20a03d7b88e6348d6283540caea4773). Thanks!

lindsaym-fa (approved):
All good! I also added an illustration —
Screenshot 2025-05-15 at 4 04 44 PMScreenshot 2025-05-15 at 4 04 53 PM

Approved! ✅

claviska:
@lindsaym-fa that looks way better than the AI-generated ones. Thanks!

PR #908, opened 2025-05-02
Merge `default` and `system` libraries,  fixes #901

### LINKED ISSUE #901: System icons are blank for family != classic or variants other than regular/solid
This came out of #893. Splitting it out so we can track its status separately.

The problem was spotted by @lindsaym-fa in https://github.com/shoelace-style/webawesome/issues/893#issuecomment-2839478486 : 

> [@LeaVerou](https://github.com/LeaVerou) It doesn't look like setting `--wa-icon-family` and `--wa-icon-variant` works 100% with system icons (`/src/components/icon/library.system.ts`).
> 
> Because system icons expect `family = 'classic'`, when I set `--wa-icon-family: sharp;` or `--wa-icon-family: duotone;` for a theme, icons that are built into components (like the chevrons in `<wa-details>`, the checkmark in `<wa-checkbox>`, etc.) don't render.
> 
> How do we want to handle this?

It is not a new bug, but it was exposed by the fact that we can now set defaults for family/variant for a whole subtree or page whereas previously you'd have to manually specify the library and/or variant on the `<wa-icon>` instance to uncover it, which would be tough for icons within components.

Solutions by @LeaVerou in https://github.com/shoelace-style/webawesome/issues/893#issuecomment-2839815140 :

> Ooh, that's a hairy one. Some ideas:
> 
> 1. We ignore `family` in the `system` library resolver. This is on par with how other single-family icon libraries work — if they don't support multiple families they just ignore this parameter.
> 
> * ✅ This is by far the lowest effort solution, all it takes is removing a conditional.
> * ✅ We maintain the benefit that these load quickly.
> * ❌ This means we get FA Classic icons for these regardless of what the theme sets, which is …not great.
> 
> 2. The `system` library falls back to `default` if the icon is not available.
> 
> * ✅ This would give us icons that use the correct family.
> * ❌ When using another family, we lose the benefit that these load quickly
> * ❌ Unclear where this logic would live. In the icon component? In icon libraries? In the system library? Every option seems architecturally poor and/or violates [DIP](https://en.wikipedia.org/wiki/Dependency_inversion_principle).
> 
> 3. Thinking some more, **the `system` library is not a separate library at all — it's a few icons from `default` that are hardcoded for performance**. So perhaps the right solution is to ditch the `system` library entirely and allow icon libraries to provide pre-fetched SVG for icons, then just provide the system icons as pre-fetched icons of the `default` library.
> 
> * ✅ Instead of adding complexity, it **removes complexity** and is IMO a better conceptual design overall.
> * ✅ Benefits all icon libraries in addition to solving the problem at hand.
> * ✅ The caching benefits uses of the `default` library too, not just when `system` is explicitly used.
> * ✅ We no longer need to remember which icons are part of system and which ones are not.
> * ✅ Can speed up _all_ icon usage, for all families and all libraries.
> * ✅ By allowing libraries to provide explicit SVG markup we can skip the roundabout step of converting the SVG into a data URI which is then stuffed into an SVG `<use>` element that the system library currently does.
> * ❌ This is also the highest effort solution, though not insurmountable. I just spent some time looking into it and I think the tradeoff is worth it, though I'd love to hear what [@claviska](https://github.com/claviska) thinks, since he knows what rabbit holes may lurk in that part of the code.

All three of us agreed on spending a few hours in this direction and if it doesn't pan out / turns into more work, going with the quick fix of 1.

LeaVerou:
With this…
- We no longer need to do `library="system"` to take advantage of prefetched FA icons, it happens automatically when using an icon that has been prefetched.
- System icons will no longer be blank when we're using a different default family/variant, the problem that prompted this (more details in #901)
- As a bonus, this enables things like:
  - All icon libraries can now declare prefetched icons
  - Prefetched icons can be added later as well (which also means icons used in just one component could actually live there rather than one place in the code needing to know about the icon needs of every single WA component)

It implements:
- Pre-fetched icons for all libraries via the `fetched` property!
- `registeredLibrary.addFetched()` to add more prefetched icons to an existing library (needed for extensibility story, see `contributing.md`)
- Update icon docs
- Update changelog
- Separate unregistered and registered library types and move all relevant types to separate `.d.ts` file so that libraries don't have to import `library.ts`. Down the line we may want to make `RegisteredIconLibrary` a class, but I was trying to keep the changes minimal.
- Kill `library="system"` with 🔥 from everywhere

I've done the latter as a separate commit, so that one can review the actual substantive changes separately by viewing the [first commit](https://github.com/shoelace-style/webawesome/pull/908/commits/ac1d412a8fd1bf0e26fac579c9ba1b82c28255ee).

Requesting review from 
- @claviska for the overall code architecture
- @claviska and @KonnorRogers for some help with the (relatively few) TS issues I had to `@ts-ignore` / `@ts-expect-error` and with tests.
- @lindsaym-fa for a second pair of eyes in case you can spot any regressions, since icons are so core to WA.


docs/docs/components/icon.md
```diff
@@ -183,7 +183,8 @@ Custom icons can be loaded individually with the `src` attribute. Only SVGs on a
 
 You can register additional icons to use with the `<wa-icon>` component through icon libraries. Icon files can exist locally or on a CORS-enabled endpoint (e.g. a CDN). There is no limit to how many icon libraries you can register and there is no cost associated with registering them, as individual icons are only requested when they're used.
 
-Web Awesome ships with two built-in icon libraries, `default` and `system`. The [default icon library](#customizing-the-default-library) is provided courtesy of [Font Awesome](https://fontawesome.com/). The [system icon library](#customizing-the-system-library) contains only a small subset of icons that are used internally by Web Awesome components.
+Web Awesome ships with one built-in icon library, `default`.
```
claviska:
```suggestion
Web Awesome ships with one built-in icon library: `default`.
```

docs/docs/components/icon.md
```diff
@@ -627,18 +628,57 @@ For security reasons, browsers may apply the same-origin policy on `<use>` eleme
...
+      }
+    }
+  }
+});
+```
+
+::: warning
+Please note that sprite sheets and fetched icons are mutually exclusive.
+If you set the `spriteSheet` property to `true`, the `fetched` property will be ignored.
+:::
```
claviska: I can't submit this as a suggestion due to deleted lines, but I took a stab at making this section a bit more clear. Hopefully I didn't miss anything :)

**Edit:** had to add \ next to backticks to get it to show in a single code field.
---

```md
### Prefetched icons { #fetched }

Prefetched icons are SVG icons that are loaded directly from code rather than from HTTP requests, making them load faster in your application. Prefetching icons improves performance by eliminating HTTP requests, reducing load times, and ensuring system icons are immediately available.

#### How to Use Prefetched Icons**

The `fetched` property allows you to provide icon SVG markup directly in your code. This creates a mapping of `(name, library, variant)` to SVG markup:

\```html
<script type="module">
  import { registerIconLibrary } from '/dist/webawesome.js';

  registerIconLibrary('default', {
    resolver: name => `/path/to/custom/icons/${name}.svg`,
    fetched: {
      check: '<svg xmlns="http://www.w3.org/2000/svg">...</svg>',
    }
  });
</script>
\```

#### Adding More Prefetched Icons

To add additional icons to an existing library, use the `addFetched` method:

\```js
import {getIconLibrary} from '/dist/webawesome.js';

let defaultLibrary = getIconLibrary('default');
defaultLibrary.addFetched({
  classic: { // family
    regular: { // variant
      'circle-info': '<svg xmlns="http://www.w3.org/2000/svg">...</svg>',
      'triangle-exclamation': '<svg xmlns="http://www.w3.org/2000/svg">...</svg>'
      // ...
    }
  }
});
\```

#### Icon Library Structure

- For simple libraries (no variants): `{ name: svg }`
- For complex libraries (with family and variant): `{ family: { variant: { name: svg } } }`

#### Best Practice

When using custom icon libraries with Web Awesome components, always prefetch system icons to ensure optimal performance. Refer to `src/components/library.default.ts` for a complete list of system icons used internally.

::: warning
Note that sprite sheets and fetched icons are mutually exclusive. If you set the `spriteSheet` property to `true`, the `fetched` property will be ignored.
:::
```

docs/docs/resources/contributing.md
```diff
@@ -352,13 +352,12 @@ Form controls should support submission and validation through the following con
 
 ### System Icons
 
-Avoid inlining SVG icons inside of templates. If a component requires an icon, make sure `<wa-icon>` is a dependency of the component and use the [system library](/components/icon#customizing-the-system-library):
+Avoid inlining SVG icons inside of templates.
+If a component requires an icon, make sure `<wa-icon>` is a dependency of the component.
```
claviska, line 356:
```suggestion
Avoid inlining SVG icons inside of templates. If a component requires an icon, make sure `<wa-icon>` is a dependency of the component.
```

docs/docs/resources/contributing.md
```diff
@@ -352,13 +352,12 @@ Form controls should support submission and validation through the following con
...
 
-```html
-<wa-icon library="system" name="..." variant="..."></wa-icon>
-```
-
-This will render the icons instantly whereas the default library will fetch them from a remote source. If an icon isn't available in the system library, you will need to add it to `library.system.ts`. Using the system library ensures that all icons load instantly and are customizable by users who wish to provide a custom resolver for the system library.
+If it is not one of the [pre-fetched icons](/components/icon#fetched) in the `default` library,
+you should add it.
+This will render the icons instantly rather than fetching them from a remote source.
```
claviska:
```suggestion
If it is not one of the [pre-fetched icons](/components/icon#fetched) in the `default` library,
you should add it. This will render the icons instantly rather than fetching them from a remote source.
```

src/utilities/deep.ts
```diff
@@ -0,0 +1,83 @@
...
+  let { filter, transformPath } = options;
+  let entries: any[][] = [];
+
+  deepEach(obj, (value, key, parent, path) => {
+    let included = filter?.(value, key, parent, path) ?? true;
+
+    if (included) {
+      let fullPath = [...path, key];
+      path = transformPath?.(fullPath) ?? fullPath;
+      entries.push([...fullPath, value]);
```
claviska: Is this supposed to be `entries.push([...path, value])`? If not, what is `path` for?

src/utilities/deep.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Iterate over a deep array, recursively for plain objects
+ * @param obj The object to iterate over. Can be an array or a plain object, or even a primitive value.
+ * @param callback. value is === parent[key]
+ * @param parentObj The parent object of the current value Mainly used internally to facilitate recursion.
+ * @param key The key of the current value. Mainly used internally to facilitate recursion.
+ * @param path Any existing path (not including the key). Mainly used internally to facilitate recursion.
+ */
+export function deepEach(obj: any, callback: EachCallback, parentObj?: object, key?: Property, path: Property[] = []) {
```
claviska: I know this is intended for icons, but I can see us reaching for this in the future so it might be worth calling out that this will stack overflow if the object has a circular ref.

src/utilities/deep.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Like Object.entries, but for deeply nested objects.
+ * For shallow objects the output is the same as Object.entries.
+ * @param obj The object to iterate over.
+ * @param options.filter - If this returns false, the entry is not added to the result.
+ * @param options.transformPath - Transform the path, e.g. to serialize it to a single key.
+ * @returns Array of arrays. In each array the last value is the value, and all values before that are the keys.
+ */
+export function deepEntries(obj: any, options: DeepEntriesOptions = {}): any[][] {
```
claviska: Same concern re: circular refs

lindsaym-fa (commented):
Nice! Only noticed one regression: we appear to be missing our indeterminate icon in checkboxes. See [Checkbox > Indeterminate](https://webawesome-git-icon-fetched-font-awesome.vercel.app/docs/components/checkbox/#indeterminate) and [Tree](https://webawesome-git-icon-fetched-font-awesome.vercel.app/docs/components/tree/). 

Screenshot 2025-05-02 at 12 19 02 PM
Screenshot 2025-05-02 at 12 19 14 PM

Screenshot 2025-05-02 at 12 23 30 PM

src/components/icon/icon.ts
```diff
@@ -136,16 +136,23 @@ export default class WaIcon extends WebAwesomeElement {
...
+      markup = library.fetched[url] as string;
+    } else {
+      try {
+        fileData = await fetch(url, { mode: 'cors' });
+        if (!fileData.ok) return fileData.status === 410 ? CACHEABLE_ERROR : RETRYABLE_ERROR;
+        markup = await fileData.text();
+      } catch {
+        return RETRYABLE_ERROR;
+      }
     }
```
claviska: Are we planning on using `library.fetched` as a cache or just as a prefetch?

- If the former, the name sounds fine but we should probably cache the fetched icon for next time
- If the latter, perhaps `prefetch` or another name is more appropriate

src/components/icon/types.d.ts
```diff
@@ -0,0 +1,33 @@
...
+export type IconLibraryMutator = (svg: SVGElement) => void;
+
+// This is a utility for decrementing a number up to 3 by one
+// e.g., Decrement[3] yields 2
+type Decrement = [never, 0, 1, 2];
+
+// Record of string → string or nested Record<string, up to (Min, Max) levels deep>
+export type IconLibraryCache<Min extends number = 1, Max extends number = Min> = Min extends 0
+  ? string | (Max extends 0 ? never : Record<string, IconLibraryCache<0, Decrement[Max]>>)
+  : Record<string, IconLibraryCache<Decrement[Min], Max>>;
```
claviska: When types get complex like this, it's good to ask if there's a simpler way. I'm not saying there _is_ in this case, but this is a bit hard to follow so a refactor or more thorough commenting might help.

It's also not clear from IntelliSense how I should use `fetched`, e.g.:

CleanShot 2025-05-02 at 12 43 17@2x

src/components/icon/library.ts
```diff
@@ -30,13 +26,21 @@ export function getIconLibrary(name?: string) {
...
-export function registerIconLibrary(name: string, options: Omit<IconLibrary, 'name'>) {
+export function registerIconLibrary(name: string, options: UnregisteredIconLibrary) {
   unregisterIconLibrary(name);
+
   registry.push({
     name,
     resolver: options.resolver,
     mutator: options.mutator,
     spriteSheet: options.spriteSheet,
+    fetched: flattenIconLibraryCache(options.resolver, options.fetched),
```
claviska: The `flattenIconLibraryCache()` function is kinda hard to follow. Can we get away with simpler typings for prefetched icons by making the expected object more rigid?

If there's a use case that requires the current implementation, let's add a more verbose comment explaining how the function works so it doesn't seem so magical :)

docs/docs/resources/contributing.md
LeaVerou: Is the only change deleting the line break? I was using [semantic line breaking](https://sembr.org/) for smaller diffs. Do you think that's not a good idea?

claviska: I was looking at the output in the docs. It should either continue or have a second break to separate it visually — although I think markdown is more forgiving about this. It just makes it easier for us humans to read :)

src/components/icon/library.ts
LeaVerou: I have an idea about how to simplify, let me try it out.

docs/docs/resources/contributing.md
claviska: (Ugh, just read more about semantic line breaking and it seems like a solution looking for a problem.)

src/components/icon/icon.ts
LeaVerou: Oh good point. Currently it's only being read, but we could totally read from it as well. `prefetch` sounds like a function name to me though.

docs/docs/resources/contributing.md
LeaVerou: In my experience, it can be a **lifesaver** in cases where many people are managing a lot of text, especially when it involves lots of small edits. But perhaps the gains here are smaller. And I've definitely seen it be taken to the extreme where source becomes unreadable because of too much line breaking, which sucks just as much.
My rule of thumb is to try and break on natural opportunities like periods and (for longer sentences) commas and not overdo it. I'm fine not following it if you don't like it, I was just explaining where the habit comes from 😊

src/utilities/deep.ts
LeaVerou: Okay, this was the push I needed to refactor these so that `path` actually includes `key`. I already suspected it was too weird keeping them separate.

src/utilities/deep.ts
LeaVerou: Better than adding a note about the problem is fixing it 😁 I'll just implement a visitor using a WeakSet so it doesn't have this issue.

src/components/icon/types.d.ts
LeaVerou: Oh is that how it shows up? That's very strange.

LeaVerou:
@lindsaym-fa 
> Nice! Only noticed one regression: we appear to be missing our indeterminate icon in checkboxes. See [Checkbox > Indeterminate](https://webawesome-git-icon-fetched-font-awesome.vercel.app/docs/components/checkbox/#indeterminate) and [Tree](https://webawesome-git-icon-fetched-font-awesome.vercel.app/docs/components/tree/).
> 
> Screenshot 2025-05-02 at 12 19 02 PM Screenshot 2025-05-02 at 12 19 14 PM Screenshot 2025-05-02 at 12 23 30 PM

Does this happen locally too? It appears it might be due to FA CORS. I cannot reproduce locally, but I've also made changes, so not sure if I inadvertently also fixed this or it was a Vercel-only issue.

lindsaym-fa:
@LeaVerou I'm seeing the indeterminate icon both locally and in the Vercel preview now, so it looks like you fixed it!

LeaVerou:
Okay, this is ready for another round!

@cory, I made several changes to address your feedback:
- Moved most of the logic of `flattenIconLibraryCache()` into a more general `flatten()` utility in `deep.ts`. `flattenIconLibraryCache()` is still there, but hopefully it is now a lot clearer (and more well commented)
- Simplified the `IconLibraryCache` type. It's still recursive, but instead of specifying a range it now specifies an explicit depth, and if we want a range we can do a union (`IconLibraryCache<0> | IconLibraryCache<1> | ...`)
- Refactored all deep utils so that now there is no separate `path` and `key` being passed around, only one `path` that includes the `key` as well (one can still get the key via `path.at(-1)`).
- Refactored `deepEach()` to avoid circular references (via a `WeakSet` of visited objects)
- Incorporated your docs suggestions (with some small tweaks)

@lindsaym-fa, can you please confirm that it works locally? I think the issue you were seeing might have been specific to the Vercel preview (compounded by the bug I describe below).

Some additional changes:

1. I noticed that FA icons were not always cached properly. Turns out that, because I was using the icon URL as the cache key and that changed when there was a kit token, it meant it was susceptible to race conditions, meaning if the `default` library got registered before the kit token was added, all the fetched icons would be effectively thrown away! To address this, I added a `getKey()` parameter that defaults to `options.resolver` if not provided and separated the code that gets the base URL and the code that gets the final URL in `library.default.ts` so that the former can be used as the cache key.
2. As discussed with @claviska, `registerIconLibrary()` now takes a single object literal instead of a separate `name` and `options` object. This means icon libraries can export a single object that just gets registered. If folks want to register an alias, they can always do `{...library, name: 'foobar'}`.

docs/docs/resources/contributing.md
LeaVerou: Just noticed this:

> I was looking at the output in the docs. It should either continue or have a second break to separate it visually — although I think markdown is more forgiving about this. It just makes it easier for us humans to read :)

Do we have this setting on in Markdown where line breaks in the source become `<br>`s? I was assuming it's off

lindsaym-fa:
Confirmed locally and on Vercel. ✅

claviska:
I noticed this while [perusing the new docs:](https://webawesome-68d56ff7b-font-awesome.vercel.app/docs/components/icon/#customizing-the-default-library)

> For example, this will change the default icon library to use [Bootstrap Icons](https://icons.getbootstrap.com/) loaded from the jsDelivr CDN.

While this is still true, it will now break many of the [previously known as] system icons because they don't share the same names between FA and Bootstrap (or whatever non-FA icon library they may be using).

We need to have a better story for users who want to customize the default library _without_ borking the "system" icons. 🤔

claviska:
> We need to have a better story for users who want to customize the default library without borking the "system" icons. 🤔

At first, I thought this might be as simple as showing users how to import the `fetched` icons in the docs and spreading it into their own library. But then we have the possibility of naming collisions.

Another option would be to prefix the "system" icons with something unique, but that defeats the purpose of caching. At that point, keeping them separate would have been a clearer line.

I'm not really sure how to move forward with this one now. I'd love to hear thoughts, but with this knowledge maybe the system library was a good idea after all.

src/components/comparer/comparer.ts
```diff
@@ -133,7 +133,7 @@ export default class WaComparer extends WebAwesomeElement {
           tabindex="0"
         >
           <slot name="handle">
-            <wa-icon library="system" name="grip-vertical" variant="solid"></wa-icon>
+            <wa-icon name="grip-vertical" variant="solid"></wa-icon>
```
KonnorRogers: What happens if a user has an icon library without this exact mapping?

`grip-vertical`


What if i use a different library that is like `vertical-grip-hand` ? How do we resolve this? There's no possible way for us to infer this. And its going to just look like the icon is broken.

Feels like we need a mapping of internal icons that a user would be able to re-map for their own icon library. (Which to my understanding, was roughly what the system icon library allowed for)

claviska:
Thinking about this more, we can't guarantee icon names will be consistent across libraries, so we definitely need some sort of system library here. While there _is_ a level of consistency within FA-curated icon names, FA will soon be supporting select third-party icon libs that use their own names and conventions.

This is a contrived example, but consider a user who wants to use a custom icon library. It has a `user` icon, which is currently a system icon in WA. What if they want to use `robot` for the system icons but not for normal icons? Without the system library, this is no longer possible. The only way would be to set `user` in their icon library to resolve to `robot`, which would break their library. 😕 

With the system library intact, they can set `user` to output their `robot` SVG and be done with it.

However, the bigger problem this PR aimed to solve was:

**Why doesn't changing the icon library in Web Awesome also change the icons in the components?**

This seems like the question to focus on. Perhaps we leave the system library intact and, from within each icon library we provide, we configure both `default` and `system` libraries so when you switch, you get a matching set by default _as well as_ the ability to customize them separately.

For free users (no themer), they can still define `default` and `system` separately.

(Sorry…I really wish I would have realized this limitation sooner!)

claviska:
[From this post:](https://github.com/shoelace-style/webawesome/issues/901#issue-3033859024)

> All three of us agreed on spending a few hours in this direction and if it doesn't pan out / turns into more work, going with the quick fix of 1.

I'm not sure how much time you've spent on this already, but for me it's turned into more than a few hours of _reviewing_. 😬 

Reflecting a bit, [a previous PR](https://github.com/shoelace-style/webawesome/pull/447) made icons configurable through CSS which [introduced a problem](https://github.com/shoelace-style/webawesome/issues/901#issue-3033859024) that caused system icons to not display for certain variants.

I'm specifically calling it a "problem" because it is now but it wasn't before. The reason I bring this up is because our ability to make icons configurable through the themer was already present and it's actually harder to do now reliably.

The APIs for customizing the default and system icon libraries are here:

- [Customizing the default library](https://backers.webawesome.com/docs/components/icon#customizing-the-default-library)
- [Customizing the system library](https://backers.webawesome.com/docs/components/icon#customizing-the-system-library)

Remember that the system library is designed for internal use. The only thing end users will care about is that the icons in the components match the ones they select in the themer and that they don't "flash" on first load.

We can do this very easily by expanding `library.system.ts` into multiple system library files to reflect each family/variant we support. Then we'll set both the default icon library and the system icon library to point to the desired resolvers/files at runtime using the above APIs.

This code can be trivially generated by the themer. It feels like we keep chasing clever solutions, but the problem was already solvable with little effort and without any additional maintenance burden. 😕 

At this point, I suggest not merging this PR and rolling back #447 so system icons never break when the family/variant changes. In fact, making system icons match the default library is definitely not a blocker for next week's release, so why don't we table that detail and improve it in a subsequent split?

LeaVerou:
The idea was that when you adjust the default library/variant, you _want_ the system icons to adapt. 
- If we don't, then that's a different (much easier!) problem. We don't have to roll back #447 for that, we can simply adjust the system icons resolver so that if we have an icon with that name it always resolves.
- If we do, @lindsaym-fa also suggested that as a quick hack we could actually embed all FA families and styles for now, though I'm not sure if we _can_ do that wrt licensing.

I could also separate out the icon libraries refactor from the new API surface, as there is value in that independently of the API changes.

claviska:
> The idea was that when you adjust the default library/variant, you want the system icons to adapt.

Yes, let's hold off on this _for now_. We received permission to embed FA Pro icons in _WA Pro_, but not WA Free. Alas, the pro repo isn't available yet.

I know people will ask for this. We do ultimately want default + system to match by default, so we'll prioritize this in a subsequent split once the WA Pro repo is fully in place.

TL;DR – for this split, let's just update the _default_ icon library and let the system icons remain FA free.

LeaVerou:
@claviska Ok, on it! Edit: [done](https://github.com/shoelace-style/webawesome/pull/921)

claviska:
Closing as discussed in today's check-in.

PR #913, opened 2025-05-07
Remove alpha flags

claviska:
- Removes `npm run start:alpha` and similar alpha scripts
- Removes `isAlpha` and `noAlpha`
- Removes alpha filters for 11ty
- Removes `data-alpha` 11ty plugin
- Removes the `show` filter

@KonnorRogers two things:

- Did I miss anything? 😅
- Does this impact the app in any way?

### TODO

- [x] Update Vercel to use `npm run build` instead of `npm run build:alpha` when this merges


KonnorRogers:
> Does this impact the app in any way?

nope. App was using the `--alpha` flag, but i removed that already.

> Did I miss anything? 😅

The github actions test config is using `build:alpha`

https://github.com/shoelace-style/webawesome/blob/d94589d113e5868194157507c0a8da0b83bd888c/.github/workflows/client_tests.js.yml#L34

https://github.com/shoelace-style/webawesome/blob/d94589d113e5868194157507c0a8da0b83bd888c/.github/workflows/ssr_tests.js.yml#L34

>  Update Vercel to use npm run build instead of npm run build:alpha when this merges

We actually dont need the Web UI for this. 

https://github.com/shoelace-style/webawesome/blob/next/vercel.sh

claviska:
@KonnorRogers thanks! Scripts updated in a9fd4ac7b36ca06119ad638f29906732dac03528.

KonnorRogers (approved):

PR #917, opened 2025-05-09
Remove ssr attributes since we're not using SSR

claviska:
Removing SSR-specific `with-` attributes to prevent confusion as promised in https://github.com/shoelace-style/webawesome/issues/261#issuecomment-2864244993


LeaVerou (dismissed):
YESSSSSS kill them with 🔥

PR #919, opened 2025-05-10
add unlisted and unpublished support

KonnorRogers:
- Proper support for `unlisted: true` and will only serve those files in development, adds support for `excludeFromSearch: true` which prevents things from being added to `search.json`


LeaVerou (dismissed):

LeaVerou:
LGTM!

So now everything with `unlisted: true` is actually removed from the build, right?

Two things:
1. We should set `unlisted: true` to the things that were previously `noAlpha`. From memory, some of these were:
	- Components cheatsheet
	- Code demo component
	- Viewport demo component
2. This _may_ cause a problem with the themer as it depends on having an `unlisted` page that is actually accessible (via a redirect). But perhaps just setting `eleventyExcludeFromCollections` on it might do the job — or worst case we add an `excludeFromIndices` or whatever flag.

KonnorRogers:
> We should set unlisted: true to the things that were previously noAlpha. From memory, some of these were:
>         Components cheatsheet
>         Code demo component
>         Viewport demo component


I can make the component cheat sheet, code demo, and viewport demo unlisted.

> This may cause a problem with the themer as it depends on having an unlisted page that is actually accessible (via a redirect). But perhaps just setting eleventyExcludeFromCollections on it might do the job — or worst case we add an excludeFromIndices or whatever flag.

Maybe I'm misunderstanding, but I think `excludeFromSearch` accomplishes this? The idea being you have an HTML file you need to serve in production, but dont really want people visiting that page directly? Or at least don't want them finding it via the search menu.

LeaVerou:
> > We should set unlisted: true to the things that were previously noAlpha. From memory, some of these were:
> > Components cheatsheet
> > Code demo component
> > Viewport demo component
> 
> I can make the component cheat sheet, code demo, and viewport demo unlisted.
> 
> > This may cause a problem with the themer as it depends on having an unlisted page that is actually accessible (via a redirect). But perhaps just setting eleventyExcludeFromCollections on it might do the job — or worst case we add an excludeFromIndices or whatever flag.
> 
> Maybe I'm misunderstanding, but I think `excludeFromSearch` accomplishes this? The idea being you have an HTML file you need to serve in production, but dont really want people visiting that page directly? Or at least don't want them finding it via the search menu.

Search is only one place it could show up. There's also the sidebar, the overview pages etc. In most systems I've seen that's what `unlisted` does: it still deploys the page but it's not included in any listing. WordPress is one example that comes to mind using it this way.
Suggestion, take it or leave it: making `unlisted` behave like this: not showing up anywhere, but the page still being deployed, and calling the flag that actually removes the content something else (`unreleased`?)

claviska:
> Suggestion, take it or leave it: making unlisted behave like this: not showing up anywhere, but the page still being deployed, and calling the flag that actually removes the content something else (unreleased?)

I agree that "unlisted" sounds like the page would be built, but not shown anywhere.

Do we need `unlisted` and `hidden` perhaps?

Also, can we start a section in docs/contributing with supported front matter for our own cheat sheet?

KonnorRogers:
> > Suggestion, take it or leave it: making unlisted behave like this: not showing up anywhere, but the page still being deployed, and calling the flag that actually removes the content something else (unreleased?)
> 
> I agree that "unlisted" sounds like the page would be built, but not shown anywhere.
> 
> Do we need `unlisted` and `hidden` perhaps?

`unpublished`, `devOnly`,  take your pick. 🤷 

I only used `unlisted` because it was what we had previously. But i agree its not clear if "unlisted" means "dont build this" vs "build this, but dont add it to search"

> Also, can we start a section in docs/contributing with supported front matter for our own cheat sheet?

Yes absolutely. We have quite a number at this point that we should probably document somewhere.

claviska:
> unpublished, devOnly, take your pick. 🤷

No strong opinions on it as long as we document it and use them consistently :)

claviska (dismissed):

KonnorRogers:
Renaming to:

```
unlisted: true # => Builds the file on npm run build, but will not add it to search, and will act like a "noindex"
unpublished: true # => Will not be a part of `npm run build`
```

Cannot `excludeFromEleventyCollection` programmatically with `unlisted` unfortunately. So will need to document you need both. :/

https://github.com/11ty/eleventy/discussions/2211

Unless @zachleat knows some secret sauce I don't.

claviska (dismissed):

claviska (approved):

PR #922, opened 2025-05-13
[Comparer] Expose wrapper as part, fix choppy dragging

LeaVerou:
This PR fixes 2 `<wa-comparer>` issues that came up while implementing the dark mode preview in the themer:
1. The wrapper was not exposed as a part. I exposed it as `base`.
2. When the elements being compared were iframes (I suspect the same issue would exist with `<object>` too) dragging was incredibly choppy as the iframe kept capturing it. With this, both compared elements get `pointer-events: none` while dragging is happening.

### DIFF
--- a/src/components/comparer/comparer.ts
+++ b/src/components/comparer/comparer.ts
@@ -23,6 +23,7 @@ import styles from './comparer.css';
  *
  * @event change - Emitted when the position changes.
  *
+ * @csspart base - The container that wraps the before and after content.
  * @csspart before - The container that wraps the before content.
  * @csspart after - The container that wraps the after content.
  * @csspart divider - The divider that separates the before and after content.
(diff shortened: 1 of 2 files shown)


src/components/comparer/comparer.ts
```diff
@@ -96,7 +97,7 @@ export default class WaComparer extends WebAwesomeElement {
     const isRtl = this.hasUpdated ? this.localize.dir() === 'rtl' : this.dir === 'rtl';
 
     return html`
-      <div class="image">
+      <div class="image" part="base">
```
claviska, line 100: Is this intentional? If so, I'm not married at all to the name "base" so if there's something more appropriate, feel free to use it.

_History: I struggled a lot with "base" and eventually settled on it but never loved it as a name._

claviska (approved):
Nice optimization! I left a comment, but it's not a blocker.

src/components/comparer/comparer.ts
LeaVerou: I figured we use it a lot elsewhere so it seemed a good fit for a generic wrapper. I'm not married to it at all. If you have a better suggestion, shoot! 

(I'm not even sure we need it, but trying to keep the change minimal)

PR #923, opened 2025-05-13
Revert addition of Awesome's `dimension.css`

lindsaym-fa:
As discussed in this week's tactical, the Awesome theme's dimensionality isn't currently compatible with other themes. Instead, its dimensionality is likely a combination of zero shadow diffusion and additional button borders, the latter of which needs more exploration before it can be replicated in the themer and the compatibility ensured for all themes. As such, this ticket reverts the addition of `styles/themes/awesome/dimension.css` from #918.

Additionally, this change includes a fix for an issue where buttons were using the wrong border color when dark mode is achieved using `.wa-invert`. Here are the results when in light mode using `.wa-invert` —

**Before**
Screenshot 2025-05-13 at 5 56 49 PM

**After**
Screenshot 2025-05-13 at 5 55 59 PM

Figured I'd fix this issue while I was moving the styles containing the bug around. 😊


LeaVerou (approved):
LGTM! Thanks for making the changes in separate commits so I could review the fix separately!

claviska (approved):

PR #925, opened 2025-05-14
Fix nav toggle behavior in `<wa-page>`

claviska:
[Reported by a backer:](https://github.com/shoelace-style/webawesome-alpha/issues/268#issuecomment-2877392854)

@KonnorRogers I'm not as familiar with the internals of `<wa-page>` as I'd like to be, so give this a good review. I didn't notice any side effects :)

<details>
<summary>Fixture to reproduce and test</summary>

```html
<wa-page disable-navigation-toggle style="outline: dashed 2px gray;">
  <header slot="header" class="wa-split">
    <div class="wa-cluster wa-gap-xs">
      <wa-icon src="/img/logo-play.svg" style="font-size: 2.5em;"></wa-icon>
      <span id="brand-name" class="wa-heading-s wa-desktop-only">{{ metadata.author.name }}</span>
    </div>
    <div class="wa-cluster wa-gap-xs">
      <wa-button href="#" size="small" variant="brand" appearance="outlined">
        <wa-icon slot="prefix" name="calendar-days" variant="solid"></wa-icon>
        Prossimi Eventi
      </wa-button>
      <wa-button href="#" size="small" variant="brand">
        <wa-icon slot="prefix" name="crown" variant="solid"></wa-icon>
        VIP
      </wa-button>
    </div>
  </header>
  <main>
    {{ content | safe }}
  </main>
</wa-page>
```

</details>

### DIFF
--- a/src/components/page/page.ts
+++ b/src/components/page/page.ts
@@ -201,8 +201,9 @@ export default class WaPage extends WebAwesomeElement {
     super.connectedCallback();
     this.pageResizeObserver.observe(this);
 
-    // Wait a cycle for the initial update to complete
-    requestAnimationFrame(() => {
+    // Wait a cycle for the initial update to complete. We use `setTimeout` intentionally to ensure DOM children also
+    // have a chance to update.
+    setTimeout(() => {
       this.headerResizeObserver.observe(this.header);
       this.subheaderResizeObserver.observe(this.subheader);
       this.bannerResizeObserver.observe(this.banner);
--- a/src/components/page/page.ts
+++ b/src/components/page/page.ts
@@ -199,24 +199,22 @@ export default class WaPage extends WebAwesomeElement {
 
   connectedCallback() {
     super.connectedCallback();
-
     this.pageResizeObserver.observe(this);
 
-    const navQuery = ":not([slot='toggle-navigation']) [data-toggle-nav]";
-
-    // check once on initial connect
-    // eslint-disable-next-line
-    this.disableNavigationToggle = Boolean(this.querySelector(navQuery));
-
-    setTimeout(() => {
+    // Wait a cycle for the initial update to complete
+    requestAnimationFrame(() => {
       this.headerResizeObserver.observe(this.header);
       this.subheaderResizeObserver.observe(this.subheader);
       this.bannerResizeObserver.observe(this.banner);
       this.footerResizeObserver.observe(this.footer);
 
-      // Check again when the element updates
-      // eslint-disable-next-line
-      this.disableNavigationToggle = Boolean(this.querySelector(navQuery));
+      if (this.hasAttribute('disable-navigation-toggle')) {
+        this.disableNavigationToggle = true;
+      } else {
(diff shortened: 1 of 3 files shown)


src/components/page/page.ts
```diff
@@ -199,24 +199,22 @@ export default class WaPage extends WebAwesomeElement {
...
-      // Check again when the element updates
-      // eslint-disable-next-line
-      this.disableNavigationToggle = Boolean(this.querySelector(navQuery));
+      if (this.hasAttribute('disable-navigation-toggle')) {
+        this.disableNavigationToggle = true;
+      } else {
+        // Otherwise, use auto-detection again
+        const navQuery = ":not([slot='toggle-navigation']) [data-toggle-nav]";
+        this.disableNavigationToggle = Boolean(this.querySelector(navQuery));
+      }
```
KonnorRogers: This should probably use `setTimeout` only because `RaF` can sometimes "miss" in large DOMs and isn't always accurate of when the DOM has finished.

KonnorRogers (dismissed):
Okay. I've chewed this over quite a bit. I know this was meant to be an easy fix, but there's some major holes in how `disableNavigationToggle` works. It was originally intended as an SSR property to be able to hint that the toggle won't exist since it doesnt have query selectors.

The problem with this implementation is a few things:

1.) There is no way to get this behavior if a `data-toggle-nav` is added in a different event cycle (IE: dynamically added)

2.) Overwriting users properties / attributes is always icky.

3.) The naming is horrible and makes it seem like an easy way to disable the toggle button.

Proposal:

Use `disableNavigationToggle` / `disable-navigation-toggle` purely as something the user sets.  Make a second reflected attribute: `hasNavigationToggle` which tracks if a `data-toggle-nav` was added. (OR since its intended to be internal, we can make it private and just add it internally so it doesn't get reflected)

`hasNavigationToggle` should also use a proper MutationObserver to track new changes.

I'm happy to do this work in a follow up split, just needed to get my ideas down.

TLDR: This PR is good for now, but we probably need to re-evaluate how `disableNavigationToggle` works, because its really poorly named.

src/components/page/page.ts
claviska: Is the goal to wait until first render? Would `this.updateComplete.then(() => ... )` be even better?

KonnorRogers: yes the goal is to wait for ~~first render~~ the light dom to finish updating. I dont believe `updateComplete` accounts for DOM children, only for the shadow DOM to finish updating.

claviska:
> yes the goal is to wait for first render the light dom to finish updating. I dont believe updateComplete accounts for DOM children, only for the shadow DOM to finish updating.

I've reverted to `setTimeout` and added a comment explaining why.

> TLDR: This PR is good for now, but we probably need to re-evaluate how disableNavigationToggle works, because its really poorly named.

Should we rename it? Revert it? What do you suggest to move this forward?

claviska:
Closing per discussion with @KonnorRogers about solving this more correctly in the near future.

PR #927, opened 2025-05-14
Add slot detection back to dialog

### LINKED ISSUE #926: 🪲 `wa-dialog` and `wa-drawer` are missing headers and footers
### Describe the bug
Dialogs and drawers across our docs are missing their headers (label + close button) and footers.

Here's one of our dialogs on the backers site:
Image

And the same dialog on `next`:
Image

### To Reproduce

1. Got to /docs/components/dialog (e.g. https://webawesome-git-next-font-awesome.vercel.app/docs/components/dialog/)
2. Open any of the example dialogs in the documentation
3. Observe that the dialog label + close button and any content slotted into the dialog footer are missing

### Browser / OS
Chrome 135.0.7049.115, Safari 18.4, Firefox 138.0.3

claviska:
Fixes #926 by reimplementing dialog slot detection

### DIFF
--- a/docs/_includes/search.njk
+++ b/docs/_includes/search.njk
@@ -1,4 +1,4 @@
-<wa-dialog id="site-search" no-header light-dismiss>
+<wa-dialog id="site-search" without-header light-dismiss>
   <div id="site-search-container">
     {# Header #}
     <header>
--- a/docs/docs/components/dialog.md
+++ b/docs/docs/components/dialog.md
@@ -27,19 +27,20 @@ keywords: modal
 
 ## Examples
 
-### Dialog with Header
+### Dialog without Header
 
-Headers can be used to display titles and more. Use the `with-header` attribute to add a header to the dialog.
+Headers are enabled by default. To render a dialog without a header, add the `without-header` attribute.
 
 ```html {.example}
-<wa-dialog label="Dialog" class="dialog-header">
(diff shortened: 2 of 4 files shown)


LeaVerou:
Two totally non-blocking suggestions: 

A more portable pattern that we already use elsewhere is to add a `has-slotted` class on the `<slot>` element itself, and then use `.has-slotted` on the slot or `:has([slot=slotname].has-slotted)` on ancestors. We already use this pattern for hints, so you can search for `has-slotted` to see the code.

This has some advantages:
1. It's a pattern that once you understand it once, you don't need to understand it again
2. We could even automate it with a helper later (e.g. elements could define which slots they want to do slot detection on and the class is automatically managed)
3. Once `:has-slotted` is widely supported (currently at [67%](https://caniuse.com/mdn-css_selectors_has-slotted)), it becomes trivial to move to it across the whole codebase in one go. And even before then, we can use `:is(:has-slotted, .has-slotted)` and only apply the class as a polyfill.

WRT naming, we already have 4 other `no-*` attributes:
image

And one `disable-*` attribute (`disable-navigation-toggle`), but we don't have a `without`, so maybe `no-header` was a better name (shorter too)? OTOH I can see the connection with the `with-*` attributes.

lindsaym-fa (approved):
All looks good!

Issue #928, opened 2025-05-14
Reimplement slot detection


claviska:
Slot detection was removed awhile ago as a prerequisite for SSR support, but it turns out we didn't need to remove it.

https://github.com/shoelace-style/webawesome/commit/e8f7342cb68e232eaec9f73b32190fa648a9e0d3

Adding it back will make the experience better for non-SSR users, which is currently everybody.

- [x] dialog - fixed in #927
- [x] drawer - fixed in #931
- [x] button
  - [x] also fix #842
- [x] alert (now callout) (not necessary)
- [x] breadcrumb item (not necessary)
- [x] card
- [x] menu item (not necessary)
- [x] dialog
- [x] drawer

claviska:
Not all components needed this, but the ones that did have been updated.

PR #931, opened 2025-05-15
add slot detection back to drawer; #928

claviska:
- Fixes broken drawers
- Light cleanup of BEM-style class names in dialog/drawer
- Removed unused classes in dialog/drawer


lindsaym-fa (dismissed):
Works great! Thanks!

PR #958, opened 2025-05-21
Add browser + screen reader support policy; fixes #146

### LINKED ISSUE #146: Clearly define our browser support policy, including which browser + screen reader combinations we support
We need to formalize this because screen readers are finicky. Our decision should be based on the most common and reliable combinations. Bugs in other combos can be addressed on an issue-per-issue basis.

As a small team, we unfortunately can't afford spending too much time on edge cases that only impact a small number of users.

claviska:
Added a page to call out browser support and screen reader support.

Improvements welcome.

[View page](https://webawesome-git-browser-support-policy-font-awesome.vercel.app/docs/resources/browser-support)


lindsaym-fa (approved):
Lovely!

PR #959, opened 2025-05-21
Rename the classic theme to shoelace; fixes #909

### LINKED ISSUE #909: Rename the "Classic" theme to "Shoelace"
This was discussed in the strategy call last week. @claviska agreed that "Classic" is a bit misleading; it sounds like "classic (i.e. timeless) style" where it actually means Shoelace. We can't use Shoelace for legacy reasons, so perhaps "Legacy" might be a better name? We generally want to discourage use of this in new projects, since it mainly exists to make onboarding Shoelace users easier.

claviska:
See #909


lindsaym-fa (approved):
Looks good! I also pushed a change to change the Shoelace description from:
> Timeless elegance that never goes out of style.

To:
> The original, familiar look you know and love from Shoelace.

PR #963, opened 2025-05-21
Rename `vertical` to `orientation="vertical"`; fixes #674

### LINKED ISSUE #674: `vertical` → `orientation="vertical"`
Currently, 2 components do orientation via a boolean `vertical` attribute (divider, split-panel) and 3 via an `orientation` attribute (button-group, carousel, radio-group).

Let's stick with `orientation`, per https://github.com/shoelace-style/webawesome/discussions/269 but also because that allows different components to have different default orientations and still retain the same syntax. This also allows us to down the line make it inherit in some contexts where that makes sense.

- [x] `<wa-divider>`
- [x] `<wa-split-panel>`

claviska:
See #674

- [x] Divider
- [x] Split panel


lindsaym-fa (dismissed):
@claviska I made a few commits to update the docs, patterns, `divider.css`, and `split-panel.css` to use `orientation='vertical'`. Calling that out in case you wanted to give these changes a once-over.

Approved otherwise!

PR #964, opened 2025-05-21
Add min-block-size to divider; fixes #675

### LINKED ISSUE #675: `<wa-divider vertical>` in horizontal flex container is invisible
This was not discovered by me — I’m just relaying it.

Testcase: https://codepen.io/leaverou/pen/NPKVjxV/abbd1b91cf1d6facf0116fa8ffd4ef83?editors=1000

Perhaps a `min-block-size: 1lh` for vertical dividers?

claviska:
See #675

- A height is no longer required on the flex container, which was a common pain point


src/components/divider/divider.css
```diff
@@ -15,4 +15,5 @@
   height: 100%;
   border-left: solid var(--width) var(--color);
```
lindsaym-fa:
```suggestion
  border-inline-start: solid var(--width) var(--color);
```

_Very_ minor, but I figure if we're setting `min-block-size`, we may want to use the logical size here, too.

Issue #375, opened 2024-12-20
Borders missing from button groups


LeaVerou:
Next:
<img width="238" alt="image" src="https://github.com/user-attachments/assets/88aecc9d-bf67-4638-8a53-e3388e2e259e" />

Backers:
<img width="240" alt="image" src="https://github.com/user-attachments/assets/27aa26c5-024e-4a0a-b1cb-3785c334ab82" />

claviska:
I know you don't like button group being a component, and especially this little hack:

```js
button.classList.add('button');
button.classList.toggle('wa-button-group-horizontal', this.orientation === 'horizontal');
button.classList.toggle('wa-button-group-vertical', this.orientation === 'vertical');
button.classList.toggle('button--first', index === 0);
button.classList.toggle('button--inner', index > 0 && index < slottedElements.length - 1);
button.classList.toggle('button--last', index === slottedElements.length - 1);
button.classList.toggle('button--radio', button.tagName.toLowerCase() === 'wa-radio-button');
```

I didn't like it either, but I found it necessary because I couldn't achieve what I wanted to with pure CSS. This was the only reliable way to get the styles button group needed for each use case.

Despite styles being changed, this logic still exists in button group. I'd suggest reverting the styles to what was working in backers as the quickest solution.

This is something that just shouldn't have broken between then and now :(

claviska:
It looks like [these classes were changed inadvertently](https://github.com/shoelace-style/webawesome/commit/5a82b7ad992d5f3f80a8c29c12e6176dbc3c7e2b#diff-7e68ca42b21f58307fc9647655fb8c99dc71cca299e21f4770725e752c03ba06L77-R80).

```diff
-        button.classList.toggle('wa-button-group__button--first', index === 0);
-        button.classList.toggle('wa-button-group__button--inner', index > 0 && index < slottedElements.length - 1);
-        button.classList.toggle('wa-button-group__button--last', index === slottedElements.length - 1);
-        button.classList.toggle('wa-button-group__button--radio', button.tagName.toLowerCase() === 'wa-radio-button');
+        button.classList.toggle('button--first', index === 0);
+        button.classList.toggle('button--inner', index > 0 && index < slottedElements.length - 1);
+        button.classList.toggle('button--last', index === slottedElements.length - 1);
+        button.classList.toggle('button--radio', button.tagName.toLowerCase() === 'wa-radio-button');
```

[...]

PR #967, opened 2025-05-21
Remove old BEM -- convention from classes and fix button group dividers

### LINKED ISSUE #308: Replace parts with --modifiers with regular parts or custom states
Not comprehensive, but some examples are below.

### Components that use parts as state modifiers

```js
/** checkbox - current
 * @csspart control
 * @csspart control--checked
 * @csspart control--indeterminate
 */

/** checkbox - proposed
 * @csspart control
 * 
 * @cssstate checked
 * @cssstate indeterminate
 */
```

### Components that misuse `--` for valid part names

```js
/** carousel.ts - current
 * @csspart navigation-button
 * @csspart navigation-button--previous
 * @csspart navigation-button--next
 */

/** carousel.ts - proposed
 * @csspart navigation-button
 * @csspart navigation-button-previous
 * @csspart navigation-button-next
 */
```

---

- [x] Audit components for offending parts
- [ ] Convert misused parts to valid part names
- [ ] Convert stateful parts to custom states

claviska:
- Removes the BEM pattern `thing--thing` in favor of `thing-thing`, which was already in use in various locations
- Fixes #308 (there was only one part that was still doing this)
- Most of these changes are internal. I added a changelog entry for the part that was affected.


claviska:
I also pushed a fix for #375. @lindsaym-fa I couldn't think of a token to use, so it's currently hardcoded as a 1px gap. Let me know if you have a better suggestion for this one.

Before
before

After
after

src/styles/utilities/button-group.css
lindsaym-fa, line 1: A gap is great for filled and accent buttons, but it doesn't yield the best results for outlined buttons.

Screenshot 2025-05-21 at 2 40 41 PM

That said, I wasn't able to come up with anything foolproof after a while of tinkering... If we supported the `appearance` attribute on `wa-button-group`, we could simply change the `gap` accordingly. Is that something we'd want to support?

src/styles/utilities/button-group.css
```diff
@@ -33,7 +34,6 @@
     --wa-border-inline-start-radius: 0;
 
     border-inline-start-color: var(--border-color, rgb(0 0 0 / 0.3));
```
lindsaym-fa:
```suggestion
```

I don't believe this style has any effect.

src/styles/utilities/button-group.css
```diff
@@ -51,18 +51,16 @@
 
   > :not(:first-child),
   &::slotted(:not(:first-child)) {
+    --wa-border-block-start-radius: 0;
     border-start-start-radius: 0 !important;
     border-start-end-radius: 0 !important;
-    --wa-border-block-start-radius: 0;
-
     border-block-start-color: var(--border-color, rgb(0 0 0 / 0.3));
```
lindsaym-fa:
```suggestion
```

Same here — this doesn't seem to have any effect.

lindsaym-fa (dismissed):
Approved with some optional comments to consider.

claviska:
Good catch. I added a check that says "if any outlined button is in the button group, collapse the gap and overlap the border."

CleanShot 2025-05-22 at 06 55 38@2x

The caveat: mixing outlined and non-outlined buttons means you won't have any gaps at all. If it becomes a concern for users, we can add more logic and styles to account for mixing.

src/styles/utilities/button-group.css
claviska: See https://github.com/shoelace-style/webawesome/pull/967#issuecomment-2900795320

lindsaym-fa (approved):

Issue #535, opened 2025-01-15
Rename the `image` slot in `<wa-card>` to `media`


LeaVerou:
It seems like we mainly use for interchangeable purposes, and pick one or the other based on the styling we need (full width vs regular). E.g. in our overview pages we use `header` for icons instead of `image` (which _sounds_ more fitting!) because we don't want the default styling that comes with the `image` slot. Is that actually the case? If so, should we rename `image` to something more general that communicates its purpose better? `hero` comes to mind, but not sure if that is widely understood by our users or esoteric jargon.

claviska:
@kelseythejackson I'm suggesting `media` to replace the poorly named `image` slot in `<wa-card>`. Do you have any other suggestions for it?

claviska:
@kelseythejackson this would be a good one to add to #874 if you're interested.

kelseythejackson:
I'll probably do this work on it's own branch

claviska:
Fixed in #975

PR #975, opened 2025-05-21
updated card slot from `image` to `media`

kelseythejackson:
This is related to #535.
It changes the `image` slot in the card component to `media`. It's broader and more appropriate.


src/components/card/card.css
```diff
@@ -50,10 +50,10 @@
   border-end-end-radius: var(--inner-border-radius);
 }
 
-.image {
+.media {
   display: flex;
 
-  &::slotted(img) {
+  &::slotted(*) {
```
kelseythejackson, line 56: I worry if using the wildcard here is too generic and needs to be more specific

claviska, line 56: At first, I wasn't sure if this was too eager and should be scoped down to `img`, `video`, and `iframe` but it _is_ called the "media" slot, so I actually think it's fine as-is.

claviska (dismissed):
Nice work. Looks great

src/components/card/card.css
claviska: oh weird this came in right as I posted this! haha https://github.com/shoelace-style/webawesome/pull/975#discussion_r2100840579

claviska:
@kelseythejackson oh this one thing this needs a line in the changelog :)

src/components/card/card.css
claviska: That said, if you think it's too generic, I'm fine with limiting it to certain elements!

claviska (approved):

Issue #943, opened 2025-05-19
Simplify native styles and component styles


claviska:
Many styles were abstracted and imported as makeshift utilities between native styles and components. This has led to a lot of chaos, so we're going to undo it.

Benefits include:

- More modular code
- Knowing exactly what styles affect what things
- Easier to maintain
- Prevents [things like this](https://github.com/shoelace-style/webawesome/issues/494)

This is a prerequisite to and might even solve:

- #684
- #627
- #626
- #392

---

- [x] Combine native styles into one CSS file: `native.css`
- [x] Use CSS layers to make them easily overridable by users
- [x] Remove the abstractions and put styles back into component CSS
- [x] Remove appearance classes from native styles and remove from documentation
  > There is no reason to emulate with classes what our components already do. Let's remove these and make things simpler for users and more consistent.
  - [x] Native styles > `.wa-button`
  - [x] Native styles > `.wa-callout`
  - [x] `.wa-textfield` / `.wa-textarea` (is there one?)
- [x] Revert passthrough styles
- [x] Eliminate `src/styles/shadow/passthrough.css`
- [x] Look at `src/styles/utilities/*`
- [x] Look at `src/styles/utilities/variants.css.js`
- [x] Make sure we have default spacing for icons in native buttons, e.g. https://github.com/shoelace-style/webawesome/issues/453 
- [x] Revert details styles #684
- [x] Side quest: determine if any other CSS files should use layers
   - [x] `src/styles/brand/*`
   - [x] `src/styles/color/*`
   - [x] `src/styles/danger/*`
   - [x] `src/styles/neutral/*`
   - [x] `src/styles/success/*`
   - [x] `src/styles/themes/*`
   - [x] `src/styles/utilities/*`
   - [x] `src/styles/warning/*`

---

### Follow ups & Questions

- [x] Should images have a subtle rounded edge by default?
- [x] Native button appearances aren't working. Which classes do we actually want to support before telling them to use `<wa-button>`?
- [x] Radio Buttons are styled wrong now. Instead of fixing, let's complete #504 and add `appearance="button"` to `<wa-radio>`
- [x] Fix broken button appearances (visible on button.md)
- [x] Fix `appearance="filled"` in textarea. It currently works in input but the appearance styles are baked into various CSS files and they no longer apply to textarea. Should we move these styles into the components that use them?

### Final Boss

- [ ] Go through and review all docs to ensure things still make sense and work properly

PR #981, opened 2025-05-21
Revert `<wa-code-demo>` in docs

### LINKED ISSUE #680: FA Pro icons not showing in CodePen examples
Our [default callout example](https://backers.webawesome.com/docs/components/callout) uses a `circle-info` icon. However, since [`circle-info` regular is a FA Pro icon](https://fontawesome.com/icons/circle-info?f=classic&s=regular), this means when the code example is opened on any other domain (e.g. on CodePen or copied) we get a blank icon.

Docs:

Image

CodePen:

Image

Worse, there is no indication of WHY this error is happening.

----

Not sure how to best go about this. Solid icons don't look as nice in callouts, and I can't find a FA Free regular icon that fits this purpose as well. 😕

claviska:
The `<wa-code-demo>` component is marked as experimental but it's way too opinionated and messy for a stable candidate. It was an opportunistic addition that was added hastily and it's made it more difficult to fix things like #680 due to the abstracted logic.

It also introduced problems with `<wa-popup>` being clipped/sized problematically in various demos like dropdown, select, and other popup-related components.

There may be value in such a component at some point, but at the moment it's easier to revert to the old behavior and inline the examples.

@lindsaym-fa I brought back the original resizer and made it even better than the tiny handle we have now.

https://github.com/user-attachments/assets/419e6145-b68e-4336-8fd0-87f10017df07

This PR also fixes #680.

---

### TODO

- [x] The `<wa-page>` docs are borked because they're not isolated anymore. I may move these to separate windows unless @KonnorRogers has another idea for them.
- [x] Normalize a handful of camelCased filenames
- [x] Review remaining docs to ensure everything is working as expected

**Note:** There are some spacing issues in various places as a result of this change, which appear to be from the ridiculously high specificity of the native style selectors, so most of that _should_ be fixed with #943 and the addition of CSS layers., e.g.:
  Image


claviska:
I fixed the `<wa-page>` docs by breaking out the two demos into separate HTML files. This works nicely, as the user can view and inspect full-sized pages now.

PR #993, opened 2025-05-23
Simplify native styles

claviska:
Details in #943


lindsaym-fa (approved):
Approved!

We have some remaining work to do to ensure theme overrides apply uniformly to both components (working ✅) and native elements (kind of working 👀), but that'll be separate from this PR.

Issue #974, opened 2025-05-21
Refactoring Web Awesome Element


claviska:
Goals:

- [x] leaner
- [x] understandable
- [x] don't break things like form validation

claviska:
Fixed in #997

Issue #950, opened 2025-05-19
Revert setting icons in CSS


claviska:
This has caused quite a bit of headache. Let's revert it and set the appropriate icon families from the themer using the public API.

### TODO

- [x] Revert nonstandard CSS icon logic
- [x] Fix incorrectly filled Lucide, Heroicons, Iconoir (and possibly other) icons sets
- [x] Verify that using custom icons for default AND system libraries works as intended
- [x] Fix broken dropdowns in docs (dropdown page)
   - **Update:** this was caused by passthrough styles forcing `display: contents` on `<wa-button>`. They're still broke in _this_ PR but they're not broke with #993 merged in.

---

The following tasks have been moved to #979 since they go hand-in-hand with the themer rework:

- Create system libraries for all FA families
-  Add fonts back to each theme
   - Premium: family=sharp, variant=regular
   - Shoelace: variant=light
   - Mellow: family=duotone, variant-light
   - Matter: variant=regular
   - Brutalist: variant=sharp;

### Follow-ups

- [x] Related issue reported by a backer: https://github.com/shoelace-style/webawesome-alpha/issues/275

claviska:
Fixed in #997

PR #997, opened 2025-05-28
Fix icons

claviska:
See #950 for details.

@KonnorRogers I believe this PR solves the concerns we had in #974. Reopen if you feel otherwise.


KonnorRogers (dismissed):
Seems fine to me. Nothing to add here.

src/components/button-group/button-group.ts
lindsaym-fa, line 1: Looks like we can no longer set `size` or `variant` on button groups. Is that intentional?

`size` before vs. now —
Screenshot 2025-05-29 at 2 13 23 PMScreenshot 2025-05-29 at 2 13 17 PM

`variant` before vs. now —
Screenshot 2025-05-29 at 2 13 43 PMScreenshot 2025-05-29 at 2 13 37 PM

src/components/card/card.ts
lindsaym-fa, line 1: With the changes to components like button and rating, components within cards no longer inherit the card's `size`, which makes the attribute much less valuable, imo. Is this intentional? If so, should we remove `size` from `wa-card`?

`size` before vs. now —
Screenshot 2025-05-29 at 2 26 59 PMScreenshot 2025-05-29 at 2 27 04 PM

src/components/dropdown/dropdown.test.ts
lindsaym-fa, line 1: Should we restore these tests now?

src/components/page/page.ts
lindsaym-fa, line 1: Not sure the culprit is in this file, but it looks like we clobbered `wa-page`'s Layout Anatomy preview.

before vs. now —
Screenshot 2025-05-29 at 2 36 45 PMScreenshot 2025-05-29 at 2 36 56 PM

src/components/radio-group/radio-group.ts
lindsaym-fa, line 1: Radio items within a radio group no longer respond to the radio group's `size`.

`size` before vs. now —
Screenshot 2025-05-29 at 2 39 11 PMScreenshot 2025-05-29 at 2 39 16 PM

lindsaym-fa (changes requested):
We can no longer set `size` on certain parent components (namely `wa-button-group`, `wa-radio-group`, and `wa-card`) and except the child components to inherit the correct size. Not sure if that's intentional, but if it is, we should make sure to update the docs and probably also remove the attributes from the parent components.

src/components/button-group/button-group.ts
claviska: That was intentional because we're no longer using `style-observer`.

src/components/card/card.ts
claviska: Also intentional because we're not using `style-observer`. I didn't even realize `<wa-card>` had a `size` attribute. I'm still discovering so many fun things. 😕

src/components/radio-group/radio-group.ts
claviska: Same result from `style-observer`, but since there's a clear parent/child relationship here, we can trivially have radio group set the size on its children. Would you like me to make that change?

src/components/dropdown/dropdown.test.ts
claviska: Nice catch. Done.

src/components/page/page.ts
claviska: Wait what? I'm not seeing that at all in any browser. 😬 

CleanShot 2025-05-29 at 14 59 41@2x

src/components/card/card.ts
lindsaym-fa: I think it's fine to remove `size` from card. Not sure how valuable it is anyway...

src/components/button-group/button-group.ts
lindsaym-fa: Got it. No objections if we remove `size` and `variant` from `wa-button-group`.

src/components/radio-group/radio-group.ts
lindsaym-fa: In this case, I think that'd be great!

claviska: fixed in 8240e3a2bc934fd35e6cc2f899c807aafecdb71d

CleanShot 2025-05-29 at 15 10 55@2x

src/components/page/page.ts
lindsaym-fa: Erm...false alarm, I guess. 😅 I'm not seeing it any more either.

src/components/button-group/button-group.ts
claviska: Oh sorry, I read that too quickly. I can update button group to work like radio group. Stand by!

claviska: Fixed in 2b7a5a6f035dc1864e55ab1bafe0e2a364cac77b

CleanShot 2025-05-29 at 15 14 01@2x

src/components/card/card.ts
claviska: Yeah, good call. If we really want size to cascade, we can plan out a more consistent way to do it, likely through special CSS custom properties.

lindsaym-fa (approved):

PR #1001, opened 2025-05-29
Update dependencies

claviska:
It's been awhile since we've updated our dependencies. I ran `npm run check-updates` and tested the build afterwards.


KonnorRogers (approved):

claviska:
Updating is causing peer dependency conflicts and a weird 11ty error with nunjucks. This isn't critical, so I'll close this to focus on other things and try updating these in smaller batches later on.

PR #1012, opened 2025-06-02
Add `<wa-popover>`

### LINKED ISSUE #587: Popover component
https://github.com/shoelace-style/webawesome-alpha/discussions/152

We have `<wa-popup>` but it's a low-level utility, this is as high level as `<wa-dialog>`. We may also want to rename `<wa-popup>` to something that indicates that, as folks keep getting confused. Then the name `<wa-popup>` is up for grabs and we can decide whether we prefer `<wa-popup>` or `<wa-popover>` for this component.

Possibly should wait until #273 is fixed. (edit: it is now fixed)

claviska:
- Adds the `<wa-popover>` component
- Closes #587 

---

Preview link: https://webawesome-git-popover-font-awesome.vercel.app/docs/components/popover

### DIFF
--- a/packages/webawesome/docs/_includes/sidebar.njk
+++ b/packages/webawesome/docs/_includes/sidebar.njk
@@ -137,6 +137,7 @@
       </ul>
     </li>
     <li><a href="/docs/components/mutation-observer/">Mutation Observer</a></li>
+    <li><a href="/docs/components/popover/">Popover</a></li>
     <li><a href="/docs/components/popup/">Popup</a></li>
     <li><a href="/docs/components/progress-bar/">Progress Bar</a></li>
     <li><a href="/docs/components/progress-ring/">Progress Ring</a></li>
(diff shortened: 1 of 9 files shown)


lindsaym-fa (commented):
All works great! Ran into one bug, but if we'd prefer to forgo a fix for now and create a separate issue to track it, let me know.

Using VoiceOver in Safari, activating a popover trigger seems to move focus to the last activated trigger:

https://github.com/user-attachments/assets/5127707f-d3d4-432e-86b1-fded4670a244



To test:
1. Open VoiceOver
2. Move focus to a popover trigger and press Enter
3. Move focus to a different popover trigger and press Enter
4. Observe focus is commandeered by the first popover trigger

claviska:
There's logic to return focus to the anchor element, but that only runs when you press escape. Clicking or tabbing outside doesn't force focus. We'll have to circle back to this one.

PR #1015, opened 2025-06-02
Fix scroll on reload

claviska:
With this PR, reloading any page will retain your scroll position instead of jumping around randomly.

https://github.com/user-attachments/assets/9a9457ec-f485-433e-ba13-b3d71b6c5f41

https://github.com/user-attachments/assets/5edb3702-940a-4c2b-9913-3f6461b76c92


claviska:
Merging to give this a try. Any issues, let me know.

Issue #998, opened 2025-05-28
Revert to Lit's standard way of assigning component styles


claviska:
```
// Lit's recommended version
static styles: CSSResultGroup = [componentStyles, styles];

// Our version
static shadowStyle = styles;
```

claviska:
Fixed in #1016

PR #1016, opened 2025-06-03
Refactor component styles and simplify custom states

claviska:
This PR addresses #998.

Unfortunately, we can't fully revert to Lit's `static styles = [...]` if we want to continue importing CSS files as strings. Furthermore, rather than include host styles and `unsafeCSS()` in every component, it makes sense to keep that logic in `webawesome-element.ts`, simplify it, and rename it to `static css` instead of `static shadowStyles`.

I also noticed that we weren't using `addCustomState()` or `removeCustomState()` anywhere in the library. In fact, we only use `toggleCustomState()` so I refactored all the custom state logic into `this.customStates()` to reduce the footprint and scope all of the custom state logic to one object.

```ts
// Add state
el.customStates.set('open', true);

// Remove state
el.customStates.set('open', false);

// Check state
const isOpen = el.customStates.has('open');
```

These are only used internally so there's no public API being changed in this PR.


KonnorRogers:
LGTM. No issues here. Seems reasonable.

KonnorRogers (approved):

PR #1017, opened 2025-06-03
Align native styles with `<wa-slider>`

### LINKED ISSUE #627: Native sliders don't completely match `<wa-slider>`
Understanding that we may not be able to get things to match 100%, I think we make some improvements to bring `<wa-slider>` and our styles for `<input type="range">` a bit closer, e.g.:
- Ensure the label and input of each use the same heights/vertical alignment
- Let `<input type="range">` use 100% width

Image

claviska:
- Makes native styles more consistent with `<wa-slider>`
- Removes unused custom properties leftover from shared styles (only two are left because they are reused)
- Fixes #627 

Before
CleanShot 2025-06-03 at 08 09 47@2x

After
CleanShot 2025-06-03 at 08 08 57@2x

### DIFF
--- a/packages/webawesome/src/styles/native.css
+++ b/packages/webawesome/src/styles/native.css
@@ -991,43 +991,20 @@
 
   /* Slider */
   input[type='range'] {
-    --thumb-color: var(--wa-form-control-activated-color);
-    --thumb-gap: calc(var(--thumb-size) * 0.125);
-    --thumb-shadow: initial;
     --thumb-size: calc(1rem * var(--wa-form-control-value-line-height));
-
-    --track-color-active: var(--wa-color-neutral-fill-normal);
-    --track-color-inactive: var(--wa-color-neutral-fill-normal);
-    --track-active-offset: 0%;
     --track-height: calc(var(--thumb-size) * 0.25);
 
     display: flex;
     flex-direction: column;
     position: relative;
-  }
-
-  input[type='range'] {
-    --percent: 0%;
-    -webkit-appearance: none;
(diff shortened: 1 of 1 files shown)


packages/webawesome/src/styles/native.css
```diff
@@ -991,43 +991,20 @@
 
   /* Slider */
   input[type='range'] {
-    --thumb-color: var(--wa-form-control-activated-color);
-    --thumb-gap: calc(var(--thumb-size) * 0.125);
-    --thumb-shadow: initial;
```
lindsaym-fa, line 996: It seems like we'd benefit from keeping these so that folks can easily customize styles for `<wa-slider>` and `<input type="range">` simultaneously, like:
```css
wa-slider,
input[type='range'] {
  --thumb-color: tomato;
}
```

What do you think?

claviska: Do we document any custom properties on any other native controls?

I disagree with keeping them _but_ I'm like a 2/10 on it so as long as we're consistent I'm happy to do it either way. If we do keep them, we should document them in all affected native elements, otherwise nobody will know about them :)

lindsaym-fa: There are some compelling pros to removing these — namely, it could really clean up `native.css`, and contrary to my first comment, encouraging folks to write separate rules to override components vs. native elements could be a good thing for predictability since there are a number of styles you may want to override that can't be shared anyway.

So, no objections to removing them! There are a number of other native elements that could use the same treatment.

lindsaym-fa (approved):

PR #1019, opened 2025-06-03
Improve rating's default a11y

### LINKED ISSUE #429: Improve accessibility of rating
Technically, both the empties and the solids have pretty poor contrast. (Not sure what I was thinking five years ago...)

Adding this issue to track a fix for 3.0.

CleanShot 2025-01-06 at 15 54 47@2x

Reported by: https://discord.com/channels/739437527907303535/1324758276902224003/1324758276902224003

claviska:
Turns out, we _do_ have an outlined star in FA free. 😆 

This PR improves the default contrast of rating, making it more accessible by showing a darker outline for inactive stars and the same solid yellow for active ones.

An `isSelected` argument was added to `getSymbol()` to allow us (and users) to hook into it. The docs have been updated as well.

@lindsaym-fa I had to use `--wa-color-neutral-border-loud` for the outline, as the fills weren't dark enough. Let me know if there's a better token for this.

Before:

CleanShot 2025-06-03 at 08 32 32@2x

After:

CleanShot 2025-06-03 at 08 31 39@2x

Fixes #429


packages/webawesome/src/components/rating/rating.css
```diff
@@ -4,7 +4,7 @@
   --size-m: var(--wa-space-l);
   --size-l: var(--wa-space-xl);
 
-  --symbol-color: var(--wa-color-neutral-fill-normal);
+  --symbol-color: var(--wa-color-neutral-border-loud);
```
lindsaym-fa:
```suggestion
  --symbol-color: var(--wa-color-neutral-on-quiet);
```

This will be a safer bet to ensure compliant contrast.

lindsaym-fa (dismissed):
Yay! Looks great!

Issue #1000, opened 2025-05-29
Color picker buttons are light mode when color picker is rendered in dark mode


claviska:
- Go to Color Picker docs
- Enable dark mode

![Image](https://github.com/user-attachments/assets/030762ba-ee8e-4e94-b704-51c7d1ab156c)

claviska:
This was actually working as expected. I decided to add `appearance="outlined"` to improve the aesthetic. Fixed in #1021.

PR #1021, opened 2025-06-03
Use outlined appearance for buttons in `<wa-color-picker>`

claviska:
This is, perhaps, opinionated, but these buttons always appeared very jarring to me, especially in dark mode. So much that it prompted me to file an issue #1000 thinking it was a bug.

It's not. The styles are correct in light and dark modes.

### Before

CleanShot 2025-06-03 at 09 13 59@2x

CleanShot 2025-06-03 at 09 15 23@2x

### After

CleanShot 2025-06-03 at 09 16 38@2x

CleanShot 2025-06-03 at 09 16 57@2x

@lindsaym-fa this is a separate issue, but the buttons have `size="small"` but don't render consistently with the small text field. I'm hoping your size PRs will resolve this!


claviska:
> @lindsaym-fa this is a separate issue, but the buttons have size="small" but don't render consistently with the small text field. I'm hoping your size PRs will resolve this!

Actually this was caused by button groups eagerly setting the size of child buttons. Fixed in [538fda7](https://github.com/shoelace-style/webawesome/pull/1021/commits/538fda7563a8bfcccf68234af80a8a60923d4a8b).

lindsaym-fa (dismissed):
Agreed! Looks much better.

PR #1023, opened 2025-06-03
Simplify color variants docs

### LINKED ISSUE #623: Color Variants doc mistakenly lists `<wa-icon>` and `<wa-icon-button>`
`<wa-icon>` and `<wa-icon-button>` don't respect color variants set via `variant=""` or `.wa-[variant]`.

Image

The [Color Variants doc](https://backers.webawesome.com/docs/utilities/color/) claims that they should:

Image

I'd recommend we simply remove this from the docs for now and add color variants to icons separately.

claviska:
- Fixes #623 
- Removes broken logic
- Removes the [unlisted] cheatsheet that nobody is using


lindsaym-fa (approved):
Lovely. ✨

Issue #1024, opened 2025-06-03
Fix `<wa-radio>` hint styles OR remove the property/styles


claviska:
From `radio.md` (not radio group):

![Image](https://github.com/user-attachments/assets/484b12cc-0bde-45a0-bfec-5cdf9a13f352)

I wasn't aware we added `hint` to individual radios. @lindsaym-fa is this something we want to support?

**If so,** we need to fix the styles for this example (it doesn't appear to be used elsewhere directly on `<wa-radio>`).

**If not,** moving the hint to the `<wa-radio-group>` and off the individual radios will solve it, but we'll also need to remove the property and any associated styles from `<wa-radio>`

To be clear, this is about hints on individual `<wa-radio>` items, NOT hints on `<wa-radio-group>` which we will continue to support.

lindsaym-fa:
I had questioned this from the get-go. imo, we should **not** support `hint` on individual radios. Seems like a recipe for a mess.

claviska:
Fixed in #1026

takesson:
I really miss the ability to explain each option. Can you add an example to the documentation how to achieve something similar in the primary slot of wa-radio?

claviska:
Radio labels accept non-interactive content, so this works fine. If you're keeping the primary label (which uses aria-description to link to the radio group) it will compete visually, but it can easily be omitted.

```html
<wa-radio-group label="Select an option" name="a" value="1">
  <wa-radio value="1">
    Option 1
    <br>
    <small style="color: var(--wa-color-text-quiet);">Hint goes here</small>
  </wa-radio>
    
  </wa-radio>
  <wa-radio value="2">
    Option 2
    <br>
    <small style="color: var(--wa-color-text-quiet);">Hint goes here</small>
  </wa-radio>
    
  </wa-radio>
  <wa-radio value="3">
    Option 3
    <br>
    <small style="color: var(--wa-color-text-quiet);">Hint goes here</small>
  </wa-radio>
</wa-radio-group>

```

<img width="196" alt="Image" src="https://github.com/user-attachments/assets/cf6cf555-6642-464f-8133-bf30befb2e8c" />

Example: https://codepen.io/team/webawesome/pen/Wbobdyr?editors=1000

PR #1026, opened 2025-06-03
Remove hint from `<wa-radio>`

claviska:
As discussed in https://github.com/shoelace-style/webawesome/issues/1024#issuecomment-2936718554


lindsaym-fa (approved):
Perfection.

PR #1029, opened 2025-06-04
Zoomable frame

### LINKED ISSUE #986: Refactor and graduate `<wa-viewport-demo>` to stable
This is a great idea, but it needs some love. I have a plan to turn this into another one of our 14 free stretch goal components without too much extra effort.

This will be the third (of 14) free stretch goal components added.

### Nice to haves

- [ ] Color scheme selector per #838

claviska:
- Removes `<wa-viewport-demo>` and related logic per #986
- Adds `<wa-zoomable-frame>` which is similar, but more like an iframe that can be zoomed (which is harder than it may seem to get right in all browsers)

This PR closes #986

### TODO

- [x] Check pro for pattern and `<wa-page>` demos that previously used `<wa-viewport-demo>`
- [x] Re-implement the light/dark toggle that syncs and updates code example color schemes

### After Merging

- [ ] Merge related PR in pro: https://github.com/shoelace-style/webawesome-pro/pull/6


claviska:
@KonnorRogers This is blocked due to not having a way to host example assets from the pro repo.

We have some patterns, e.g. the shopping cart drawer example, that needs to use an iframe that points to an HTML file, but those files can't be committed to the free repo.

lindsaym-fa (approved):
No notes. ✨

claviska:
@kelseythejackson both this and https://github.com/shoelace-style/webawesome-pro/pull/6 have been merged. At this point, there should be no more instances of `<wa-viewport-demo>` anywhere.

I only updated one pattern (shopping cart drawer), so if you find any others that need it, feel free to use this PR as a reference:

https://github.com/shoelace-style/webawesome-pro/pull/6/files

TL;DR – use `<wa-zoomable-frame src="/assets-pro/path/to/file.html">` and create the file in the `assets-pro` directory (newly added today). Note that each demo needs to have the scripts/styles to load WA:

https://github.com/shoelace-style/webawesome-pro/blob/next/docs/assets-pro/patterns/shopping-cart-drawer.html

I'm also happy to help you get anything remaining translated over!

Issue #246, opened 2024-11-22
Document declarative conventions and ensure alignment in existing components


LeaVerou:
There are several components that support declarative commands, which we currently do ad hoc with `data-` attributes:
- `data-toggle-nav` on `<wa-page>`
- [`data-dialog="close"`](https://webawesome-git-next-font-awesome.vercel.app/docs/components/dialog#dismissing-dialogs) for `<wa-dialog>`

While we can't yet use the [invoker API](https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API), it may be a good idea to use invoker-compatible **syntax** like `command="--toggle-nav"` so that we could seamlessly switch to it later. It costs us nothing and makes WA look forward-facing to those in the know (who are few, but tend to be influential). That said, I hate it that they went with the `--` convention because it looks ugly AF. But that ship has sailed, and this syntax is where we're heading towards, so it could be argued that we may as well be the first to use it…

An alternative path could be to use our own command attribute, e.g. `wa-command="toggle-nav"` or `wa-action="toggle-nav"`.

E.g. (if we go with `data-wa-`, which is a separate issue)

```html
<button data-wa-action="toggle-nav">
<button data-wa-action="close-dialog">
```

WRT opening dialogs, idea by @claviska:
```html
<button data-wa-action="open-dialog id">

<!-- OR, to use a selector -->
<button data-wa-action="open-dialog" data-wa-action-target="& + dialog">
```

Alternatively:

```html
<button data-wa-action="open-dialog #id">

<!-- OR, to use a selector -->
<button data-wa-action="open-dialog & + dialog">
```

claviska:
See the thread here regarding that API. https://bsky.app/profile/cory.laviska.com/post/3lb5oobabs22k

Specifically:

> Is there a shortcut? I’m currently using data-dialog=“open id” and it would be more ergonomic to continue using one attribute.
>
> command=“open:id”
>
> or similar would be great.

I like a one-attribute approach over what we got, but here we are…

I do have a question about invokers that I'm still uncertain of. If we start using `command` and `commandfor` in our code, what happens when browsers finally support it? Will it fail gracefully, allowing custom elements to take over? Or will it throw a warning/error? 🤔

LeaVerou:
> See the thread here regarding that API. https://bsky.app/profile/cory.laviska.com/post/3lb5oobabs22k
> 
> Specifically:
> 
[...]

PR #1032, opened 2025-06-04
Add data- invokers to dialog and drawer

claviska:
Per https://github.com/shoelace-style/webawesome/issues/246#issuecomment-2605605920, this expands support for data attribute invokers.

We already have:

- `<button data-dialog="close">`
- `<button data-drawer="close">`

This adds:

- `<button data-dialog="open <id>">`
- `<button data-drawer="open <id>">`

Where `<id>` is the ID of a dialog/drawer in the same document that will be opened when the button is clicked. This works on any type of button (native shown for brevity).

I've updated the docs for dialog and drawer to show an example.

### DIFF
--- a/packages/webawesome/src/components/dialog/dialog.ts
+++ b/packages/webawesome/src/components/dialog/dialog.ts
@@ -6,6 +6,7 @@ import { WaAfterShowEvent } from '../../events/after-show.js';
 import { WaHideEvent } from '../../events/hide.js';
 import { WaShowEvent } from '../../events/show.js';
 import { animateWithClass } from '../../internal/animate.js';
+import { parseSpaceDelimitedTokens } from '../../internal/parse.js';
 import { lockBodyScrolling, unlockBodyScrolling } from '../../internal/scroll.js';
 import { HasSlotController } from '../../internal/slot.js';
 import { watch } from '../../internal/watch.js';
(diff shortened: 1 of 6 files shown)


packages/webawesome/src/components/dialog/dialog.ts
```diff
@@ -264,6 +265,28 @@ export default class WaDialog extends WebAwesomeElement {
...
+
+      if (dialog?.localName === 'wa-dialog') {
+        dialog.open = true;
+      } else {
+        console.warn(`A dialog with an ID of "${id}" could not be found in this document.`);
+      }
+    }
+  }
+});
+
```
KonnorRogers, line 289: Because we never know how many times this may be added to a page, I'd much rather see it extracted into a function.

```
document.addEventListener("click", handleDialogClick)
```

to get de-duplication and not risk any weird behavior with things like Turbo.

packages/webawesome/src/components/drawer/drawer.ts
```diff
@@ -280,6 +281,28 @@ export default class WaDrawer extends WebAwesomeElement {
   }
 }
 
+//
+// Watch for data-drawer="open *" clicks
+//
+document.addEventListener('click', (event: MouseEvent) => {
```
KonnorRogers, line 287: Same thing here with extract function.

packages/webawesome/src/internal/parse.ts
```diff
@@ -0,0 +1,7 @@
+/** Parses a space-delimited set of tokens and returns an array with all whitespace removed. */
+export function parseSpaceDelimitedTokens(input: string): string[] {
+  return input
+    .split(' ')
+    .map(token => token.trim())
+    .filter(token => token !== '');
+}
```
KonnorRogers, line 7:
```suggestion
  return input
     .trim()
     .split(/(\s|\r|\n|\f)+/)
```

^ this should actually make it so you dont need to map or filter. Not a big deal, but small optimization if you care 🤷

KonnorRogers (approved):
I get where we're going with this....but I kinda hate it in its current form.

Instead of `data-dialog="open <id>"` Why can't we do:

```
data-command="open"
data-command-for="<id>"
```

or even:

```
data-command="open <id>"
```

we could even scope like:

```
data-wa-command=""
```

I feel like by having the dialog / drawer specific attributes kinda takes away the power?

PR #1034, opened 2025-06-05
Slider rework

### LINKED ISSUE #77: Refactor `<wa-slider>` to use a faux slider and support multiple thumbs
- [x] Rework `<wa-slider>` to use a faux slider, which simplifies the styling API and the styles themselves
- [x] Add support for dual thumbs
- [x] Add support for vertical orientation

claviska:
I've reworked `<wa-slider>` completely. It now uses a faux slider instead of `<input type="range">`

  - Added support for range sliders with dual thumbs using the `range` attribute
  - Added vertical orientation support with `orientation="vertical"`
  - Added visual markers at each step with `with-markers`
  - Added contextual reference labels with `with-references` and the `reference` slot
  - Added tooltips showing current values with `with-tooltip`
  - Added customizable indicator offset with `indicator-offset` attribute
  - Added value formatting support with the `valueFormatter` property
  - Improved the styling API to be consistent and more powerful (no more browser-specific selectors and pseudo elements to style)
  - Updated to use consistent `with-*` attribute naming pattern

---

@lindsaym-fa Please have a look at the design and overall API. I have a bit of tunnel vision right now, so let me know if I missed anything.

@KonnorRogers Please look at the form-associated logic and validation. I had to add a new validator for this, so a second pair of eyes will be helpful. I've tested form submission for both single and dual sliders and they submit as expected.

Closes #77


claviska:
@KonnorRogers most of the failing tests are from the form control base tests. I'm guessing my implementation missed something. Any ideas?

claviska:
Thanks for the green check, @KonnorRogers! @lindsaym-fa this one is all ready for a design review! 😄

packages/webawesome/src/components/slider/slider.css
```diff
@@ -1,214 +1,227 @@
...
-  &:dir(rtl) {
-    --dir: left;
+/* Add extra space between slider and label, when present */
+#label:has(*:not(:empty)) ~ #slider {
+  &.horizontal {
+    margin-block-start: 0.5em;
+  }
+  &.vertical {
+    margin-block-end: 0.5em;
   }
```
lindsaym-fa, line 32: I added `#label:has(*:not(:empty)) ~` to these styles so that we only add extra margin between the label and the slider when the label slot actually has content. This should help ensure expected alignment and spacing when folks need to use different labeling techniques, like `aria-labelledby`.

lindsaym-fa (approved):
This is _awesome_. 🤩

PR #1035, opened 2025-06-05
Add hover queries

### LINKED ISSUE #990: Add hover media query to all hover styles
It's kind of annoying, but we should add the following media query around all `:hover` declarations in all components and theme overrides to prevent touch devices from leaving trails.

Reported here in Shoelace: https://github.com/shoelace-style/shoelace/issues/2444

Thankfully, we can nest it like so:

```css
a {
  background: lightblue;
  color: black;

  @media (hover: hover) {
    &:hover {
      background: darkblue;
      color: white;
    }
  }
}
```

claviska:
Adds `@media (hover: hover) { ... }` around component hover styles to prevent [sticky hover states](https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/).

It's so nice that we can nest these! 😅 

Fixes #990

### DIFF
--- a/packages/webawesome/docs/assets/styles/code-examples.css
+++ b/packages/webawesome/docs/assets/styles/code-examples.css
@@ -116,10 +116,12 @@
     padding: 0.5rem;
     cursor: pointer;
 
-    &:hover {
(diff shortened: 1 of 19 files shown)


lindsaym-fa (approved):
Pushed a small commit to add `@media (hover: hover)` to the hover styles for `<wa-radio appearance="button">` since that was the culprit in the original issue.

All looks good!

PR #1062, opened 2025-06-17
Fix tag sizes

### LINKED ISSUE #1059: Tag sizes no longer work
Despite importing the size styles, applying a size to `<wa-tag>` doesn't seem to do anything.

Image

lindsaym-fa:
Closes #1059

We previously encountered a similar issue in #993 where the `appearance` attribute wasn't working for `<wa-button>` until [this fix](https://github.com/shoelace-style/webawesome/pull/993/commits/e99fe2adff1e541fea24d904fc76d57b11bb09ac) was implemented.

Because `<wa-tag>`'s `:host` styles specify `font-size: inherit`, and these styles are unlayered, `font-size: inherit` takes precedence over the layered `font-size` styles in `size.css`. We fix this by wrapping `<wa-tag>`'s `:host` styles in its own layer and changing the order of internal styles so that the `wa-component` layer is lower in the cascade order than any of the styles in the `wa-utilities` layer.

### DIFF
--- a/packages/webawesome/src/components/tag/tag.css
+++ b/packages/webawesome/src/components/tag/tag.css
@@ -1,21 +1,23 @@
-:host {
-  display: inline-flex;
-  gap: 0.5em;
-  border-radius: var(--wa-border-radius-m);
-  align-items: center;
-  background-color: var(--background-color, var(--wa-color-fill-quiet));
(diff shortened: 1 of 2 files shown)


claviska (approved):
Looks great!

PR #1063, opened 2025-06-20
Rework setting initial values in `<wa-select>`

KonnorRogers:
### TODO

- [x] Update select
- [x] Update docs


claviska:
@KonnorRogers Is this something I can take over or should I wait for you?

KonnorRogers:
@claviska you're more than welcome to take it over. I think it covers _most_ of the edge cases. Documentation does need to be updated, tests fixed, and remove the top example I was using for testing. So totally up to you how you want to handle it. You can always merge it and we can fix it later.

PR #1079, opened 2025-06-26
Rename caret => with-caret in `<wa-button>`

claviska:
To align with the `with-*` and `without-*` convention for optional content, I've updated `<wa-button caret>` to `<wa-button with-caret>` including all example usages.

There was one instance of caret being used in pro that I fixed in a separate PR.


lindsaym-fa (approved):
Looks great! ✨

PR #1048, opened 2025-06-10
Themer rework

claviska:
<br>

**June 10, 2025** – The themer rework has commenced. Just about everything that was there has been pulled out in f6a70d1110cd51b6eee3499720f3af89028f7587. Time to rebuild. Sister PR: https://github.com/shoelace-style/webawesome-pro/pull/10

---

### TODOs

Some of this is tied so heavily together, it's going to be easier to reimplement them instead of untying the knots one by one. This section is a living list of things that have been temporarily removed or still need to be done to support this effort.

- [x] Reimplement color scheme selector
- [x] Reimplement `<wa-page>` demo
- [x] Reimplement theme selector
- [x] https://github.com/shoelace-style/webawesome/issues/1051
- [x] Move PRO themes over to the pro repo
- [x] Add component headers back
- [x] Recreate components index page
- [x] Recreate palettes page
- [x] Recreate themes page
- [x] Reimplement the themer
  - [Demo of old themer](https://webawesome-git-lm-consolidate-theme-docs-font-awesome.vercel.app/docs/themes/edit/)
  - [x] Starting themes
  - [x] Color palettes
  - [x] Fonts
  - [x] Icons
        - Premium: family=sharp, variant=regular
        - Shoelace: variant=light
        - Mellow: family=duotone, variant-light
        - Matter: variant=regular
        - Brutalist: variant=sharp;
  - [x] Elements
  - [x] Improve form control styles

<details>
<summary>Archived</summary>

### Prereqs

- [x] #950
- [x] #972 

### Things of interest / cleanup

- [x] `docs/assets/scripts/sidebar-tweaks.js`
- [x] `docs/assets/scripts/my.js`
- [x] `docs/assets/scripts/tweak/*` (wtf)
- [x] `docs/assets/scripts/util/*`
- [x] `docs/_data` folder because it's not intuitive (set data in `.eleventy.js` or in frontmatter instead)

</details>

PR #1080, opened 2025-06-26
Remove custom properties that are easily styled with CSS parts

### LINKED ISSUE #1046: Simplify superfluous custom properties in components
In #1027, many custom properties were cleaned up from native styles. This issues tracks cleaning them up in component styles.

- Remove any unnecessary custom properties from component styles
- Ensure custom properties that remain are correctly documented
- Ensure custom properties that were eliminated are removed from the docs

lindsaym-fa:
This work removes a majority of custom properties we had previously added to components in favor of prioritizing styling via `::part()` selectors.

To this end, this PR also eliminates `styles/utilities/appearance.css` in favor of including the relevant appearances within each component's own styles. This makes it much easier to track and understand where certain styles are being set.


lindsaym-fa:
The only necessary remaining work here is to fix any theme overrides that may have been targeting the removed custom properties. Because https://github.com/shoelace-style/webawesome/pull/1048 implements a number of changes to those theme stylesheets, I'll wait until that PR merges before I pursue those fixes.

claviska (approved):

claviska (approved):

PR #1104, opened 2025-07-01
Adjust package.json styles exports

ajmchambers:
Error when attempting to import styles:

```
import "@awesome.me/webawesome/dist/styles/themes/default.css"
import "@awesome.me/webawesome/dist/styles/webawesome.css"
```

Adjusts package.json exports from `./dist/themes` to `./dist/styles`.


claviska:
@KonnorRogers This seems fine to me. I believe we do want to expose everything in `/dist/styles/*`, which is where themes live now. Any concerns from you about this change?

KonnorRogers:
nope. Send it 🚀 

Seems like it was just missed in the conversion.

KonnorRogers (approved):

claviska:
Thanks, @ajmchambers!

PR #1111, opened 2025-07-01
Add opt-in class to highlight table rows on hover

lindsaym-fa:
Moves existing `:hover` styles for table rows into an opt-in class, `.wa-hover-rows`

### DIFF
--- a/packages/webawesome/docs/_layouts/component.njk
+++ b/packages/webawesome/docs/_layouts/component.njk
@@ -26,7 +26,7 @@
     <p>Learn more about <a href="/docs/usage/#slots">using slots</a>.</p>
 
     <wa-scroller>
-      <table class="component-table">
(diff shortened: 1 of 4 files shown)


lindsaym-fa:
@claviska, any thoughts on naming? I have `.wa-hover-rows` currently to keep it concise, but I worry it may not be descriptive enough. Others I considered: `.wa-hoverable-rows`, `.wa-highlight-row-on-hover`

`.wa-hoverable-rows` actually feels better to me now that I've taken a second look, but I'd love to hear your ideas.

claviska:
I like `.wa-hover-rows`, but if you're feeling strongly about `.wa-hoverable-rows` that works for me too!

lindsaym-fa:
We'll keep it as-is. Thanks!

claviska (approved):

PR #1125, opened 2025-07-02
Themes page fix

claviska:
- This fixes the themes page where themes are previewed by ensuring the correct palette class is applied to the preview. The copy/paste code has also been updated to include it
- Also fixes the copy button's hover state
- Also removes the code block links due to poor UX, as the IDs are incrementally generated and fragile (each example has a heading that's a lot less likely to change; please use that instead)

CleanShot 2025-07-02 at 15 43 50@2x


lindsaym-fa (approved):
Made some changes so that we're updating with the intended brand color, too.

Approved! ✨

PR #1128, opened 2025-07-02
Add GitHub repository links

lindsaym-fa:
Reorganizes our docs header, sidebar, and homepage to make way for links to this repository.

---

Moves search button to the sidebar:
Screenshot 2025-07-02 at 5 39 21 PM

---

Which makes room for GitHub and Star links in the header:
Screenshot 2025-07-02 at 5 44 12 PM

---

Removes references to backers from the beta callout (I'll be milking this "bigger and beta" joke for as long as I can):
Screenshot 2025-07-02 at 5 40 07 PM

---

Updates the links lower on the homepage to include GitHub and Discord with a separate section for socials:
Screenshot 2025-07-02 at 5 39 37 PM

### DIFF
--- a/packages/webawesome/docs/_includes/base.njk
+++ b/packages/webawesome/docs/_includes/base.njk
@@ -92,7 +92,7 @@
       {% if hasSidebar %}
         {# Mobile selectors #}
         <div class="wa-mobile-only" slot="navigation-header">
-          <div class="wa-cluster wa-gap-xs">
+          <div class="wa-cluster wa-gap-xs" style="flex-wrap: nowrap;">
             {% include "theme-selector.njk" %}
             {% include "color-scheme-selector.njk" %}
           </div>
--- a/packages/webawesome/docs/_includes/base.njk
+++ b/packages/webawesome/docs/_includes/base.njk
@@ -61,19 +61,27 @@
           <wa-badge variant="brand" appearance="filled" class="wa-desktop-only">Beta</wa-badge>
         </div>
 
-        <div id="docs-toolbar" class="wa-cluster wa-gap-xs">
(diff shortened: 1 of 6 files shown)


lindsaym-fa:
@claviska, the Discord link shows that the invite is coming from me. Would you prefer we update this with an invite link from you instead? 😄 
Screenshot 2025-07-02 at 5 45 24 PM

claviska:
> Would you prefer we update this with an invite link from you instead? 😄

It doesn't matter to me as long as you're comfortable with it.

claviska:
Hmm, there's a lot of real estate being wasted at the top of the mobile sidebar:

CleanShot 2025-07-03 at 09 02 21@2x

I wanted to see what it looked like if the selects were on the same row. By setting `flex-wrap: nowrap` on the `<div class="cluster">`, it looks like this:

CleanShot 2025-07-03 at 09 00 19@2x

There's less wasted space, but I wonder what it would look like to have the search up there too. 🤔 

CleanShot 2025-07-03 at 09 09 40@2x

Here's what I did in the dev console to get here:

- hid the `display-input` part of both selects
- applied `width: fit-content` to the `listbox` part of each select
- remove light DOM margin that caused extra space
- moved mobile search into the same cluster

It's more compact, but I'm not sure I love it since you can't see the selected theme's label now. And the search should probably stretch closer to the close button.

I think I like the second image (with search below the header), but maybe you have some better ideas on how to clean this up.

lindsaym-fa:
Agreed that your second screenshot is a good balance between space usage and usability! I pushed some changes to replicate that.

claviska (approved):

PR #1145, opened 2025-07-07
(docs/tokens/color): Fix links pointing to / instead of the current page

randomguy-2650:
This fixes links referring to headings on the page, but are referring to headings on the root of the page (/) instead of the current page.


lindsaym-fa (approved):

lindsaym-fa:
Thanks, @randomguy-2650!

PR #1148, opened 2025-07-07
fix the build on windows

dotjoe:
fix the build on windows

  - make-react importPath variable
  - esbuild entryPoints globby usage requires forward slashes


KonnorRogers (approved):
I'm kinda surprised this is the only change needed to run on windows....anyways, thanks!

dotjoe:
I think the test script also needs to use something like cross-env to set that CSR_ONLY env variable

KonnorRogers:
Yea.... that was gonna be my next question. None of us use Windows...and i imagine a number of other things are broken for windows users not using something like Git bash / WSL.

PR #1153, opened 2025-07-08
Revise native styles documentation

lindsaym-fa:
Dramatically updates our documentation of native styles by:
- Adding and improving examples to comprehensively demo all native styles
- Rewriting section descriptions to remove broken links, list relevant HTML tags, and refer to associated design tokens and WA components
- Implementing minor improvements to native styles, including:
  - Adding `<hr>` as an element that receives the default content spacing (region: Content Flow)
  - Moving list margin from `<ol>`/`<ul>` to `<li>` to be more compatible with our layout utilities, which zero out the margin of children
  - Correct `<progress>` height to match `<wa-progress-bar>`


claviska:
This looks great!

Not a blocker, but lots of demos have bottom margins like this that make them awkward. Should we address it in `docs.css`?

CleanShot 2025-07-14 at 16 08 43@2x

claviska (approved):

lindsaym-fa:
@claviska The extra whitespace was actually fixed in #1156, which hadn't yet been merged into this branch. You'll see much better spacing in the demo examples in `next` now. 😄

PR #1157, opened 2025-07-09
Replace old `--wa-flow-spacing` with `--wa-content-spacing`

lindsaym-fa:
Simply replaces our old `--wa-flow-spacing` property, which was only in use in the docs, with the `--wa-content-spacing` token from the applied theme.

### DIFF
--- a/packages/webawesome/docs/assets/styles/code-examples.css
+++ b/packages/webawesome/docs/assets/styles/code-examples.css
@@ -3,7 +3,7 @@
   border: var(--wa-border-style) var(--wa-panel-border-width) var(--wa-color-neutral-border-quiet);
   border-radius: var(--wa-border-radius-l);
   color: var(--wa-color-text-normal);
-  margin-block-end: var(--wa-flow-spacing);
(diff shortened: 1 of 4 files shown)


KonnorRogers (dismissed):
Looks good to me, do we have a changelog entry for the property change? I couldn't find one.

lindsaym-fa:
@KonnorRogers `--wa-flow-spacing` was only something we used internally. Looks like I added `--wa-content-spacing` was added to themes and used in native styles as a part of our native styles refactoring (#1043) but I missed adding the addition to the changelog. Would it make sense to add it to the `beta.1` changelog retroactively?

KonnorRogers:
> @KonnorRogers `--wa-flow-spacing` was only something we used internally. Looks like I added `--wa-content-spacing` was added to themes and used in native styles as a part of our native styles refactoring (#1043) but I missed adding the addition to the changelog. Would it make sense to add it to the `beta.1` changelog retroactively?

sounds reasonable to me

lindsaym-fa:
Changelog added!

KonnorRogers (approved):

PR #1169, opened 2025-07-12
(l10n: fr, it): Use smart quotes for localisation keys

randomguy-2650:
#### Description

This PR updates localisation keys to use smart quotes. These changes apply to French and Italian.

I also fixed the Italian language name to be **Italiano** instead of Italian, it should now be more consistent with other languages.

The reason I created this PR is to match the browser defaults for other stuff like the `<q>` element, which encloses text in smart quotes (“”) rather than straight quotes ("").


randomguy-2650:
After reading some other stuff about `aria-label`s, smart quotes are generally not recommended for this case, so I’ll close this PR.

PR #1171, opened 2025-07-13
(l10n: it): Fix language name in `it.ts`

randomguy-2650:
#### Description

This fixes the language name used in `it.ts`, a localisation file.


claviska (approved):
Nice catch! Thanks

PR #1173, opened 2025-07-14
fix badge pulsing

### LINKED ISSUE #1188: Badge - pulse animation not visible due to missing variable declaration
Documentation for badge has this issue, at the "Drawing Attention" section. The issue happens because of the missing `--background-color` variable. 
https://webawesome.com/docs/components/badge/#drawing-attention
I can submit a PR, but the new theme and token style is still new to me. 

Image

Previously, Shoelace v2 defined it like this. 

`  /* Pulse modifier */
  .badge--pulse {
    animation: pulse 1.5s infinite;
  }

  .badge--pulse.badge--primary {
    --pulse-color: var(--sl-color-primary-600);
  }

  .badge--pulse.badge--success {
    --pulse-color: var(--sl-color-success-600);
  }

  .badge--pulse.badge--neutral {
    --pulse-color: var(--sl-color-neutral-600);
  }

  .badge--pulse.badge--warning {
    --pulse-color: var(--sl-color-warning-600);
  }

  .badge--pulse.badge--danger {
    --pulse-color: var(--sl-color-danger-600);
  }`

KonnorRogers:
Fix badge pulsing reported here:

https://discord.com/channels/739437527907303535/1394349360074194975

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -8,6 +8,13 @@ Web Awesome follows [Semantic Versioning](https://semver.org/). Breaking changes
 
 Components with the <wa-badge variant="warning">Experimental</wa-badge> badge should not be used in production. They are made available as release candidates for development and testing purposes. As such, changes to experimental components will not be subject to semantic versioning.
 
+## Next
+
+### Bug Fixes and Improvements {data-no-outline}
+
+- Fixed a bug in `<wa-badge>` where `appearance="pulse"` was not working as expected [pr:1173]
+- Fixed a missing TypeScript type for `<wa-badge>` for its `attention` property missing `bounce` value. [pr:1173]
+
 ## 3.0.0-beta.2
 
 ### New Features {data-no-outline}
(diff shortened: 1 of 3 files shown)


claviska (commented):
needs changelog

lindsaym-fa (dismissed):
@KonnorRogers Breaking the pulse effect was totally my bad. Thanks for giving attention to this.

Because we have a variety of different badge appearances now, the pulse color looks best by matching the badge's background color for some appearances and matching the badge's border color for others. I pushed a change to leverage the `--pulse-color` property we already have internally for the pulsating effect and add it to the docs so that folks can customize as necessary.

If that doesn't seem like the right direction to you, lmk!

KonnorRogers:
> @KonnorRogers Breaking the pulse effect was totally my bad. Thanks for giving attention to this.
> 
> Because we have a variety of different badge appearances now, the pulse color looks best by matching the badge's background color for some appearances and matching the badge's border color for others. I pushed a change to leverage the `--pulse-color` property we already have internally for the pulsating effect and add it to the docs so that folks can customize as necessary.
> 
> If that doesn't seem like the right direction to you, lmk!

I had a funny feeling you had different intentions for this which is why I tagged you for review because it did seem out of place :D Thanks for handling that and that seems like a totally reasonable path to me!

lindsaym-fa (dismissed):

claviska (dismissed):

PR #1183, opened 2025-07-14
build speed improvements and add optional incremental builds

KonnorRogers:
- This PR drops full builds ~1second from 6.5seconds down to 5.5seconds (2.5seconds 11ty builds down to 1.5seconds) by re-using the same `node-html-parser` instance.
- In addition, in devMode it lazily transforms the second instance of nunjucks which cuts ~200ms off of full builds.
- introduces incremental loading as default. to disable incremental, use `npm run start -- --no-incremental` out incremental loading with Eleventy. Incremental builds of a markdown file seem to only take ~0.20seconds to take effect, so would eventually love to see them become the "standard" as they are much faster.

- [x] - need to fix logging from 11ty. Logging needs to be overwritten, while its not a public API, we could probably just stub `console.*` for the eleventy calls? or overwrite 11ty's logger.

Screenshot 2025-07-14 at 6 18 58 PM

Screenshot 2025-07-14 at 6 50 52 PM


packages/webawesome/scripts/docs.js
```diff
@@ -1,14 +1,98 @@
...
+   */
+  if (process.env.SKIP_ELEVENTY === 'true') {
+    return;
+  }
+
+  spinner?.start('Writing the docs');
+
+  if (isIncremental) {
+  } else {
+  }
```
claviska: What does this block do? 😆

claviska (dismissed):
Nice work!

How do we test incremental, e.g. does this work?

```sh
npm start --incremental
```

packages/webawesome/scripts/docs.js
KonnorRogers: whole lotta nothing. deleting it.

KonnorRogers:
> Nice work!
> 
> How do we test incremental, e.g. does this work?
> 
> ```shell
> npm start --incremental
> ```

```shell
npm start -- --incremental
```

claviska:
> npm start -- --incremental

Any reason we don't make this the default so the team can test it out this week? 🤔

KonnorRogers:
> > npm start -- --incremental
> 
> Any reason we don't make this the default so the team can test it out this week? 🤔

🤷 i dont see why not. then to disable it `-- --no-incremental` then?

claviska:
Works for me!

claviska (dismissed):

PR #1190, opened 2025-07-16
Improve accessibility of `<wa-animated-image>`

### LINKED ISSUE #1177: `<wa-animated-image>` is not operable by keyboard or assistive technology
### Describe the bug
[Animated Image](https://shoelace.style/components/animated-image) has no controls in the focus order so the animation can't be played/paused using a keyboard. Because it has no controls with a role or name, they can't be played/paused using assistive technologies (AT) like a screen reader or voice control. 

This is an accessibility bug. This web component is a great way to make animated images meet WCAG success criterion [2.2.2 Pause, Stop, Hide](https://www.w3.org/WAI/WCAG22/Understanding/pause-stop-hide.html) but as implemented it fails other criteria, including [2.1.1: Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html) and [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html).

### To Reproduce

1. Go to [Shoelace sl-animated-image example](https://cdpn.io/pen/debug/ByoNzGM/0ce04e30d691f1b81d9c1b7e57c58fc0)
2. Use Tab key to focus "button before" button
3. Press Tab again to focus the next focusable element
4. See focus is on "button after," `<sl-animated-image>` has been skipped

### Demo

[Codepen example](https://codepen.io/ccwilcox/pen/ByoNzGM/0ce04e30d691f1b81d9c1b7e57c58fc0)

### Screenshots
Video of reproduced behavior. Description: Pressing Tab focuses "button before," pressing Tab again focuses "button after," pressing Shift-Tab focuses "button before" again (keypresses displayed on-screen).
  
https://github.com/user-attachments/assets/69a57082-58a2-4b2a-8460-9d670c51da06

### Browser / OS
 The problem is not browser or OS-specific.

### Additional information
- [Slotting in <button> elements](https://codepen.io/ccwilcox/pen/xbwGOXR/bab4faf9e245d36090f7c30d985addbb) for the play/pause buttons doesn't help because the click listeners are on the  elements.
- Adding a click listener to `[part="control-box"]` should allow slotted-in <button> elements to work but keyboard and AT operability should instead be provided by default. A single button should suffice (e.g. make `part="control-box"` a <button> instead of a <div>), the icon and name of the button should change based on the element's play state ("play" and "pause" icons and names should still overridable using slots).
- The button(s) should always be in the focus order; when focused, the button should be visible, even if the animation is playing.
- BTW, because the component uses the write-to-`<canvas>` technique, it works not only with animated GIF and WebP but with _any_ animated image format, like animated PNG or AVIF.

claviska:
Fixes #1177


claviska:
@lindsaym-fa I opted to put the focus ring around the control box instead of the image.

lindsaym-fa (dismissed):

PR #1210, opened 2025-07-21
Add `icon-position` to `<wa-details>`

lindsaym-fa:
Per conversation in #1099

Adds `icon-position="start | end"` to `<wa-details>`

### DIFF
--- a/packages/webawesome/docs/docs/components/details.md
+++ b/packages/webawesome/docs/docs/components/details.md
@@ -48,6 +48,23 @@ Use the `expand-icon` and `collapse-icon` slots to change the expand and collaps
 </style>
 ```
 
+### Icon Position
+
+The default position for the expand and collapse icons is at the end of the summary. Set the `icon-position` attribute to `start` to place the icon at the start of the summary.
+
+```html {.example}
+<div class="wa-stack">
+  <wa-details summary="Start" icon-position="start">
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna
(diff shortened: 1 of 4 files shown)


claviska (changes requested):
This is an exemplary PR!

The only thing I noticed is that, in RTL, the icons should be mirrored.

CleanShot 2025-07-22 at 09 42 19@2x

One thing to look out for — we need to make sure it continues to work properly with custom expand/collapse icons, so flipping with CSS probably isn't the way to go here.

I'm happy to help if you'd like me to take a look!

lindsaym-fa:
@claviska I'll need a hand with this. 

Setting `dir="rtl"` on the root element or `<wa-details>` itself renders the `chevron-left` icon as intended —

Here's `dir="rtl"` on `<wa-details>` (example in the docs):
Screenshot 2025-07-22 at 11 43 40 AM

Here's `dir="rtl"` on `<html>`:
Screenshot 2025-07-22 at 11 44 18 AM

But setting `dir="rtl"` on any other ancestor or parent doesn't have the same effect. Here, I've set `dir="rtl"` on the `.code-example-preview` wrapper:
Screenshot 2025-07-22 at 11 44 53 AM

(This is also the case in beta.3.)

Any ideas? 🤔

claviska:
Oh no...you're right! Due to platform limitations [that I've complained about before](https://github.com/whatwg/html/issues/7039), we only support `lang` and `dir` on the `<html>` element and on the host element itself. Applying language or directionality on an ancestor (other than `<html>`) isn't supported. 🤦🏻‍♂️

From our docs: https://webawesome.com/docs/localization#multiple-locales-per-page

> For performance reasons, the lang and dir attributes must be on the component itself, not on an ancestor element.

Carry on!

claviska (approved):

PR #1213, opened 2025-07-22
Fix dialog and drawer header

### LINKED ISSUE #1209: Dialog & Drawer don't show header when label is blank
The `without-header` attribute should be the source of truth for when the header shows, with the default being false. This deviates from the previous behavior which only showed the header if a label was provided.

We need to update both components to check for `this.withoutHeader` instead of checking for labels. And we should probably default the label to "Dialog" for both (since they're technically both dialogs and this would be announced to screen readers if a label is omitted).

claviska:
Fixes #1209 by ensuring the header is always shown based on the `without-header` attribute, not the detection of a label.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -8,6 +8,14 @@ Web Awesome follows [Semantic Versioning](https://semver.org/). Breaking changes
 
 Components with the <wa-badge variant="warning">Experimental</wa-badge> badge should not be used in production. They are made available as release candidates for development and testing purposes. As such, changes to experimental components will not be subject to semantic versioning.
 
+## Next
+
+### New Features {data-no-outline}
+
+### Bug Fixes and Improvements {data-no-outline}
+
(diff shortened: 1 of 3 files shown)


packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -395,4 +403,4 @@ Many of these changes and improvements were the direct result of feedback from u
 
 </details>
 
-Did we miss something? [Let us know!](https://github.com/shoelace-style/webawesome-alpha/discussions)
\ No newline at end of file
+Did we miss something? [Let us know!](https://github.com/shoelace-style/webawesome-alpha/discussions)
```
KonnorRogers: This link actually no longer exists.

KonnorRogers (dismissed):
looks good to me! just noticed the link to the dead repo 🙈

packages/webawesome/docs/docs/resources/changelog.md
claviska: Fixed

PR #1222, opened 2025-07-23
Update the Icon component to use Font Awesome 7

randomguy-2650:
#### Description

This PR updates the Icon component to use Font Awesome 7.

The `swap-opacity` attribute has also been implemented with this PR. Pro+ icon packs aren’t implemented in this PR.

---

### Reviewer Notes

- https://docs.fontawesome.com/web/style/icon-canvas#automatic-width-by-default (looks like it’s mainly about using CSS instead of bare SVGs, which doesn’t contain the JS package)


packages/webawesome/src/components/icon/icon.css
```diff
@@ -5,16 +5,19 @@
   --secondary-opacity: 0.4;
 
   display: inline-flex;
+  justify-content: center;
   box-sizing: content-box !important;
   width: auto;
-  height: 1em;
+  height: 1.25em;
```
claviska: I'm not sure we want to make this change.

packages/webawesome/src/components/icon/library.system.ts
```diff
@@ -9,31 +9,174 @@ export const icons: { [key: string]: { [key: string]: string } } = {
...
+      </svg>
+    `,
+    xmark: `
+      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 640 640">
+        <!--! Font Awesome Free v7.0.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2025 Fonticons, Inc. -->
+        <path
+          d="M183.1 137.4C170.6 124.9 150.3 124.9 137.8 137.4C125.3 149.9 125.3 170.2 137.8 182.7L275.2 320L137.9 457.4C125.4 469.9 125.4 490.2 137.9 502.7C150.4 515.2 170.7 515.2 183.2 502.7L320.5 365.3L457.9 502.6C470.4 515.1 490.7 515.1 503.2 502.6C515.7 490.1 515.7 469.8 503.2 457.3L365.8 320L503.1 182.6C515.6 170.1 515.6 149.8 503.1 137.3C490.6 124.8 470.3 124.8 457.8 137.3L320.5 274.7L183.1 137.4z"
+        />
+      </svg>
+    `,
```
claviska: If we move forward with this, let's compress these down into one line like the previous ones. Please leave the copyrights in place.

packages/webawesome/src/components/icon/library.system.ts
```diff
@@ -9,31 +9,174 @@ export const icons: { [key: string]: { [key: string]: string } } = {
...
+    star: `
+      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 640 640">
+        <!--! Font Awesome Free v7.0.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2025 Fonticons, Inc. -->
+        <path
+          d="M320.1 32C329.1 32 337.4 37.1 341.5 45.1L415 189.3L574.9 214.7C583.8 216.1 591.2 222.4 594 231C596.8 239.6 594.5 249 588.2 255.4L473.7 369.9L499 529.8C500.4 538.7 496.7 547.7 489.4 553C482.1 558.3 472.4 559.1 464.4 555L320.1 481.6L175.8 555C167.8 559.1 158.1 558.3 150.8 553C143.5 547.7 139.8 538.8 141.2 529.8L166.4 369.9L52 255.4C45.6 249 43.4 239.6 46.2 231C49 222.4 56.3 216.1 65.3 214.7L225.2 189.3L298.8 45.1C302.9 37.1 311.2 32 320.2 32zM320.1 108.8L262.3 222C258.8 228.8 252.3 233.6 244.7 234.8L119.2 254.8L209 344.7C214.4 350.1 216.9 357.8 215.7 365.4L195.9 490.9L309.2 433.3C316 429.8 324.1 429.8 331 433.3L444.3 490.9L424.5 365.4C423.3 357.8 425.8 350.1 431.2 344.7L521 254.8L395.5 234.8C387.9 233.6 381.4 228.8 377.9 222L320.1 108.8z"
+        />
+      </svg>
+    `,
```
claviska: Please compress :)

claviska (commented):
Thanks for the jumpstart on this. More testing is indeed needed, and I've left a few comments in the meantime.

packages/webawesome/src/components/icon/icon.css
randomguy-2650: This was done as the icon canvases in FA 7 has been updated to 20×20. I tested this beforehand and when reverting it to 1em, it is the correct font size. It just doesn’t look like it, as in FA 7, there usually is a 2 unit space around the entire icon.

Maybe we could scale all the paths in the SVG by 1.25, that would remove all the extra whitespace around the icon introduced in FA 7.

packages/webawesome/src/components/icon/library.system.ts
randomguy-2650: This is also resolved.

packages/webawesome/src/components/icon/icon.css
randomguy-2650: The paths have been scaled up by 1.25 and centred in the SVG. Resolved.

claviska:
Thanks so much for kicking this off! I grabbed your changes and finished it up in #1301.

PR #1228, opened 2025-07-24
(LICENSE): Update copyright year and improve typographical symbol usage

randomguy-2650:
#### Description

This PR updates the license file by updating the copyright year and replacing straight quotes ("") with smart quotes (“”), which looks nicer. This also updates (c) with ©.

In the license text [here](https://opensource.org/license/mit), it uses smart quotes as well, so I think it’s best to make it look nicer and be consistent.


claviska:
Thanks, but we're going to revise this as soon as we get the official commercial license in hand. :)

PR #1235, opened 2025-07-27
Fixed a typo in dropdown doc, "skidding" instead of "offset" attribute

nguyenyou:
Should be "skidding" instead of "offset" attribute.


claviska (approved):
Thanks!

PR #1236, opened 2025-07-27
Allow <wa-input type="file">

benabel:
Following the proposition in #1238.

I have added a visual test with some css to handle `input type="file"` and is `wa` counterpart.

image

### DIFF
--- a/packages/webawesome/src/styles/native.css
+++ b/packages/webawesome/src/styles/native.css
@@ -519,7 +519,6 @@
             [type='button'],
             [type='checkbox'],
             [type='color'],
-            [type='file'],
             [type='hidden'],
             [type='image'],
             [type='radio'],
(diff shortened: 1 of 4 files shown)


packages/webawesome/src/styles/native.css
```diff
@@ -859,8 +858,7 @@
     /* Exclude inputs that don't accept text */
     [type='button'],
     [type='checkbox'],
-    [type='color'],
-    [type='file'],
+    [type='color']
```
benabel, line 861: Needs a coma. I added it in my next branch but doesn't seem to be reflected here.

claviska:
Thanks for the PR. I think having a separate component is going to be better than overloading `<wa-input>` even more than it already is. We do plan on adding a File Input fairly soon, which you can track in #1240.

In the meantime, folks can use `<input type="file">` with their own styles. I do think it's reasonable that we add [Native Styles](https://webawesome.com/docs/utilities/native) for `<input type="file">`, but we need to get the design ironed out to match all WA themes, so I'm also opening #1241.

benabel:
Thanks for your answer @claviska and the awesome work in this library.

PR #1274, opened 2025-08-06
Do not crush labels with nested tags

### LINKED ISSUE #1272: wa-checkbox label with markup gets scrunched
### Describe the bug
If a wa-checkbox label includes nested tags, all spacing around those tags is removed . (wa-radio and sl-checkbox do not have this behavior.)

### To Reproduce

1. Create a wa-checkbox whose label (default slot) includes other tags, e.g.:
```html
<wa-checkbox>
  Enable <kbd>Enter</kbd> key <small>(desktop only)</small>
</wa-checkbox>
```
2. Look at it in a browser

Results: spacing is lost around the `<kbd>` and `<small>` elements, and they are aligned vertically centered on the text (rather than with its baseline)
Expected: compare to to sl-checkbox or wa-radio

Workaround: use an extra `<div>` inside the wa-checkbox.

### Demo

https://codepen.io/mvedmunds/pen/MYamMjR?editors=1000

### Screenshots
Image

### Additional information
3.0.0-beta.4

I think this is because wa-checkbox renders its [default slot directly as its label part](https://github.com/shoelace-style/webawesome/blob/56d678b50427c8279072138bd5daf95f2dda3feb/packages/webawesome/src/components/checkbox/checkbox.ts#L233). The `<slot>` element uses `display: contents` by default, so its contents are subject to the container's flex layout. (And a solution might be something like `slot[part="label"] { display: block; }` in its stylesheet.)

Shoelace tended to use [wrapper divs around its slots](https://github.com/shoelace-style/shoelace/blob/1ef287c3c13c794e8e5ce4fddf29511510c8c65b/src/components/checkbox/checkbox.component.ts#L263-L265), insulating their contents from the container layout. (Which in many cases led to other counterintuitive behavior. But in wa-checkbox, this seems wrong.)

whjvenyl:
fixes https://github.com/shoelace-style/webawesome/issues/1272
This pull request makes small but important improvements to the layout and styling of the `button` and `checkbox` components in the UI library. The changes ensure better alignment and consistent display of labels for these components.

Styling improvements for label display:

* Updated `.label` styling in `button.css` to use `inline-block` display for button labels except when the button is an icon button, improving label alignment.
* Added a rule to `checkbox.css` to set `[part~='label']` as `inline-block`, ensuring proper label display for checkboxes.

### DIFF
--- a/packages/webawesome/src/components/button/button.css
+++ b/packages/webawesome/src/components/button/button.css
@@ -184,6 +184,10 @@
 }
 
 .label {
+  display: inline-block;
+}
+
+.is-icon-button .label {
   display: flex;
 }
 
--- a/packages/webawesome/src/components/button/button.css
+++ b/packages/webawesome/src/components/button/button.css
@@ -186,6 +186,9 @@
 .label {
   display: flex;
 }
+:not(.is-icon-button) .label {
+  display: inline-block;
+}
 
 .label::slotted(wa-icon) {
   align-self: center;
--- a/packages/webawesome/src/components/checkbox/checkbox.css
+++ b/packages/webawesome/src/components/checkbox/checkbox.css
@@ -45,6 +45,9 @@
   vertical-align: middle;
   cursor: pointer;
 }
+[part~='label'] {
+  display: inline-block;
+}
 
 /* Checked */
 [part~='control']:has(:checked, :indeterminate) {
--- a/packages/webawesome/src/components/checkbox/checkbox.css
+++ b/packages/webawesome/src/components/checkbox/checkbox.css
@@ -45,6 +45,7 @@
   vertical-align: middle;
   cursor: pointer;
 }
+
 [part~='label'] {
   display: inline;
 }
--- a/packages/webawesome/src/components/checkbox/checkbox.css
+++ b/packages/webawesome/src/components/checkbox/checkbox.css
@@ -45,6 +45,9 @@
   vertical-align: middle;
   cursor: pointer;
 }
+[part~='label'] {
+  display: inline;
+}
 
 /* Checked */
 [part~='control']:has(:checked, :indeterminate) {


packages/webawesome/src/components/button/button.css
```diff
@@ -186,6 +186,9 @@
 .label {
   display: flex;
 }
+:not(.is-icon-button) .label {
+  display: inline-block;
+}
```
lindsaym-fa:
```suggestion
.label {
  display: inline-block;
}

.is-icon-button .label {
  display: flex;
}
```

This works perfectly, but mind if we swap the order here? This might better communicate that `inline-block` is the default.

packages/webawesome/src/components/checkbox/checkbox.css
```diff
@@ -45,6 +45,9 @@
   vertical-align: middle;
   cursor: pointer;
 }
+[part~='label'] {
+  display: inline-block;
```
lindsaym-fa:
```suggestion
  display: inline;
```

We use `inline` for `wa-radio`'s labels, so we may want to make this one consistent.

lindsaym-fa (commented):
Thanks, @whjvenyl! This is working great. Just a couple of suggestions to ensure consistency.

lindsaym-fa (dismissed):

packages/webawesome/src/components/checkbox/checkbox.css
```diff
@@ -45,6 +45,9 @@
   vertical-align: middle;
   cursor: pointer;
 }
+[part~='label'] {
```
claviska, line 49:
```suggestion

[part~='label'] {
```

claviska (approved):

claviska:
I'm not sure why Vercel is fussing about this, but the changes and the workflow have been approved so I'm going to merge this.

Issue #1224, opened 2025-07-23
Make tooltip border easier to customize


claviska:
### Discussed in https://github.com/shoelace-style/webawesome/discussions/1221

<div type='discussions-op-text'>

<sup>Originally posted by **zedix** July 23, 2025</sup>
Hi Web Awesome team 👋🏻

I wanted to suggest opening up the possibility of customizing the `border` (and possibly the `padding` and `box-shadow`) of the tooltip component. This would be especially useful when displaying a tooltip with a light background, where a border becomes almost essential for visibility:

```css
/* wa-tooltip stylesheet */
.body {
   border: var(--wa-tooltip-border, 0);
   padding: var(--wa-tooltip-padding, 0.25em 0.5em);
}
```

While it's possible to achieve this level of customization using `wa-tooltip::part(body)`, it can become redundant (parts + exported parts) when aiming for design consistency across components — for example, between `wa-tooltip` and `wa-slider` tooltips.

<img width="235" height="143" alt="image" src="https://github.com/user-attachments/assets/bda5dd1a-154c-4b0f-8a01-3fdef625f58c" />

A side note that exported parts may not be right in the `wa-slider` to achieve this:

<img width="467" height="162" alt="image" src="https://github.com/user-attachments/assets/c4b1da96-2696-445a-af68-1a29b28755ba" />

Thanks

</div>

claviska:
Currently, the user has to do this with two parts and it's not very obvious:

```css
wa-tooltip::part(body) {
  border: solid 2px tomato;
}

wa-tooltip::part(base__arrow) {
  border-bottom: solid 2px tomato; 
  border-right: solid 2px tomato; 
}
```

We can probably add `--border-width` and `--border-color` custom properties for the tooltip. @lindsaym-fa do you think we need `--border-style` as well?

lindsaym-fa:
We actually have a number of design tokens for tooltip styles:

- `--wa-tooltip-arrow-size`
- `--wa-tooltip-background-color`
- `--wa-tooltip-border-radius`
- `--wa-tooltip-content-color`
- `--wa-tooltip-font-size`
- `--wa-tooltip-line-height`

It seems more than reasonable to add `--wa-tooltip-border-width` and `--wa-tooltip-border-color`. `--wa-tooltip-border-style` also makes sense to round out the group (especially considering we have similar tokens for other component groups, like `--wa-form-control-border-style` and `--wa-panel-border-style`).

---

[...]

PR #1277, opened 2025-08-06
Fix exportparts of tooltip inside wa-slider component

zedix:
This fix is a follow-up to https://github.com/shoelace-style/webawesome/issues/1224

Fix the `exportparts` of tooltip which seems incorrect.

❌ Incorrect:

```css
wa-slider::part(tooltip__content) {
  /* customize tooltip body */
}
```

✅ Expected:

```css
wa-slider::part(tooltip__body) {
  /* customize tooltip body */
}
```


claviska (approved):
Good catch. Thanks for the PR!

PR #1284, opened 2025-08-07
Working on fixing element update warnings

KonnorRogers:


packages/webawesome/scripts/build.js
```diff
@@ -218,8 +218,9 @@ export async function build(options = {}) {
       outdir: getCdnDir(),
       chunkNames: 'chunks/[name].[hash]',
       define: {
-        'process.env.NODE_ENV': '"production"', // required by Floating UI
+        'process.env.NODE_ENV': '"development"', // required by Floating UI
       },
+      conditions: ['development'],
```
KonnorRogers, line 223: We need to make sure we have a conditional here so we don't ship production builds with "dev" versions of Lit.

KonnorRogers:
Closing for now, will reference in future to get the work done.

Issue #1127, opened 2025-07-02
Update appearances to be type safe


claviska:
I believe the only change required is to use "filled-outlined" instead of "filled outlined", which will solve the TypeScript/tooling issue and make it more explicit as to _which_ combinations of appearances are acceptable.

I'm open to other ideas, but this is the simplest solution I can think of.

Affected components that use filled/outlined styles:

- [x] Badge
- [x] Button
- [x] Callout
- [x] Details
- [x] Input
- [x] Select
- [x] Tag
- [x] Textarea

---

Related discussions:
- #1290
- #1373

lindsaym-fa:
Agreed that adding `filled-outlined` is the way to go! (Apologies, I thought I had responded to this already.)

claviska:
Fixed in #1586

PR #1290, opened 2025-08-08
Add missing appearance to button.ts

rlconst:
According to https://webawesome.com/docs/components/button/ I can use 'filled outlined'


claviska:
Thanks for the PR! We're tracking this in https://github.com/shoelace-style/webawesome/issues/1127 but I'm not sure we have consensus on a solution yet from @lindsaym-fa.

I'm going to close this because:

- Most likely we'll be adding it as `filled-outlined`
- Other components are affected
- Docs will need to be updated

rlconst:
My bad. Should've search first8 авг. 2025 г., в 21:10, Cory LaViska ***@***.***> написал(а):﻿
Closed #1290.

PR #1295, opened 2025-08-09
Add jsx types to package.json exports

ajmchambers:
Noticed this was missing while trying out the new JSX types and attempting to extend JSX's IntrinsicElements.


claviska:
This looks good to me. The only question I have for @KonnorRogers is whether we should export it as is:

```
./dist/custom-elements-jsx.d.ts
```

Or from a more ergonomic endpoint, e.g.

```
./jsx
```

claviska (approved):

PR #1299, opened 2025-08-12
Use button-group variant in child buttons

### LINKED ISSUE #1298: Button Group variant is not working
### Describe the bug
a `wa-button-group` doesn't forward it's `variant` prop to the child `wa-button`s

### To Reproduce

1. Go to [docs](https://webawesome.com/docs/components/button-group/#theme-buttons)
2. All buttons are neutral, but they should be themed.

Image

wendevlin:
fix #1298

- I wasn't able to let the `wa-button` overwrite the groups variant, so I removed this docs for now.

### DIFF
--- a/packages/webawesome/docs/docs/components/button-group.md
+++ b/packages/webawesome/docs/docs/components/button-group.md
@@ -119,16 +119,6 @@ Theme buttons are supported through the button group's `variant` attribute.
 </wa-button-group>
 ```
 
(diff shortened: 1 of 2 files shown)


claviska:
Thanks for the PR. We've decided against this behavior in https://github.com/shoelace-style/webawesome/issues/1298#issuecomment-3202030430 because keeping it in sync will require an expensive mutation observer. I'll make sure the docs are aligned with the intended behavior.

PR #1316, opened 2025-08-15
Add `stroke-width` style based on `--track-width` to `<wa-spinner>`

curtiskeller:
Update spinner to respect track width property. Issue https://github.com/shoelace-style/webawesome/issues/1317

### DIFF
--- a/packages/webawesome/src/components/spinner/spinner.css
+++ b/packages/webawesome/src/components/spinner/spinner.css
@@ -24,12 +24,14 @@ svg {
 
 .track {
   stroke: var(--track-color);
+  stroke-width: var(--track-width);
 }
 
 .indicator {
   stroke: var(--indicator-color);
   stroke-dasharray: 75, 100;
   stroke-dashoffset: -5;
+  stroke-width: var(--track-width);
   animation: dash 1.5s ease-in-out infinite;
   stroke-linecap: round;
 }


claviska:
We may want to use em units instead of pixels, as this changes makes the spinner look thinner at various sizes.

Current
CleanShot 2025-08-19 at 15 41 15@2x

Proposed
CleanShot 2025-08-19 at 15 41 50@2x

@lindsaym-fa what do you think?

Maintainer notes:

- [ ] Needs changelog

lindsaym-fa:
I fell down a bit of a rabbit hole with this one, but **if we'd like to keep the scope of this issue contained, I'd recommend setting `--track-width: 4px` by default.** 

It seems because we're setting a `viewBox` and `cx`, `cy`, and `r` attributes that are relative to that `viewBox` on the internal `<svg>` elements, then scaling the `viewBox` based on the font size, the pixel values within consequently scale proportional to the font size too. Using `em` values has the effect of exaggerating the scaling since the stroke essentially scales twice — once because of the applied `em` value, and again because of the scaled `viewBox`.



And there's the 🐰🕳️: 4px =/= 4px. That's seems...unexpected.

For example, here's what `--track-width: 4px` looks like in the doc's Size example:

https://github.com/user-attachments/assets/cc49ce2f-d00e-49fd-bcc4-91c6f166c649

None of those strokes are exactly 4px. The largest example, which uses `font-size: 3rem`, is the closest. In any case, we wouldn't expect 4px to scale with the font size.

Here's what `--track-width: 0.25em` looks like for comparison:

https://github.com/user-attachments/assets/f5553b18-0a5c-4c7f-9b69-b149c7a8a972

We'd expect the stroke width of the first to render as 4px, the second as 8px, and the third as 12px. Instead, the first looks closer to ~1px, the second ~5px, and the third ~10px. And, notably, the svg starts to be clipped at the largest size.

There's more to resolve here to ensure that the applied `--track-width` is accurate. @claviska, do we want to add `--track-width` in this PR (by the way, thank you for catching this @curtiskeller!) and investigate the overall scale/sizing issue separately? Or should we handle them all together?

claviska:
Great analysis. I would vote for handling them all together unless there’s a good reason not to.

curtiskeller:
No problem! Just to confirm, you would like me to change the `--track-width` default value to 4px, then leave the rest of the changes as-is to handle track and indicator at the same time? @lindsaym-fa

lindsaym-fa:
No further action required from you right now, @curtiskeller! I'll start digging into how we can fix the sizing for `<wa-spinner>` so that the set `--track-width` is actually accurate, and we'll merge your commit along with that fix when it's ready. I suggested 4px because it works well for `<wa-spinner>` in its current state, but once we fix the underlying sizing issue, a different value might be more appropriate. I'll keep you posted!

claviska:
Alternative PR: https://github.com/shoelace-style/webawesome/pull/2235

claviska:
This has been fixed in #2235.

Issue #1318, opened 2025-08-15
[wa-callout][wa-tag] - unimplemented appearance typing `outlined accent`


curtiskeller:
For these 2 components it appears there is an unimplemented appearance type `outlined accent`. The current css looks like it would conflict on border-color. Documentation does not mention this appearance as well.

PR: https://github.com/shoelace-style/webawesome/pull/1319

PR #1319, opened 2025-08-15
#1318 - remove unimplemented appearance typing

curtiskeller:
Remove typings for unimplemented appearances. `outlined accent`

Issue: https://github.com/shoelace-style/webawesome/issues/1318


claviska:
Thanks for catching this!

PR #1337, opened 2025-08-19
Allow the slider's thumb to receive focus when tapping or dragging

### LINKED ISSUE #1312: [wa-slider] does not focus when clicked
### Describe the bug
Clicking on the slider does not focus it and arrow keys can't be used to move the thumb and change the value.

This is due to calling `preventDefault` on the `pointerDown` event [here](https://github.com/shoelace-style/webawesome/blob/next/packages/webawesome/src/internal/drag.ts#L89). Instead of calling preventDefault, consider using [touch-action](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) to prevent scrolling.

### To Reproduce

1. Click on the slider track to set the value
2. Use arrow keys to move the value via thumb

Expected: arrow keys move thumb and value changes
Actual: slider is not focused and arrow keys do not move thumb/value

### Demo

[repro](https://lit.dev/playground/#gist=89cc5f44da71332aa7216c4f49d3514a)

### Browser / OS
 - OS: [e.g. Mac, Windows]
 - Browser: [e.g. Chrome, Firefox, Safari]
 - Browser version: [e.g. 22]

claviska:
Fixes #1312

Thanks, @sorvell!

### DIFF
--- a/packages/webawesome/custom-elements-manifest.js
+++ b/packages/webawesome/custom-elements-manifest.js
@@ -182,6 +182,10 @@ export default {
     jsxTypesPlugin({
       fileName: 'custom-elements-jsx.d.ts',
       outdir,
+      defaultExport: true,
+      componentTypePath: (_name, _tag, modulePath) => {
+        return `./${modulePath}`;
+      },
     }),
 
     //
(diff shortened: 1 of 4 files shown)


claviska:
ugh, some of #1336 leaked into this one 🫠

PR #1343, opened 2025-08-20
Rename `icon-position` to `icon-placement`

claviska:
Per https://github.com/shoelace-style/webawesome/discussions/1340, this renames the recently introduced `icon-position` attribute on `<wa-details>` to `icon-placement`.

Placement is used for start/end/top/bottom/etc. and position is generally reserved for numeric values and percentages, so this makes the library more consistent in its use of those terms.


lindsaym-fa (dismissed):
Good call!

PR #1346, opened 2025-08-20
Apply `position: relative` only to buttons that contain badges

### LINKED ISSUE #1226: Sticky th strange rendering behaviour
### Describe the bug
When using sticky th in tables, some elements do not scroll behind the th, but instead _above_ it, for example:
wa-button
wa-icon if opacity is not 1

### To Reproduce
See demo below

### Demo
See CodePen at [this link](https://codepen.io/Stefano-Rivoir/pen/yyYJJWx)

If you change wa-icon opacity to 1, wa-icons are rendered correctly.
Buttons are always rendered above the th.

### Screenshots
None

### Browser / OS
 - OS: Windows
 - Browser: Chrome
 - Browser version: 138

### Additional information
Shoelace did not have this problem

lindsaym-fa:
Partially addresses #1226.

Relative positioning is only necessary on buttons when they contain badges in order to support positioning the badge at the top right corner of buttons, shown here: https://webawesome.com/docs/components/badge/#with-buttons 
This change applies `position: relative` under this condition rather than by default.

### DIFF
--- a/packages/webawesome/src/components/button/button.css
+++ b/packages/webawesome/src/components/button/button.css
@@ -1,6 +1,9 @@
 @layer wa-component {
   :host {
     display: inline-block;
+  }
+
+  :host(:has(wa-badge)) {
     position: relative;
   }
 }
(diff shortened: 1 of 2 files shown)


packages/webawesome/src/components/button/button.css
```diff
@@ -1,6 +1,9 @@
 @layer wa-component {
   :host {
     display: inline-block;
+  }
+
+  :host(:has(wa-badge)) {
```
claviska, line 6: ezgif-5179646c5b0ff4

claviska (dismissed):

claviska (approved):

PR #1351, opened 2025-08-20
Fixes spacing between clear and password toggle icons in `<wa-input>`

### LINKED ISSUE #1325: wa-input, spacing between toggle and clearable inner buttons
### Describe the bug
When a wa-input has both with-clear and password-toggle attribute, the two buttons are not spaced

### To Reproduce
See demo

### Demo

[Codepen](https://codepen.io/Stefano-Rivoir/pen/myeXMjR)

### Screenshots

Image

### Browser / OS
 - OS: Windows
 - Browser: Chrome
 - Browser version: 139

claviska:
I went with 2xs spacing, but feel free to adjust.

CleanShot 2025-08-20 at 17 30 11@2x

Closes #1325


lindsaym-fa (approved):
I used `--wa-form-control-padding-inline` and moved the style to apply to the clear button and the password toggle whenever either of them are present — doesn't seem like a bad idea to always put a bit of space between these and the input value, and it closely matches how we style the clear and expand icons in `<wa-select>`.

I did notice that our system icons seem smaller than `<wa-icon>` icons, which gives the appearance of more space between the clear and password toggle icons than there actually is. I'll create a separate issue for this tomorrow!

ETA: Created #1356 for this☝️.

Issue #105, opened 2024-05-13
Toast Stack + Toast Item


claviska:
As part of #104, alerts will no longer be the thing that enables toast notifications. Toasts were shimmed into alerts and need accessibility improvements anyways — so let's do it right this time.

Moving forward, we will add `<wa-toast>` and `<wa-toast-item>` to the library and provide utility functions to initiate toast notifications.

### Open questions

- Do we have a rough design for what toasts will look like?

/cc @lindsaym-fa

steunix:
Just out of curiosity, will this component be included in one of the next betas or will they be added later in the 3.x stable cycle? It's the last bit of WA I'm waiting and longing for ;)

claviska:
I can see them shipping in a beta version. I don't have a definite timeline for this one just yet, though!

steunix:
> I can see them shipping in a beta version. I don't have a definite timeline for this one just yet, though!

That's great. Keep up the great work!

lindsaym-fa:
re: design, I'm imagining slots for `icon`, content (default), and `action` with a close button + progress ring to show a countdown of when the toast will close automatically, along with a `without-close` attribute to optionally hide the close button.

Something like this:

<img width="420" height="117" alt="default toast with a subtle outlined style, box shadow, and minimal brand coloring" src="https://github.com/user-attachments/assets/5a7be5ac-c601-4af5-a08b-1e9cd0dde1c3" />

With `variant=""` and `appearance=""` options:

<img width="808" height="653" alt="toasts in brand, neutral, success, warning, and danger variants followed by outlined, filled, filled + outlined, and accent appearances" src="https://github.com/user-attachments/assets/bd386184-28db-44dd-b9fb-d22beaf0f373" />

Here's a rough pen for the styles: https://codepen.io/omnifides/pen/XJmRQmp

h/t to @claviska for the inspiration of surrounding the close button with the progress indicator. 🌟

ricky11:
Toast is what we're waiting for before we can think of deploying this to prod, what can we use in replacement of it, wonder why it was left out as core component in most ui's.

GedMarc:
I added one into the angular-awesome npm package in the meantime if you are using angular
Also for us, its considered a base component hehe

ricky11:
using vue 😁, and need my toast.

aaronpinero:
[...]

PR #1380, opened 2025-08-29
Alert component

whjvenyl:
Based on the discord discussion https://discord.com/channels/739437527907303535/1409143004580675624 I thought I'd post my alert implementation as well.

It's not rocking those fancy visuals from https://github.com/shoelace-style/webawesome/issues/105 that @lindsaym-fa made, but show's how a possible wrapper solution around a `wa-callout` could look like.


claviska:
Thanks for the inspiration. Our plan is to introduce `<wa-toast>` and `<wa-toast-item>` which, as you pointed out, @lindsaym-fa has sketch out in #105.

While it's probably possible to reuse callout, there are accessibility considerations for the toast stack and the toast item, as well as additional features such as the timer that would need to be baked in. Because of that, I'm not sure it makes sense to build toasts on top of it.

I'm closing this to get it off our PR review list, but feel free to keep the discussion going!

PR #1395, opened 2025-09-03
Reflect icon properties to its attributes

### LINKED ISSUE #1389: WaIcon properties no longer reflect
In Shoelace, the `SlIcon` component would reflect properties like `name` and `library`. I was surprised to see that `WaIcon` did not behave the same in this regard. I noticed that in #447, the `reflect` setting was replaced on those properties with a `cssProperty` setting, but then the `cssProperty` setting was removed in #997. However, in that latter PR, the original `reflect` setttings were not brought back. Would there be a downside to bringing back the reflectiveness of those WaIcon properties that was present before the former PR was merged?

### Context

In my situation, I'm ultimately trying to work around the fact WaIcon sets the `fill` property to `currentColor` on its `svg` CSS part. This causes some of the SVGs I receive from my UX team to render incorrectly, namely those SVGs that set their `fill` attribute to `none` in their `<svg>` root element (the CSS part's style overrides the value set in the SVG). For example, our icon of an eye ends up looking like a football:

Image

Where instead it should look more like this (this is when I disable the `fill` property on the `svg` part in Dev Tools):

Image

I've tried to compensate for this by writing styles like this to get the browser to "forget" the part's style and use the fill setting in the SVG, but so far I haven't been able to get this to work:

```css
wa-icon::part(svg) {
  fill: revert-layer; /* also tried inherit, unset, revert, initial, but no dice. */
}
```

So my next plan was to write styles to target specific icons and get their fills correct like this:

```css
wa-icon[name="eye"]::part(svg) {
  fill: none;
}
```

But then that didn't work in the Storybook project I'm trying use to demo these icons. Storybook's MDX implementation uses React to render docs and component styles side-by-side. For reasons I don't totally understand, rendering a WaIcon either as a custom element or using the WaIcon React wrapper in React's JSX sets the `name` prop as a property and not an attribute in this case. Because of that plus the fact WaIcon doesn't reflect its properties now (below is the markup that is ultimately rendered), the above style has no effect.

Image

Contrasting that with SlIcon, the similar style works as expected because the rendered markup has the reflected attribute.

Image

### Alternatives

As I was writing this up, I realized I can address this by adjusting individual path elements in the SVG icons to directly set their fill properties. So while I'm not necessarily blocked by this issue, I think I'd still like to see if ultimately WaIcon's properties are supposed to ultimately reflect like they did prior to #447?

mvromer:
Fixes #1389. This simply adds back the `reflect` setting removed on some of WaIcon's properties in #447.

### DIFF
--- a/packages/webawesome/src/components/icon/icon.ts
+++ b/packages/webawesome/src/components/icon/icon.ts
@@ -46,7 +46,7 @@ export default class WaIcon extends WebAwesomeElement {
   @state() private svg: SVGElement | HTMLTemplateResult | null = null;
 
   /** The name of the icon to draw. Available names depend on the icon library being used. */
-  @property() name?: string;
+  @property({ reflect: true }) name?: string;
 
   /**
    * The family of icons to choose from. For Font Awesome Free, valid options include `classic` and `brands`. For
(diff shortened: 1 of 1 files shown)


claviska (approved):
Thanks!

PR #1406, opened 2025-09-05
Add intersection observer component

claviska:
Adds the `<wa-intersection-observer>` utility component.

@lindsaym-fa feel free to update any of the docs or examples as you see fit 😄


lindsaym-fa (approved):
This looks great to me! 

I don't have any immediate/quick suggestions for the docs, but this would be a great component to use with patterns — especially our full page ones. Hoping I have some time to explore that during cooldown. 😄

claviska:
@lindsaym-fa pre-baked animation utility classes would work really well with it too.

PR #1415, opened 2025-09-08
Allow bundlers access to package.json

roydukkey:
There are cases where tooling extending Web Awesome can get important information from the `package.json` (i.e. version). Exporting the file will given bundlers access to the file.

### DIFF
--- a/packages/webawesome/package.json
+++ b/packages/webawesome/package.json
@@ -29,7 +29,8 @@
     "./dist/react": "./dist/react/index.js",
     "./dist/react/*": "./dist/react/*",
     "./dist/translations": "./dist/translations",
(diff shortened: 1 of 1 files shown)


claviska:
May I ask which bundler you're having trouble with? I've never run into this issue before.

roydukkey:
Here is a [playground](https://stackblitz.com/edit/vitejs-vite-y5aksmrp?file=src%2Fmain.js,package.json&terminal=dev) that shows the issue.

```
[plugin:vite:import-analysis] Missing "./package.json" specifier in "@awesome.me/webawesome" package
```

KonnorRogers (approved):
Totally forgot about import attributes, good catch!

PR #1435, opened 2025-09-11
fixes to the card component

kelseythejackson:
I made some updates to the vertical card that created some regressions.

Screenshot 2025-09-11 at 11 24 38 AM

This caused the utility classes to no longer apply.

I jumped in the code and added some conditionals to the render method because creating an empty slot that won't be used didn't feel good.

Added a `flex-grow` of 1 to the `header` and `footer` slots so that the classes applied will continue to work.

Screenshot 2025-09-11 at 12 37 08 PM


packages/webawesome/src/components/card/card.css
```diff
@@ -76,10 +76,20 @@
   }
 }
 
+/* Make these slots "full width"  */
+::slotted([slot='header']),
+::slotted([slot='footer']) {
+  flex-grow: 1;
```
lindsaym-fa: Looks like this causes the images in our docs indices to be left-aligned instead of centered.

I wonder: is this necessary? Would the `.has-actions` addition you made below be sufficient?

Before:
Screenshot 2025-09-11 at 1 42 38 PM

After:
Screenshot 2025-09-11 at 1 42 33 PM

KonnorRogers: its probably needed to recreate the old block behavior?

packages/webawesome/src/components/card/card.css
```diff
@@ -76,10 +76,20 @@
...
+}
+
+/* Push slots to sides when the action slots renders */
+.has-actions {
+  justify-content: space-between;
+}
```
KonnorRogers: Any reason not to apply the flex only when actions are slotted?

```suggestion
.has-actions {
  display: flex;
  align-items: center;
  justify-content: space-between;
}
```

packages/webawesome/src/components/card/card.css
```diff
@@ -76,10 +76,20 @@
...
+
 .header {
   display: flex;
   align-items: center;
```
KonnorRogers:
```suggestion
.header {
  display: block;
```

packages/webawesome/src/components/card/card.css
```diff
@@ -94,7 +104,6 @@
 .footer {
   display: flex;
   align-items: center;
```
KonnorRogers:
```suggestion
.footer {
  display: block;
```

KonnorRogers (commented):
I don't know how much we plan to keep the "old" behavior of cards, but any reason not to conditionally apply flex only when actions are slotted?

lindsaym-fa:
I committed what was missing from @KonnorRogers suggested changes, updated some of the docs content, and also fixed an issue where images in horizontal cards would get squashed.

Anyone care to take another look? Cool to merge?

lindsaym-fa (approved):

claviska:
@lindsaym-fa looks like this just needs a changelog entry.

lindsaym-fa:
Changelog added. (I also snuck in a missing changelog entry for the Figma plugin while I was here. 🤫)

claviska (approved):

KonnorRogers (approved):

PR #1437, opened 2025-09-12
Fix popup arrow to avoid overlays

wendevlin:
- Add a clip-path to the popup arrow
  - this avoids overlay with the container when padding is reduced or the arrow size is increased.
-  original PR: home-assistant/webawesome#8

### DIFF
--- a/packages/webawesome/src/components/popup/popup.css
+++ b/packages/webawesome/src/components/popup/popup.css
@@ -49,6 +49,7 @@
   rotate: 45deg;
   background: var(--arrow-color);
   z-index: 3;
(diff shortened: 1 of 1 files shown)


claviska:
It looks like this clips too much, causing borders to show through. This is from the [popover component](https://webawesome-git-fork-wendevlin-fix-popup-arrow-font-awesome.vercel.app/docs/components/popover) from the preview.

CleanShot 2025-09-12 at 09 42 00@2x

claviska:
Closing since this causes a visual artifact we don't want, but if there's a better solution I'm happy to reopen.

Issue #1447, opened 2025-09-14
Icon button with caret is clipped


medmunds:
### Describe the bug
A wa-button that contains only a wa-icon in its default slot is constrained to be square (an "[icon button](https://webawesome.com/docs/components/button/#icon-buttons)"). But if the wa-button also has the with-caret attribute, both the icon and the caret are clipped by the square button.

### To Reproduce

1. Render this:
   ```html
   <wa-button with-caret>
     <wa-icon name="save" label="Save"></wa-icon>
   </wa-button>
   ```

Results: clipped icon and caret (see red box in the screenshot below)
Expected: see last example in the screenshot below

### Demo

https://codepen.io/mvedmunds/pen/xbwoJdG?editors=1000

### Screenshots

<img width="279" height="457" alt="Image" src="https://github.com/user-attachments/assets/4a95f2d1-5a45-462d-9e9d-3abab4268d8c" />

### Browser / OS
 - Web Awesome 3.0.0-beta.5

### Additional information
Adding `slot="start"` to the wa-icon prevents the clipping, but results in excessive spacing between the icon and the caret. (Compare the last two items in the screenshot above.)

claviska:
Fixed in #1461

PR #1461, opened 2025-09-17
update to the button  if it has an icon in it's default slot

kelseythejackson:
This is a PR for #1447 that should fix the issue the user was having when adding an icon to the default slot of a button with the `with-caret` attribute.  
Screenshot 2025-09-17 at 2 53 59 PM


claviska (approved):
Looks good to me in all sizes. Thanks, @kelseythejackson!

PR #1467, opened 2025-09-18
Relative links to `cluster` and `stack` fixed

bryceglass:
Fixed some internal links from `/docs/layout/` to `docs/utilities/`


claviska (approved):
Thanks!

PR #1473, opened 2025-09-19
Remove duplicated declaration

agraboso:
A trivial change to remove a duplicated declaration in the `<wa-split-panel>` component.


claviska (approved):
Nice catch. Thanks for the PR

PR #1486, opened 2025-09-23
updated offset value

### LINKED ISSUE #1472: Checkbox and Switch required label indicator is missing spacing
### Describe the bug
The required label indicator "*" is not offset from the label and is instead joined to it. `wa-input`, `wa-textarea`, and `wa-select` all have a gap between the label and the indicator.

### To Reproduce

1. Add the `required` attribute to a `wa-slider` with a label

### Demo

https://codepen.io/roydukkey/pen/VYewpoy?editors=1000

### Screenshots

Image

Image

### Browser / OS

```
  System:
    OS: macOS 15.6.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 88.97 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Brave Browser: 138.1.80.120
    Safari: 18.6
```

### Additional information
Provide any additional information about the bug here.

kelseythejackson:
A bug fix per #1472, this updates the spacing of the required indicator next to the label on certain inputs

Fixes #1472


claviska:
This solves the issue for checkboxes and switches:

CleanShot 2025-09-23 at 10 57 03@2x

CleanShot 2025-09-23 at 10 57 14@2x

But it also causes input, textarea, slider, et al to have an extra gap:

CleanShot 2025-09-23 at 10 57 41@2x

CleanShot 2025-09-23 at 10 57 52@2x

The difference between checkbox/switch and the others is that the former are `display: inline-flex` so these components may need to be adjusted independently instead.

packages/webawesome/src/components/checkbox/checkbox.css
```diff
@@ -96,5 +96,5 @@ input {
 :host([required]) [part~='label']::after {
   content: var(--wa-form-control-required-content);
   color: var(--wa-form-control-required-content-color);
-  margin-inline-start: var(--wa-form-control-required-content-offset);
+  margin-inline-start: calc(var(--wa-form-control-required-content-offset) * -1.5);
```
lindsaym-fa: Multiplying by a negative value here causes an inverse relationship between the `--wa-form-control-required-content-offset` token and the final margin, so when we change the token to a positive value like `0.25em`, we see results like this:

Image

Ideally, folks should be able to set `--wa-form-control-required-content-offset` to whatever value they want, so however that token is set should yield expected and consistent results across all components.

lindsaym-fa (changes requested):
We're closer, but this approach still has unexpected consequences (explained in the next comment).

The crux of the issue seems to that this pair of selectors in `form-control.css` places the required indicator on different internal elements depending on the component:
```css
:is([part~='form-control-label'], [part~='label']):has(*:not(:empty)) {
    ...
    :host([required]) &::after {
        ...
    }
}
```

So, in the shadow root for e.g. `<wa-input>`, we're seeing `::after` rendered for the `<label>` element:
```html
<label part="form-control-label label" class="label" for="input" aria-hidden="false">
    <slot name="label">Label</slot>
    ::after
</label>
```

Whereas in the shadow root for e.g. `<wa-checkbox>`, we're seeing `::after` rendered for the `<slot>` element:
```html
<label part="base">
    <span part="control">...</span>
    <slot part="label">
        ::after
    </slot>
</label>
```

Ultimately, this causes an **extra** space for most form controls (other than switch and checkbox) that isn't intended.

These changes might get us to where we need to be (more testing required):
1. Un-nest the `:host[required]` selector in `form-control.css` and change it to something like `:host([required]) slot:is([name='label'], [part='label'])::after {...}`  This ensures we're always adding `::after` to the relevant `<slot>` element instead of something different.
2. Update the default `--wa-form-control-required-content-offset` value of all themes to be positive `0.1em` since we're no longer working against the extra, unintended space.

kelseythejackson:
Tested them out, should be good to go.

Screenshot 2025-10-01 at 9 59 23 AM
Screenshot 2025-10-01 at 9 59 16 AM
Screenshot 2025-10-01 at 9 59 02 AM
Screenshot 2025-10-01 at 9 58 54 AM

lindsaym-fa (approved):
Fix approved! ✨ We'll also need:

1. A changelog entry
2. Another PR for the Pro repo to update the token value for Pro themes

claviska:
@lindsaym-fa the pro repo PR is here: https://github.com/shoelace-style/webawesome-pro/pull/61

claviska:
Changelog added. With @lindsaym-fa's final blessing on https://github.com/shoelace-style/webawesome-pro/pull/61 this is ready to merge.

lindsaym-fa:
Good to go!

PR #1488, opened 2025-09-23
Docs: Add Theme-based Scripts + Styles Templating Blocks

talbs:
This work adds a couple of new nunjucks [blocks](https://mozilla.github.io/nunjucks/templating.html#block) into our `base.njk` layout and `header.njk`.

The purpose of these is to allow for easy and safe overriding of the current theme as set by the Docs' theme-switching logic, including:
- setting the current theme
- loading the current theme's CSS/assets
- adding the current theme's classes to the `<html>` element.

### DIFF
--- a/packages/webawesome/docs/_includes/base.njk
+++ b/packages/webawesome/docs/_includes/base.njk
@@ -26,20 +26,22 @@
       document.documentElement.classList.toggle('wa-dark', isDark);
     </script>
 
-    <script type="module">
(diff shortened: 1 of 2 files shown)


talbs:
Closing this PR and direction in favor of https://github.com/shoelace-style/webawesome/pull/1491 (which is mucho mejor por que @KonnorRogers es muy inteligente y guapo).

PR #1490, opened 2025-09-24
Docs: Update Header

talbs:
This PR **updates the documentation header** by converting theme and color scheme selectors from `wa-select` to `wa-dropdown` components and streamlining the header layout. 

The changes improve the visual design by using icon-only buttons with tooltips instead of filled selectors.

- Refactored theme and color scheme selectors to use dropdown components with icon triggers
- Enhanced tooltip functionality with dynamic content updates during theme transitions
- Simplified header layout and moved GitHub buttons to a dedicated include file
- Moved version + beta info to bottom of Sidebar content

### DIFF
--- a/packages/webawesome/docs/_includes/sidebar.njk
+++ b/packages/webawesome/docs/_includes/sidebar.njk
@@ -399,3 +399,16 @@
     <li><a href="/docs/tokens/component-groups/">Component Groups</a></li>
   </ul>
 </wa-details>
+
+<wa-divider style="--spacing: var(--wa-space-xl);"></wa-divider>
+
+<div class="wa-stack wa-gap-xs" id="colophon">
+  {% include "logo-simple.njk" %}
+  <div class="wa-cluster wa-gap-xs">
+    <h2 class="wa-heading-xs">Web Awesome</h2>
+    <wa-badge id="version-beta-badge" variant="orange" appearance="filled" style="font-size: var(--wa-font-size-2xs);">Beta</wa-badge>
+    <wa-tooltip for="version-beta-badge" distance="2" style="font-size: var(--wa-font-size-xs);">Here be freshly made Awesome&hellip; and possible dragons</wa-tooltip>
+  </div>
+  <p class="wa-caption-s">Version {{ package.version }}</p>
+  <p class="wa-caption-s">&copy; Fonticons, Inc.</p>
+</div>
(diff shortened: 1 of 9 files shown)


packages/webawesome/docs/_includes/sidebar.njk
```diff
@@ -399,3 +395,16 @@
...
 </wa-details>
+
+<wa-divider style="--spacing: var(--wa-space-xl);"></wa-divider>
+
+<div class="wa-stack wa-gap-xs" id="colophon">
+  {% include "logo-simple.njk" %}
+  <div class="wa-cluster wa-gap-xs">
+    <h2 class="wa-heading-xs">Web Awesome</h2>
+    <wa-badge id="version-beta-badge" variant="orange" appearance="filled" style="font-size: var(--wa-font-size-2xs);">Beta</wa-badge>
+    <wa-tooltip for="version-beta-badge" distance="2" style="font-size: var(--wa-font-size-xs);">Here be freshly made Awesome&hellip; and possible dragons</wa-tooltip>
```
claviska, line 406:
```suggestion
    <wa-tooltip for="version-beta-badge" distance="2" style="font-size: var(--wa-font-size-xs);">Here be freshly made Awesome&hellip;and possible dragons</wa-tooltip>
```

claviska: I love this! What will we replace it with next month when we drop the beta label? 🤔

claviska (approved):
CleanShot 2025-09-24 at 14 00 32@2x

This is great. I'm totally fine with the brush here, but I wonder if we should use something like a 2x2-rounded-corners-with-colors-icon to draw a bit more attention to it since that's a pretty huge feature. (Warning: this might look like garbage; proceed with caution!)

claviska:
Oh, I also want to verify that the avatar/login link will show at the top right in the app on docs pages.

talbs:
Thanks, @claviska. Here are a couple of screenies from my local app showing the avatar/login bit...

image

image

packages/webawesome/docs/_includes/sidebar.njk
talbs: @claviska, hmm... a `<wa-khaleesi>` component?

talbs: On a more serious note - I think we can do some celebrating (and pointing to the changelog or official announcement from here).

lindsaym-fa (approved):

PR #1503, opened 2025-09-26
Make shadow visible in dark theme (#1493)

### LINKED ISSUE #1493: Shadow in dark theme is invisible
The shadow color is defined as
```
--wa-color-shadow:  color-mix(in oklab, var(--wa-color-surface-lowered) calc(var(--wa-shadow-blur-scale) * 32% + 40%), transparent);
--wa-color-surface-lowered: 
 color-mix(in oklab, var(--wa-color-surface-default), black 20%);
```

In dark theme surface lowered should mix with white not black, otherwise it's black on black

hhiruko:
Updated `--wa-color-surface-lowered` variable mix color in dark theme from **black** to **white** to improve color contrast. 
Since the `--wa-color-shadow` variable mixes with `--wa-color-surface-lowered`, it makes shadow color visible in dark theme.

Changes were made in the following themes:
- Default
- Awesome
- Shoelace

Fixes #1493

### DIFF
--- a/packages/webawesome/src/styles/themes/awesome.css
+++ b/packages/webawesome/src/styles/themes/awesome.css
@@ -102,7 +102,7 @@
 
     --wa-color-surface-raised: var(--wa-color-neutral-10);
     --wa-color-surface-default: var(--wa-color-neutral-05);
-    --wa-color-surface-lowered: color-mix(in oklab, var(--wa-color-surface-default), black 20%);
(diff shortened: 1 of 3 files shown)


lindsaym-fa:
Thanks for your PR, @hhiruko! We don't plan to make shadows lighter as suggested in #1493, and we want to keep the default value for `--wa-color-surface-lowered` darker than `-default` and `-raised` because this is used for background colors for elements with low elevation, not just as a basis for shadow colors.

PR #1518, opened 2025-09-29
Fix 11ty / server handoff for Brian

KonnorRogers:


talbs (approved):
Works as intended! Thanks for all of the fixin'

PR #1519, opened 2025-09-30
Remove unsupported tag ‘plain’ appearance visual test

roydukkey:
Tag does not have a plain appearance.


claviska (approved):
Nice catch. Thanks for the PR!

PR #1526, opened 2025-09-30
Round `--content-width-*` and use in the docs

lindsaym-fa:
- Updates our `--content-width-*` properties to use `round()` so that resulting values are always whole pixels to prevent the odd rendering bug like we previously had for `<wa-scroller>` (see https://github.com/shoelace-style/webawesome/pull/907#pullrequestreview-2841640899 and #1525 for context)
- Use `--content-width-*` and `--line-length-*` for our docs pages

### DIFF
--- a/packages/webawesome/docs/assets/styles/utils.css
+++ b/packages/webawesome/docs/assets/styles/utils.css
@@ -4,11 +4,12 @@
 
 @layer wa-utils {
   :root {
-    /* max-width for fixed width app-based pages */
-    --content-width-l: 132ch;
-    --content-width-m: 110ch;
-    --content-width-s: 80ch;
-    --content-width-xs: 50ch;
+    /* max-width for fixed width app-based pages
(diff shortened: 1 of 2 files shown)


packages/webawesome/docs/assets/styles/utils.css
```diff
@@ -4,11 +4,12 @@
 
 @layer wa-utils {
   :root {
-    /* max-width for fixed width app-based pages */
-    --content-width-l: 132ch;
-    --content-width-m: 110ch;
-    --content-width-s: 80ch;
-    --content-width-xs: 50ch;
+    /* max-width for fixed width app-based pages
+     * We round() these to avoid obscure component rendering bugs caused by widths with fractional pixels */
```
talbs, line 8: ✨

talbs (approved):
lookin-good-gettin-presents

---

Nice sync up and thanks for the rounding fix/enhancement!

PR #1533, opened 2025-10-01
Fix missing form-control-label part on texarea

zedix:
The `part="form-control-label"` is available on `wa-input`, `wa-select`, `wa-radio-group`, `wa-color-picker` but is missing on `wa-textarea`.

This MR fixes the consistency, and makes it easier to apply global custom styles, for example:

```css
::part(form-control-label) {
  /* advanced styling for block labels */
}

:is(wa-switch, wa-checkbox, wa-radio)::part(label) {
   /* advanced styling for inline labels */
}
```


claviska (approved):
Thanks for catching this!

"Consistency is the key to success." — Kevin

PR #1536, opened 2025-10-02
Fix wa-button tap highlight

wendevlin:
Android Chrome shows a tap highlight, when you tap a button.
This PR fixes the shape of this highlight by adding the border radius to the host and hide the overflow.

before:
grafik

after:
grafik


wendevlin:
Unfortunately the `overflow: hidden` breaks the focus-visible outline.

PR #1539, opened 2025-10-03
Docs: Add Planned Features to Navigation

talbs:
This work adds planned features to the documentation navigation with visual badges and styling. This enhancement helps users identify which components are planned but not yet implemented, linking to relevant GitHub issues for more information.

* Adds specific planned features (Charts, Combobox, Data Grid, Datepicker) to the sidebar with appropriate styling
* Introduces a new "planned" badge component for marking planned features

This work also tackles a few minor Sidebar UI improvements...
* Adds the opposite of `.icon-embiggen` -> `.icon-shrink` to contextually reduce the visual size of icons slightly. 
* Applies `.icon-shrink` to experimental feature iconography in the docs Sidebar
* ~Updates the pro badge component to support a shrink parameter for a more appropriately scaled size in some places (DRYing out cases like https://github.com/shoelace-style/webawesome-app/pull/131#discussion_r2392453502).~

---

| Preview |
|--------|
| CleanShot 2025-10-02 at 21 15 58@2x |
| Planned Features don't have links on their label, but have a "planned" badge linking to the GH Issue |

### DIFF
--- a/packages/webawesome/docs/_includes/pro-badge.njk
+++ b/packages/webawesome/docs/_includes/pro-badge.njk
@@ -1,5 +1,12 @@
-{% macro proBadge(proBadgeDescription, proBadgeId) %}
-  {% set badgeId = proBadgeId or ("pro-badge-" + ("" | uniqueId(8))) %}
-  <wa-badge appearance="accent" pill class="pro" id="{{ badgeId }}">Pro</wa-badge>
-  <wa-tooltip for="{{ badgeId }}">{{ proBadgeDescription or "This requires access to Web Awesome Pro" }}</wa-tooltip>
+{% macro proBadge(params) %}
+  {% set description = params.description or "This requires access to Web Awesome Pro" %}
+  {% set badgeId = params.id or ("pro-badge-" + ("" | uniqueId(8))) %}
+  {% set shrink = params.shrink or false %}
+  {% if shrink %}
+    {% set classes = "pro shrink" %}
+  {% else %}
+    {% set classes = "pro" %}
+  {% endif %}
+  <wa-badge appearance="accent" pill class="{{ classes }}" id="{{ badgeId }}">Pro</wa-badge>
+  <wa-tooltip for="{{ badgeId }}">{{ description }}</wa-tooltip>
 {% endmacro %}
--- a/packages/webawesome/docs/_includes/pro-badge.njk
+++ b/packages/webawesome/docs/_includes/pro-badge.njk
@@ -1,12 +1,6 @@
 {% macro proBadge(params) %}
   {% set description = params.description or "This requires access to Web Awesome Pro" %}
   {% set badgeId = params.id or ("pro-badge-" + ("" | uniqueId(8))) %}
-  {% set shrink = params.shrink or false %}
-  {% if shrink %}
-    {% set classes = "pro shrink" %}
-  {% else %}
-    {% set classes = "pro" %}
-  {% endif %}
-  <wa-badge appearance="accent" pill class="{{ classes }}" id="{{ badgeId }}">Pro</wa-badge>
+  <wa-badge appearance="accent" pill class="pro" id="{{ badgeId }}">Pro</wa-badge>
   <wa-tooltip for="{{ badgeId }}">{{ description }}</wa-tooltip>
 {% endmacro %}
--- a/packages/webawesome/docs/assets/styles/docs.css
+++ b/packages/webawesome/docs/assets/styles/docs.css
@@ -218,6 +218,11 @@ wa-button.delete {
   }
 }
 
+/* planned sidebar item */
+.is-planned {
+  color: var(--wa-color-neutral-50);
+}
+
 #sidebar-close-button {
   display: none;
 }
--- a/packages/webawesome/docs/assets/styles/docs.css
+++ b/packages/webawesome/docs/assets/styles/docs.css
@@ -220,7 +220,7 @@ wa-button.delete {
 
 /* planned sidebar item */
 .is-planned {
-  color: var(--wa-color-neutral-50);
+  color: var(--wa-color-text-quiet);
 }
 
 #sidebar-close-button {
--- a/packages/webawesome/docs/assets/styles/docs.css
+++ b/packages/webawesome/docs/assets/styles/docs.css
@@ -181,6 +181,12 @@ wa-page > header {
     }
   }
 
+  li wa-badge {
+    /* adding 1 scale below --wa-font-size-2xs to handle badge proportions in sidebar nav */
+    --font-size-3xs: round(calc(var(--wa-font-size-2xs) / 1.125), 1px); /* 10px */
+    font-size: var(--font-size-3xs);
+  }
+
   wa-details {
     --show-duration: 0;
     --hide-duration: 0;
--- a/packages/webawesome/docs/assets/styles/utils.css
+++ b/packages/webawesome/docs/assets/styles/utils.css
@@ -88,6 +88,7 @@
   /* planned badge */
   wa-badge.planned {
     background-color: var(--wa-color-neutral-fill-quiet);
+    color: var(--wa-color-neutral-on-quiet);
 
     + wa-tooltip {
       font-size: var(--wa-font-size-xs);
--- a/packages/webawesome/docs/assets/styles/utils.css
+++ b/packages/webawesome/docs/assets/styles/utils.css
@@ -79,10 +79,6 @@
       font-size: var(--wa-font-size-xs);
       --max-width: unset;
     }
-
-    &.shrink {
-      font-size: round(calc(var(--wa-font-size-2xs) / 1.125), 1px); /* ~10px */
-    }
   }
 
   /* planned badge */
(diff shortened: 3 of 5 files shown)


packages/webawesome/docs/_includes/pro-badge.njk
```diff
@@ -1,5 +1,12 @@
-{% macro proBadge(proBadgeDescription, proBadgeId) %}
-  {% set badgeId = proBadgeId or ("pro-badge-" + ("" | uniqueId(8))) %}
-  <wa-badge appearance="accent" pill class="pro" id="{{ badgeId }}">Pro</wa-badge>
-  <wa-tooltip for="{{ badgeId }}">{{ proBadgeDescription or "This requires access to Web Awesome Pro" }}</wa-tooltip>
+{% macro proBadge(params) %}
```
talbs, line 1: I'll make quick PRs to update `webawesome-pro` and `webawesome-app` instances of this macro to use this new param syntax.

packages/webawesome/docs/assets/styles/utils.css
```diff
@@ -79,6 +79,20 @@
...
+    &.shrink {
+      font-size: round(calc(var(--wa-font-size-2xs) / 1.125), 1px); /* ~10px */
+    }
+  }
+
+  /* planned badge */
+  wa-badge.planned {
+    background-color: var(--wa-color-neutral-fill-quiet);
+
+    + wa-tooltip {
```
talbs, line 88: We may want to think about making a more general rule for tooltip `font-size` on badges, but I thought that blanket call was a bit premature with this being the second instance. Letting it ride like we're in WAs Vegas!

packages/webawesome/docs/_includes/pro-badge.njk
```diff
@@ -1,5 +1,12 @@
...
+  {% set description = params.description or "This requires access to Web Awesome Pro" %}
+  {% set badgeId = params.id or ("pro-badge-" + ("" | uniqueId(8))) %}
+  {% set shrink = params.shrink or false %}
+  {% if shrink %}
```
talbs: This is long windeded and feels old English-style, but the ternary operator syntax wasn't working. 

☝️ _Dost thou feel any discomfort therefrom?_

packages/webawesome/docs/_includes/pro-badge.njk
talbs: Draft PR created - https://github.com/shoelace-style/webawesome-app/pull/138.

packages/webawesome/docs/assets/styles/utils.css
```diff
@@ -79,6 +79,20 @@
...
     }
+
+    &.shrink {
+      font-size: round(calc(var(--wa-font-size-2xs) / 1.125), 1px); /* ~10px */
+    }
+  }
+
+  /* planned badge */
+  wa-badge.planned {
+    background-color: var(--wa-color-neutral-fill-quiet);
```
lindsaym-fa, line 85:
```suggestion
    background-color: var(--wa-color-neutral-fill-quiet);
    color: var(--wa-color-neutral-on-quiet);
```

Just to make sure that we're hitting appropriate contrast ratios. We can't _necessarily_ guarantee that themes will be designed so that `var(--wa-color-*-on-normal)` (the default badge text) works with `various(--wa-color-*-fill-quiet)`, but we can use the intended pairing to be extra safe.

packages/webawesome/docs/assets/styles/docs.css
```diff
@@ -218,6 +218,11 @@ wa-button.delete {
   }
 }
 
+/* planned sidebar item */
+.is-planned {
+  color: var(--wa-color-neutral-50);
```
lindsaym-fa: I'm torn here... `var(--wa-color-neutral-50)` doesn't offer compliant contrast in all situations — e.g., in dark mode for all existing themes and in light mode for Mellow, whose `surface-default` isn't white.

My best suggestion is to use `var(--wa-color-text-quiet)` for this, but I agree that it doesn't quite look subtle enough for most light mode themes...

packages/webawesome/docs/_includes/pro-badge.njk
lindsaym-fa: My primary discomfort is that I'm not sure `.shrink` is doing what we expect it to in the sidebar...

In `docs.css`, we have this:
```css
#sidebar,
#outline {
  ...
  li wa-badge::part(base) {
    font-size: var(--wa-font-size-2xs);
    font-weight: var(--wa-font-weight-bold);
  }
}
```

Because we're targeting `::part(base)`, it supersedes any font size we set on `wa-badge` itself. So, in the sidebar, `.shrink` only actually affects the padding of the badge, not the content's font size.

Secondary discomfort: We're not applying `.shrink` to the Planned badges, so Pro and Planned have different footprints.

My gut feelings: 

- We should retire the above ☝️ styles from `docs.css`.
- `.shrink` feels like a characteristic of badges in a certain context rather than of the specific type of badge. Rather than having this as a parameter for Pro badges (and adding the same parameter to Planned badges), should we instead scope the shrink styles to all badges alongside links in the sidebar?

Not sure what the right answer is. Just 🍟🍔🥤 for thought.

lindsaym-fa (commented):
Wonderful! Some thoughts and ramblings to consider.

packages/webawesome/docs/assets/styles/docs.css
talbs: Yeah, I reached for `var(--wa-color-text-quiet)` initially, but had the same reaction.

But if that's what we have to use for compliance, I'll stick with it.

packages/webawesome/docs/_includes/pro-badge.njk
talbs: The shrink class (and CSS that I added prior to DRYing things out) was in reaction to the size of the Pro badge specifically since we `text-transform: uppercase;` that badge's text.

I think I'll try 2 things...

1. I'll try removing those styles from `docs.css` and sizing `wa-badges` as our docs intend by setting a `font-size` on the badge.
1. I'll remove the `text-transform` from the Pro badges - I think having the size and capitalization sync up across badges is a bigger win than losing any minimal emphasis the ALLCAPS treatment gets us.

lindsaym-fa: Sounds good to me!

talbs: Okay, I've pushed a couple of commits (https://github.com/shoelace-style/webawesome/pull/1539/commits/96f3abf1f472fb8b73ec4e9695f27020c090cc05 and https://github.com/shoelace-style/webawesome/pull/1539/commits/74226f4c7e3312a3212c900488342c7d4a2bcacb). Here's where we are...

* Pro Badges are no longer titlecased.
* All `wa-badges` are now rendering at a ~10px `font-size` (extending the `--wa-font-size-` scale one manually for this case).

CleanShot 2025-10-03 at 11 42 24@2x

talbs:
Okay, @lindsaym-fa. Thanks for the review - I think I addressed all of your very useful feedback. Mind taking another spin through?

packages/webawesome/docs/assets/styles/docs.css
```diff
@@ -181,6 +181,12 @@ wa-page > header {
     }
   }
 
+  li wa-badge {
+    /* adding 1 scale below --wa-font-size-2xs to handle badge proportions in sidebar nav */
+    --font-size-3xs: round(calc(var(--wa-font-size-2xs) / 1.125), 1px); /* 10px */
```
lindsaym-fa, line 186: Out-of-scope for this, but I think this might be worth adding to our font scale in themes.

I'll give that a quick think, but opinions welcome!

lindsaym-fa (approved):
Lookin' good! ✨

packages/webawesome/docs/assets/styles/docs.css
talbs: Yeah, I had the same thought, but figured I'd unofficially extend it for us first. Also, decided to not `--wa-` prefix because it was not wa-fficial. That seemed like a good practice to follow.

talbs: FWIW, it may be useful - but does creating a `font-size` that small (~10px in the default theme + browser default sizing) warrant any a11y/legibility concerns we need to address in docs (i.e. warning: this is a sharp knife - know when and how to use it, chef)?

claviska (approved):
This looks great! I realize the "planned" badges are clickable but is that discoverable enough? Should the component name also be clickable? 🤔 

Not blocking, of course!

packages/webawesome/docs/assets/styles/docs.css
lindsaym-fa: Unofficially extending it for our use is totally the right call.

There isn't anything explicitly non-conformant about small text sizes as long as the text can scale with typical browser and OS zoom features, which the default values in our themes allow for.

talbs:
> I realize the "planned" badges are clickable but is that discoverable enough? Should the component name also be clickable? 🤔

Yeah, that was "by design".

I figured if a feature wasn't ready yet, we should reflect that by having the label in the Sidebar be unlinked/plain text (along with a more slightly subtle color). To me, the "planned" tag drew enough attention to the status of things and if folks wanted more details, making that badge the link felt more semantically correct - since we're really linking to the project management/planning efforts for this.

I'm a 5/10 on this, so if you feel more strongly, I'm happy to link the entire visual line.

What say ye, @claviska?

claviska:
No concerns here :)

PR #1544, opened 2025-10-03
Add link to Figma Design Kit doc

lindsaym-fa:
Relies on https://github.com/shoelace-style/webawesome-pro/pull/58
- Adds a link to new Figma Design Kit doc to the sidebar
- Adds common styles for step icons to outline ordered instructions

---

@talbs, mind letting me know your thoughts on:
1. Where the Figma kit belongs in the sidebar
2. Whether or not the `.step-icon` styles belong in `utils.css`?

I figure `.step-icon` would be good to use in the 2FA flow too, which I can make a quick PR to update if everything's kosher.


packages/webawesome/docs/assets/styles/utils.css
```diff
@@ -99,6 +99,15 @@
     line-height: var(--wa-form-control-value-line-height);
     padding-inline: 0.33em;
   }
+
+  /* step icons for ordered instructions */
+  .step-icon {
+    --primary-color: var(--wa-color-neutral-on-normal);
+    --secondary-color: var(--wa-color-neutral-fill-normal);
+    --secondary-opacity: 1;
+    font-size: var(--wa-font-size-2xl);
```
talbs:
```suggestion
    font-size: var(--font-size, var(--wa-font-size-2xl));
```

NIT: Can we set a default font-size here using a custom prop so that it can be overriden depending on context? I could see us using this icon in various heading levels (with it scaling to match).

talbs: Thinking about the idea of this icon scaling based on context, should this rule set a `font-size`?

If we wanted to override the `inherit` behavior with a specific size, that seems to be a situational call best handled by custom or inline styling, no?

packages/webawesome/docs/_includes/sidebar.njk
```diff
@@ -17,6 +17,12 @@
 <ul>
   <li><a href="https://github.com/shoelace-style/webawesome/discussions" target="_blank">Help &amp; Support</a></li>
   <li><a href="https://github.com/shoelace-style/webawesome/">Source Code</a></li>
+  <li>
```
talbs, line 21: Hmmm, the Resources area seems like the best current section - but it is getting a bit "junk drawer"-like. 

This placement is okay for now, but a larger IA revamp will hopefully find a good future home.

talbs: I would also suggest placing references/links to this in some `webawesome-app` views.

* Should we mention (and link to it from the Plan Card on /teams)?
* Should we also make an assets  or downloads area within a Team view (since Pro-access is tied to the teams' plan)? I could see us including a link to the Web Awesome Pro ZIP there too FWIW.

talbs (commented):
Nice work adding this. 

I left some thoughts on both where to link and the `step-icon` utility inline. Let me know what you think on each and if you want to chat anything out.

packages/webawesome/docs/_includes/sidebar.njk
lindsaym-fa: Wholeheartedly agree on both points. 

I'll see if I can make time to add links to the app before launch, but even if not, I created https://github.com/shoelace-style/webawesome-app/issues/153 to track it.

packages/webawesome/docs/assets/styles/utils.css
lindsaym-fa: Perhaps the best of both worlds: I updated `.step-icon` to use a relative size (`font-size: 2em;`). This way, they're larger than the text by default (which so far has been the preference in 2FA and here in the Figma doc), but it adapts to the surrounding text and we can still override as the situation demands. Does that sit right with you?

packages/webawesome/docs/assets/styles/utils.css
```diff
@@ -109,6 +109,15 @@
     line-height: var(--wa-form-control-value-line-height);
     padding-inline: 0.33em;
   }
+
+  /* step icons for ordered instructions */
+  .step-icon {
```
talbs, line 114: 📶 💪

packages/webawesome/docs/assets/styles/docs.css
```diff
@@ -386,6 +386,27 @@ h1.title {
   }
 }
 
+/* Images & Figures */
```
talbs, line 389: 💥

packages/webawesome/docs/_layouts/component.njk
```diff
@@ -297,11 +297,11 @@
   <div class="component-help">
     <strong>Need a hand?</strong>
     <wa-button size="small" appearance="filled" variant="neutral" href="https://github.com/shoelace-style/webawesome/issues" target="_blank">
-      <wa-icon slot="prefix" name="bug"></wa-icon>
+      <wa-icon slot="start" variant="regular" name="bug"></wa-icon>
```
talbs, line 300: Nice fix!

packages/webawesome/docs/_layouts/component.njk
```diff
@@ -297,11 +297,11 @@
...
       Report a bug
     </wa-button>
     <wa-button size="small" appearance="filled" variant="neutral" href="https://github.com/shoelace-style/webawesome/discussions" target="_blank">
-      <wa-icon slot="prefix" name="message-question"></wa-icon>
+      <wa-icon slot="start" variant="regular" name="message-question"></wa-icon>
```
talbs, line 304: Equally nice fix!

talbs (approved):
Great success! Get on the road and 🚚 this to Mergeburgh!

PR #1545, opened 2025-10-03
Optimize theme builder font data

lindsaym-fa:
Currently, when selecting a single, custom font in the theme builder, all of the paired fonts from the originating theme are included in the request.

So, if we want e.g. Quicksand, the theme builder output also includes Crimson Pro:
```
@import url('https://fonts.bunny.net/css2?family=Crimson+Pro:ital,wght@0,200..900;1,200..900&family=Quicksand:wght@300..700&display=swap');
```

This change separates the URLs for individual fonts in the schema so that only the selected font is included. Also, I've removed the unsupported `opsz` parameter that prevents fonts from loading (see also: https://github.com/shoelace-style/webawesome-pro/pull/59).

We can probably improve performance further by concatenating all selected fonts in a single request. For now, though, at least we're not importing more fonts than necessary.


claviska (approved):
Nice catch! Thanks for cleaning this up.

PR #1547, opened 2025-10-04
fix broken checkboxes

KonnorRogers:
Fixes fundamentally broken checkbox behavior that was resetting `value` when checked -> unchecked.


claviska:
Does this also affect `<wa-switch>`?

packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -16,6 +16,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 - Fixed a bug in `<wa-button>` where slotted badges weren't properly positioned in buttons with an `href` [issue:1377]
 - Fixed focus outline styles in `<wa-details>` and native `<details>` [issue:1456]
 - Fixed focus outline styles in `<wa-scroller>`, `<wa-dialog>`, and `<wa-drawer>` [issue:1484]
+- Fixed a bug in `<wa-checkbox>` where it's value would revert to `""` when checked / unchecked [pr:1547]
```
randomguy-2650: Should be “its” instead of “it‘s”.

```suggestion
- Fixed a bug in `<wa-checkbox>` where its value would revert to `""` when checked / unchecked [pr:1547]
```

randomguy-2650 (changes requested):
I’ve left a comment on a grammar error in the changelog.

KonnorRogers:
> Does this also affect `<wa-switch>`?

Possibly. I'd have to double check `<wa-switch>`.

claviska (approved):

PR #1586, opened 2025-10-14
Replace "filled outlined" with "filled-outlined"

### LINKED ISSUE #1127: Update appearances to be type safe

claviska:
Fixes #1127

@lindsaym-fa Note how this replaces `~=` with `=` in appearance selectors in the following elements:

- Badge
- Button
- Callout
- Details
- Input
- Select
- Tag
- Textarea

I also added more appearance examples to those that were lacking.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14017,8 +14017,7 @@
         "qr-creator": "^1.0.0"
       },
       "devDependencies": {
-        "@wc-toolkit/jsx-types": "^1.3.0",
-        "eleventy-plugin-git-commit-date": "^0.1.3"
+        "@wc-toolkit/jsx-types": "^1.3.0"
       },
       "engines": {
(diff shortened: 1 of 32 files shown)


lindsaym-fa (approved):
Added a couple of extra commits to change labels from "filled + outlined" to "filled-outlined".

Otherwise, 🔥.

PR #1587, opened 2025-10-14
Focus dialog and drawer when opening

### LINKED ISSUE #1302: Drawer and Dialog not gaining focus correctly
### Describe the bug
When a screen reader user presses a button to open a popup they should be brought directly to the content of the dialog but when not using the tabindex attribute this does not happen correctly. It seems to focus the dialog but you have to press enter again or tab if there is a focusable element in the dialog to actually enter it.

### To Reproduce

1. Go to https://webawesome.com/docs/components/drawer/
2. Turn on NVDA or another screen reader
3. Click on Open Drawer under the "Drawer" h1
4. Notice that the screen reader just says dialog
5. Try to navigate using the arrow keys
6. Notice none of the content is read
7. Press enter
8. Now you can navigate around the dialog

### Browser / OS
 - OS: Windows
 - Browser: Firefox
 - Browser version: 141.0.3

### Additional information
This is not the expected behavior. See:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog
All of the examples bring the user directly to the dialog content.

claviska:
Fixes #1302

- [x] Using the keyboard now ensures the dialog/drawer receives focus after opening
- [x] VoiceOver is correctly announcing when dialogs/drawers open


lindsaym-fa (approved):
✨

PR #1589, opened 2025-10-14
Prevent tab group body truncation

### LINKED ISSUE #1401: Tab Group overflow clipped
### Describe the bug
Overflowing content within the Tab group is being clipped.

### To Reproduce
Use the Slider component inside of the Tab Panel in a Tab Group.

### Demo
https://codepen.io/roydukkey/pen/gbaZGgd?editors=1000

### Browser / OS
 - OS: Mac
 - Browser: Chrome
 - Browser version: 140.0.7339.81

### Additional information

Wonder if simply removing this rule will correct the behavior. Though, I'm not sure if the rule was added to support other behavior.

https://github.com/shoelace-style/webawesome/blob/0e56ed0cbb2901f47c2dcd652c1feca7f49d3d15/packages/webawesome/src/components/tab-group/tab-group.css#L30

claviska:
Fixes #1401 as discussed


lindsaym-fa (approved):
🚀

PR #1615, opened 2025-10-16
component(callout): Update `@since` JSDoc tag from 2.0 to 3.0

randomguy-2650:
#### Description

Web Awesome introduced the Callout component, as an effort to separate the “all‐in‐one” Alert component in Shoelace. In Shoelace, you cannot really find the Callout component, and in <https://shoelace.style/components/alert#variants>, the example is basically identical to <https://webawesome.com/docs/components/callout/#variants>, so it probably was copied over.

To prove this even more (if even necessary at this point), <https://shoelace.style/components/callout> points to a 404, so a Callout page wasn’t even hidden in Shoelace, it just never existed.


claviska:
Great detailed catch! Thank you!

claviska (approved):

Issue #1446, opened 2025-09-14
React - Definitions are incomplete


GedMarc:
### Describe the bug
React components don't specify the allowed attributes, events, and other properties accordingly to the webawesome documentation site 

### To Reproduce

1. Open any component, worst affected seems to be wa-badge
2. Verify the attributes, css parts and etc  from the documentation against the react file
3. Lots of misses in the implementation (e.g. badge only specifies one css part, doesn't specify any attributes like variant, pill, etc or anything else on the webawesome documentation)

All react components affected

jake-danton:
When I follow the setup to use web components in React 19, I don't even get the events like "onClick" defined. Though they work just fine if you ignore the Typescript error.

GedMarc:
Making good progress. @claviska do we need to add tests for this? I can't seem to find where that is hmm, adding it though I think is quite a bit chunkier

GedMarc:
In time for 1.0.0?

GedMarc:
# WebAwesome React Types Parity Checklist

This checklist is to verify that the React type definitions for each WebAwesome component contain all (and possibly more) of the definitions we have in our source of truth for the web components. 
This is React‑specific

- Scope: React type definitions under `react/` only
- Goal: Ensure React types are a superset or, at minimum, a complete mirror of public APIs exposed by the underlying web components
- Audience: Maintainers doing a parity pass after React types update
- Date: 2025-10-17

Legend
- [ ] Not verified yet
- [x] Verified / OK
- [!] Action required (gap or mismatch)

Global checks (apply to all components)
- [x] All React wrappers are exported from `react/index.d.ts`
- [x] Each wrapper forwards refs to the correct web component instance type (ForwardRefExoticComponent with InstanceType of underlying component)
- [x] React props include all public web component attributes/properties
- [x] React props include event handlers for all public custom events (names and payload types)
- [x] React props include standard React HTML attributes without collisions (via Omit<...>)
- [x] Slot content is representable in React (via children or named slot patterns)
- [x] All public methods on the element instance are callable through the ref type
- [x] All CSS parts and CSS custom properties are documented in JSDoc of the React wrapper types (where applicable)
- [x] Default values in JSDoc match the web component defaults
- [x] Deprecated/alias props and events are marked consistently
- [x] Controlled/uncontrolled patterns are documented (e.g., value vs defaultValue where relevant)
- [x] Form-associated components expose required props (name, disabled, required, form, value) consistently
[...]

PR #1619, opened 2025-10-17
React Definitions - Resolves for #1446

GedMarc:
Update react builders to include the wa- events, css properties, and variants for react

- Enhance React wrappers to include ergonomic DOM event bindings alongside custom events.
- Implement cleanup of stale `@cssproperty` tags from JSDocs to avoid redundancy.
- Add support for auto-expanding known CSS variable scale families in CEM metadata.
- Enable CSS data inclusion for JetBrains plugin integration.

Tests not included

Applied an AI checklist to ensure no components outstanding and all fields present after updates. Posted on issue thread #1446

### DIFF
--- a/packages/webawesome/custom-elements-manifest.js
+++ b/packages/webawesome/custom-elements-manifest.js
@@ -153,6 +153,138 @@ export default {
       },
     },
 
+    {
+      name: 'wa-css-vars',
+      packageLinkPhase({ customElementsManifest }) {
+        const root = __dirname;
+
+        // Known token scales we want to surface when any member is used
+        const SPACE_SCALES = ['2xs', 'xs', 's', 'm', 'l', 'xl', '2xl'];
+        const FONT_SIZE_SCALES = ['2xs', 'xs', 's', 'm', 'l', 'xl', '2xl'];
+        const BORDER_RADIUS_SCALES = ['s', 'm', 'l', 'xl', '2xl', 'pill', 'circle'];
+        const BORDER_WIDTH_SCALES = ['s', 'm', 'l'];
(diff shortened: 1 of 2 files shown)


GedMarc:
@jake-denton This is the generated react results, possible to spot check?

[react.zip](https://github.com/user-attachments/files/22968562/react.zip)

claviska:
I'm confused. We should be getting standard events and such for free via [@lit/react](https://lit.dev/docs/frameworks/react/). Everything else comes from the CEM. Does Shoelace exhibit the same problem?

There may also be confusion between _how_ these are being consumed. With React 19, you don't need the wrappers. Before React 19, you generally do need the wrappers. I'm not sure which method @jake-danton was using [in this comment](https://github.com/shoelace-style/webawesome/issues/1446#issuecomment-3310287940).

Are we trying to fix the React wrappers or raw types for the custom elements themselves?

marcomuser:
From my perspective it would make sense to double-down on the native React 19+ support as that streamlines integration across frameworks and is clearly the way forward. The support for custom elements is pretty decent now in React 19 and since that has been stable for nearly a year now, I think efforts should be focused around that.

Currently, it might be a bit surprising that there are several ways to consume Web Awesome components from React. Every component's docs  has a dedicated React tab in the `Importing` section which implies this is the way to go for React. However, it really should just be recommended for React versions < 19. Would it maybe just be easier to only support React 19+ to have a clean cutoff point?

claviska:
> From my perspective it would make sense to double-down on the native React 19+ support as that streamlines integration across frameworks and is clearly the way forward.

I agree. However, we committed to supporting them in 3.0 so they'll be around for a while longer. But I want to reduce the maintenance burden to support React, not increase it. This PR adds a lot of code to get something we should be getting for free, so I'd prefer to isolate the actual problem and solve for that instead.

PR #1626, opened 2025-10-17
Improve step icon color visibility

lindsaym-fa:
The secondary color of our step icons was feeling a bit light. This change styles step icons with different neutral tints in light and dark mode for bespoke contrast.

@talbs please let me know if you have any concerns or other suggestions!


talbs (approved):
You're totally bespeaking my language with this bespoke approach. Makes sense. 

It's great to see an example of how to get surgical and define custom things for light/dark responsibly.

PR #1634, opened 2025-10-21
Utils: Add Cosmetic Backgrounds

talbs:
This PR adds two new cosmetic background utility classes to enhance visual design options in the documentation styles.

- Introduces a `.background-dot-grid` utility class that creates a subtle dot grid pattern using CSS radial gradients
- Adds a `.background-wa-pattern` utility class that overlays a repeating SVG pattern with configurable opacity and color

This supports the efforts in https://github.com/shoelace-style/webawesome-pro/pull/45.

---

@lindsaym-fa, I added these here in `utils.css` as we may use them across the various repos that make up `webawesome.com`. Let me know if you'd prefer a different place OR a different approach on either.


lindsaym-fa (approved):
🎵 _Isn’t she lovely, isn’t she wonderful_ 🎵

PR #1635, opened 2025-10-21
Themer integration

claviska:
Part 1 of the broader `themer-integration` collection.


lindsaym-fa (approved):
Approved! I made a stopgap change to the Themes doc to remove the "Create a Theme" button (since that link has been removed) and instead direct folks to log in and/or head to their teams to create a project.

There's plenty more to do to improve the experience on this page, but hopefully this is helpful for now!

PR #1636, opened 2025-10-22
Fix cancel event from within drawer or dialog triggering them to close

### LINKED ISSUE #1585: Event `cancel` triggering a `wa-hide` inside Drawer
### Describe the bug

Prior discussion here: https://github.com/shoelace-style/webawesome/discussions/1541

There is a case where an input of type file is triggering a `cancel` if dismissed. This is caught by drawer which then emits `wa-hide` without checking the source of the cancel event causing the drawer to close when dismissing a file input.

### To Reproduce

1. Have an `<input type='file'>` inside a `<wa-drawer>` element
2. Open the drawer
3. Click on the file input
4. Dismiss the file selector
5. See the drawer close as well.

### Demo

Image

The common checks for comparing the event source to see if the drawer caused it won't work since the drawer is first catching the cancel and then emitting the hide. Seems like the drawer should be the one checking this.

Image

### Browser / OS
 - OS: Mac
 - Browser: Brave
 - Browser version: Brave: 1.81.137, Chromium: 139.0.7258.158

Jeremy-Walton:
# Problem

Fixes #1585 

File inputs within dialog or drawer emit a cancel event when closed which bubbles up to the dialog where it is captured and then re-emitted as a `sl-hide` which causes the drawer or dialog to be closed even though the dialog was not the element emitting the event.

## Before

<table border="0">
 <tr>
    <td>
      demo1
    </td>
    <td>
      demo2
    </td>
 </tr>
</table>

## After

<table border="0">
 <tr>
    <td>
      demo3
    </td>
    <td>
      demo4
    </td>
 </tr>
</table>

### DIFF
--- a/packages/webawesome/src/components/dialog/dialog.test.ts
+++ b/packages/webawesome/src/components/dialog/dialog.test.ts
@@ -119,6 +119,19 @@ describe('<wa-dialog>', () => {
         expect(el.open).to.be.true;
       });
 
+      it('should not close when bubbled cancel event originates from within the drawer', async () => {
+        const el = await fixture<WaDialog>(html` <wa-dialog open><input type="file" /></wa-dialog> `);
+        const input = el.querySelector('input')!;
+
+        await clickOnElement(el); // Chromium wants the page to have been clicked prior to closing the dialog.
+        const cancelEvent = new Event('cancel', { bubbles: true });
+        input.dispatchEvent(cancelEvent);
+
+        await aTimeout(250);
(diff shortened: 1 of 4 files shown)


claviska (approved):
Thanks for the really well-formed PR! I'll add a changelog entry and get this in.

scriswell:
Great update! Thank you, @Jeremy-Walton. And thanks to @claviska for the review and getting this merged in.

PR #1641, opened 2025-10-23
Fix checked switch styles in Shoelace theme

### LINKED ISSUE #1487: `wa-switch` checked styles not updating with Shoelace theme applied
### Describe the bug
The `wa-switch` doesn't update it's styles when toggling.
This is only reproducible with the Shoelace theme. 

### To Reproduce

1. Go to https://webawesome.com/docs/components/switch/
2. Set the theme to Shoelace
3. Toggle the switches on the page

ER:
The switches should update to checked/unchecked styles

AR:
The switch thumb moves but the styles do not change

### Demo

Reproducible on https://webawesome.com/docs/components/switch/
See STR above.

### Screenshots
The initially unchecked switched in the checked state
Image

The initially checked switch in the unchecked state
Image

### Additional information
The `wa-switch` styles in [shoelace.css](https://github.com/shoelace-style/webawesome/blob/cf80b57a79c9e832b736f5d50f0cedb4dd0e6c4b/packages/webawesome/src/styles/themes/shoelace.css#L450C5-L472) differ from the other themes which causes part of the issue. 

The `checked` attribute on `wa-switch` does not have `reflect: true` which causes the `wa-switch` keep the styles it was initially created with because the `checked` attribute is never added/removed which causes the styles to never updated because the [selectors](https://github.com/shoelace-style/webawesome/blob/cf80b57a79c9e832b736f5d50f0cedb4dd0e6c4b/packages/webawesome/src/styles/themes/shoelace.css#L450C5-L472) assume that the `checked` attribute will change.

I think the fix is to add `reflect: true` to the property definition of the check attribute? But I could be wrong!

lindsaym-fa:
Closes #1487 

Fixes Shoelace theme custom `wa-switch` styles to target `:state(checked)` instead of `[checked]`

### DIFF
--- a/packages/webawesome/src/styles/themes/shoelace.css
+++ b/packages/webawesome/src/styles/themes/shoelace.css
@@ -457,7 +457,7 @@
         border-color: var(--wa-color-gray-50);
       }
 
-      &[checked]::part(control) {
+      &:state(checked)::part(control) {
         background-color: var(--wa-form-control-activated-color);
         border-color: var(--wa-form-control-activated-color);
       }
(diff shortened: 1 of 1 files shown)


claviska (approved):
🥾

PR #1644, opened 2025-10-23
Avoid errors on earlier CustomStateSet API

medmunds:
Protect WebAwesomeElement.customStates accessors from errors in browsers implementing the earlier revision of the CustomStateSet API.

Resolves #1633.


claviska (approved):
Thanks! Hopefully this clears things up for older browsers.

PR #1647, opened 2025-10-24
Abstract Code Highlighter Syntax Coloring

talbs:
This PR abstracts the code highlighter syntax coloring by introducing CSS custom properties (variables) with fallback values, making the color scheme more customizable and themeable.

**Key Changes:**
- Introduced CSS custom properties for code syntax highlighting colors with existing color values as fallbacks
- Separated comment, string, and literal token styles into distinct rulesets for better organization
- Added a dark mode background color custom property

---

@lindsaym-fa, mind reviewing this and confirming its a good idea?

### DIFF
--- a/packages/webawesome/docs/assets/styles/code-highlighter.css
+++ b/packages/webawesome/docs/assets/styles/code-highlighter.css
@@ -2,22 +2,25 @@
 pre[id*='code-block-'] {
   color-scheme: dark;
   color: white;
-  background-color: var(--wa-color-neutral-20);
+  background-color: var(--wa-code-background, var(--wa-color-neutral-20));
 
   /* Ensures a discernible background color in dark mode
    * Useful for themes that use gray-20 as --wa-color-surface-default  */
   .wa-dark & {
-    background-color: var(--wa-color-surface-lowered);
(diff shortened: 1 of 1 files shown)


packages/webawesome/docs/assets/styles/code-highlighter.css
```diff
@@ -2,22 +2,25 @@
 pre[id*='code-block-'] {
   color-scheme: dark;
   color: white;
-  background-color: var(--wa-color-neutral-20);
+  background-color: var(--wa-code-background, var(--wa-color-neutral-20));
```
talbs: One general question: Should these be given the `wa-` prefix, or should we wait and make them available without the prefix (assuming possible official code block component work would want to define officially prefixed values)?

lindsaym-fa: Given these custom props are scoped to a specific element, `wa-` isn't strictly necessary. We typically reserve that for globally scoped custom properties.

lindsaym-fa (approved):
Absolutely great idea. Approved, pending (admittedly optional) removal of the `wa-` prefix.

PR #1652, opened 2025-10-24
Use `--wa-font-weight-code` in native styles

lindsaym-fa:
Adds `font-weight: var(--wa-font-weight-code);` to:

- `<code>`
- `<samp>`
- `<var>`
- `<pre>`
- `<kbd>`


claviska (approved):

lindsaym-fa:
Duplicates #1653 (but that one has a changelog so it wins)

PR #1653, opened 2025-10-24
Use `--wa-font-weight-code` in native styles

claviska:
Testing since #1652 is strangely borked.


lindsaym-fa (approved):
Changelog is an elite move. We'll take this one.

PR #1654, opened 2025-10-24
Fix links for style utilities

CodingDoug:


claviska (approved):
Nice catch. Really appreciate the PR!

claviska:
I'm not sure why the build is tripping up on this. It's possibly a Prettier error that's getting swallowed. I created #1691 which looks identical but passes fine. Thanks again!

PR #1666, opened 2025-10-27
Fix: Home View's Page + Open Graph Titles

talbs:
This PR fixes title rendering for the home page by introducing smart formatting that avoids redundant site name duplication. When the page title matches the site name, it now displays only once instead of appearing as "Site Name | Site Name".


claviska (approved):
This looks fine to me. We _could_ end up with duplicate suffixes if folks do "First | Second" in front matter, but that's no different than a typo imo. :shipit:

PR #1670, opened 2025-10-28
Fix theme and palette option spacing

lindsaym-fa:
Fixes the spacing between options at `/docs/themes` and `/docs/color-palettes`.

Before
image

After
Screenshot 2025-10-27 at 11 01 24 PM


lindsaym-fa:
Snuffed by #1669 🥲

PR #1689, opened 2025-10-30
Remove unsupported property; closes #1677

### LINKED ISSUE #1677: Button group variants don't seem to be working
### Describe the bug
Variants, I believe, are supposed to change the appearance of a button or button group. It seems like for button groups, this functionality is broken

### To Reproduce

1. Go to "https://webawesome.com/docs/components/button-group/#theme-buttons"
2. Observe

claviska:
This property wasn't removed during an alpha cleanup. It didn't get highlighted because it's public. The styles also aren't used anywhere, so those have been removed as well.


talbs (approved):
lookin-good-gettin-presents

PR #1694, opened 2025-10-30
Fix `appearance="filled-outlined"` in `<wa-card>`

### LINKED ISSUE #1671: Docs - filled-outlined not recorded in changelog for wa-card
### Describe the bug
The list of impacted components for the breaking `filled-outlined` appearance option does not include `<wa-card>` as impacted.

Additionally, `filled-outlined` is not showing as an option in the attributes list for `<wa-card>` - only found as part of the examples above.

kelseythejackson:
Added `wa-card` to the changelog of the breaking change that came with the `filled-outlined` attribute and added it to the list of attributes.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -10,7 +10,7 @@ Web Awesome follows [Semantic Versioning](https://semver.org/). Breaking changes
 
 Components with the <wa-badge variant="warning">Experimental</wa-badge> badge should not be used in production. They are made available as release candidates for development and testing purposes. As such, changes to experimental components will not be subject to semantic versioning.
 
-### Next
+## Next
 
 - 🚨 BREAKING: Changed `appearance="filled outlined"` to `appearance="filled-outlined"` in the `wa-card` component [issue:1671]
 
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -16,6 +16,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
   - `<wa-badge>`
   - `<wa-button>`
   - `<wa-callout>`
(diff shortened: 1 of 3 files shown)


kelseythejackson:
This PR fixes the issue raised in #1671

packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -16,6 +16,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
   - `<wa-badge>`
   - `<wa-button>`
   - `<wa-callout>`
+  - `<wa-card>`
```
claviska, line 27: Looks like we missed this one before 3.0.0 went out. This change will need to go under a new section called `### Next` so it lands in the correct place when the next version goes out.

kelseythejackson: Ah, gotcha. Updated

packages/webawesome/src/components/card/card.css
lindsaym-fa:
```suggestion
:host([appearance='plain']) {
```

lindsaym-fa:
```suggestion
:host([appearance='outlined']) {
```

lindsaym-fa:
```suggestion
:host([appearance='filled']) {
```

lindsaym-fa (changes requested):
A couple of stray `~=` to clean up, otherwise this looks good!

packages/webawesome/src/components/card/card.css
kelseythejackson: updated

kelseythejackson: updated

kelseythejackson: updated

lindsaym-fa (dismissed):
All good!

lindsaym-fa (approved):

PR #1715, opened 2025-11-03
Control 11ty-Generated Headings' Display via Front Matter

talbs:
This PR adds functionality to hide 11ty-generated page titles through front matter configuration conditionally. When `hasGeneratedTitle` is set to `false` in a page's front matter, the generated title will be hidden using CSS.

- Added `hasGeneratedTitle` front matter variable with default value of `true`

---

@lindsaym-fa mind checking my choices here and making sure this is a better way to go than per-view custom CSS?

### DIFF
--- a/packages/webawesome/docs/_includes/base.njk
+++ b/packages/webawesome/docs/_includes/base.njk
@@ -28,7 +28,7 @@
   </head>
   <body class="layout-{{ layout | stripExtension }} page-{{ pageClass or page.fileSlug or 'home' }}{{ ' page-wide' if wide }}">
     {% if hasBanner == undefined %}{% set hasBanner = true %}{% endif %}
-    {% if hasHiddenGeneratedTitle == undefined %}{% set hasHiddenGeneratedTitle = false %}{% endif %}
+    {% if hasGeneratedTitle == undefined %}{% set hasGeneratedTitle = true %}{% endif %}
 
     {% set defaultWaPageAttributes = defaultWaPageAttributes or { view: 'desktop', 'disable-navigation-toggle': true, 'mobile-breakpoint': 1180, 'disable-sticky': 'banner' } %}
     {% set waPageAttributes = waPageAttributes or {} %}
@@ -119,7 +119,7 @@
         <div id="flashes">{% server "flashes" %}</div>
 
         {% block header %}
-          <h1 class="title title-generated{{ ' is-hidden' if hasHiddenGeneratedTitle }}">{{ title }}</h1>
+          <h1 class="title title-generated{{ ' hidden' if not hasGeneratedTitle }}">{{ title }}</h1>
         {% endblock %}
 
         {% block beforeContent %}{% endblock %}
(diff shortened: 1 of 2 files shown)


packages/webawesome/docs/_includes/base.njk
```diff
@@ -27,9 +27,8 @@
     </script>
   </head>
   <body class="layout-{{ layout | stripExtension }} page-{{ pageClass or page.fileSlug or 'home' }}{{ ' page-wide' if wide }}">
-    {% if hasBanner == undefined %}
-      {% set hasBanner = true %}
-    {% endif %}
+    {% if hasBanner == undefined %}{% set hasBanner = true %}{% endif %}
+    {% if hasHiddenGeneratedTitle == undefined %}{% set hasHiddenGeneratedTitle = false %}{% endif %}
```
lindsaym-fa: I'm curious if it would be more appropriate and memorable to use `hasGeneratedTitle` which is set to `true` by default. Pages where the generated title is hidden would define `hasGeneratedTitle: false` in that case. Would that be an improvement from your POV?

packages/webawesome/docs/assets/styles/utils.css
```diff
@@ -68,6 +68,10 @@
   /* #endregion */
 
   /* #region shared UI */
+  /* hide 11ty generated titles (set hasHiddenGeneratedTitle front matter to true to hide the title) */
+  .title-generated.is-hidden {
```
lindsaym-fa: Smallest possible nit: Do we use verbs in CSS classes elsewhere? Would `.hidden` be appropriate instead?

lindsaym-fa (commented):
Amazing improvement! 🎉 A couple of considerations re:naming, but I'm happy to approve if you prefer the naming as-is.

packages/webawesome/docs/_includes/base.njk
talbs: Great suggestion - refined!

packages/webawesome/docs/assets/styles/utils.css
talbs: Good call - updated!

talbs:
Okay, @lindsaym-fa. This is ready for another look after addressing your `fa-solid` feedback.

packages/webawesome/docs/_includes/base.njk
```diff
@@ -120,7 +119,7 @@
         <div id="flashes">{% server "flashes" %}</div>
 
         {% block header %}
-          <h1 class="title">{{ title }}</h1>
+          <h1 class="title title-generated{{ ' hidden' if not hasGeneratedTitle }}">{{ title }}</h1>
```
lindsaym-fa: Another thought: Would it make sense to do something like this, in which case we wouldn't need `.title-generated` or `.hidden`?

```suggestion
          {% if hasGeneratedTitle %}
            <h1 class="title">{{ title }}</h1>
          {% endif %}
```

lindsaym-fa (dismissed):
One final pondering, but approved nonetheless!

packages/webawesome/docs/_includes/base.njk
talbs: Okay, I double-checked and this can be removed! I'll simplify this even more.

talbs: Updated - so clean and simple!

lindsaym-fa (approved):
🚢

PR #1717, opened 2025-11-03
Fix <wa-tree>: restore spinner visibility during lazy loading

### LINKED ISSUE #1678: Tree has broken lazy loading status
### Describe the bug
The tree component, as far as I can remember, used to show a spinner to represent a loading status when the lazy loading functionality was being employed. It no longer does so

### To Reproduce

1. Go to 'https://webawesome.com/docs/components/tree/#lazy-loading'
2. Click to expand the tree in the example
3. Observe the absence of the spinner while loading tree items

the-marolie:
Fixes: https://github.com/shoelace-style/webawesome/issues/1678

This PR restores the loading spinner that appears while a tree item is fetching its children during lazy loading.
Previously, the spinner wasn’t rendered because it was placed inside the existing expand-icon slot, which was being hidden by the expand/collapse CSS selectors.

The spinner is now rendered through a dedicated spinner slot, ensuring it displays correctly without interfering with expand or collapse icons.

Preview:
image

### DIFF
--- a/packages/webawesome/src/components/tree-item/tree-item.ts
+++ b/packages/webawesome/src/components/tree-item/tree-item.ts
@@ -252,6 +252,7 @@ export default class WaTreeItem extends WebAwesomeElement {
         class="${classMap({
           'tree-item': true,
           'tree-item-expanded': this.expanded,
+          'tree-item-loading': this.loading,
           'tree-item-selected': this.selected,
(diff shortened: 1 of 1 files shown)


claviska:
Thanks for the PR. This one has been fixed in a similar way in #1786.

PR #1718, opened 2025-11-04
Allow Setting `data-no-anchor` at the View-Level

talbs:
This PR adds support for globally disabling heading anchors on pages through a `hasAnchors` flag. Pages can now set `hasAnchors: false` in their front matter to suppress anchor link generation for all headings on that page.

- Added a `hasAnchors` variable with a default value of `true` in the template
- Modified the `<html>` tag to conditionally add the `data-no-anchor` attribute when `hasAnchors` is set to `false`

One can still generate anchors for a view and then surgically chose to not via the `data-no-anchor` being set on a DOM element.

---

- @lindsaym-fa, similar to https://github.com/shoelace-style/webawesome/pull/1715, mind making sure this approach feels right and is helpful to managing when we want to disable anchor generation at a view level?

### DIFF
--- a/packages/webawesome/docs/_includes/base.njk
+++ b/packages/webawesome/docs/_includes/base.njk
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" data-fa-kit-code="38c11e3f20" data-version="{{ package.version }}" class="wa-cloak">
+<html lang="en" data-fa-kit-code="38c11e3f20" data-version="{{ package.version }}" class="wa-cloak"{% if hasAnchors == false %} data-no-anchor{% endif %}>
   <head>
     {% include 'head.njk' %}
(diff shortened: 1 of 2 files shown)


lindsaym-fa (changes requested):
I think we're missing a piece of the equation, aren't we? What is the inclusion of `data-no-anchor` on `<html>` supposed to be doing?

Here's what I'm seeing locally:
Image

talbs:
Sorry about that, @lindsaym-fa. I think I resolved things by iterating on our `anchor-headings.js` transformer in https://github.com/shoelace-style/webawesome/pull/1718/commits/563df1f054f8aa171c0ff7009e7dfbebe61eee99.

Things should be working as intended now. Mind giving it another look?

lindsaym-fa (approved):
Excellent! 🎉

PR #1719, opened 2025-11-05
Update VSCode extensions.json

coliff:
Add EditorConfig to help ensure consistent line-endings, formatting etc.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,7 @@
 {
   "recommendations": [
     "dbaeumer.vscode-eslint",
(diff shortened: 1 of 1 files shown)


claviska:
The editorconfig file is left over from Shoelace. Since then, we've adopted Prettier which takes care of formatting. I'd prefer not to bring in another since they may compete, but you're welcome to leave it installed locally in your own projects.

coliff:
EditorConfig works well with Prettier and does some things that Prettier doesn't. They are complimentary :-) https://prettier.io/docs/configuration#editorconfig

PR #1723, opened 2025-11-05
Add missing and allow passing of aria attributes for dialogs

### LINKED ISSUE #1722: Dialog components do not set or allow aria attributes
### Describe the bug

The `wa-dialog`, `wa-drawer` and `wa-popover` components use `<dialog ...` but do not set or allow passing in aria attributes.

### To Reproduce

Use `wa-dialog`, `wa-drawer` or `wa-popover` and try to set `aria-labelledby` and `aria-describedby` attributes.

### Demo

If the bug isn't obvious, please provide a link to a CodePen or Fiddle with a minimal reproduction. Bugs that have repros get attention faster than those that don't.

Tip: use the CodePen button on any example in the docs!

### Screenshots
If applicable, add screenshots to help explain the bug.

NA

### Browser / OS
 - OS: Linux
 - Browser: Chrome
 - Browser version: 142.0.7444.59

### Additional information

https://github.com/shoelace-style/webawesome/blob/e2c7286ccc550384080027b86d4b9b41dd9d5959/packages/webawesome/src/components/dialog/dialog.ts#L215-L224

https://github.com/shoelace-style/webawesome/blob/e2c7286ccc550384080027b86d4b9b41dd9d5959/packages/webawesome/src/components/drawer/drawer.ts#L228-L241

https://github.com/shoelace-style/webawesome/blob/e2c7286ccc550384080027b86d4b9b41dd9d5959/packages/webawesome/src/components/popover/popover.ts#L290

The current workaround is to manipulate the shadow dom:

https://github.com/home-assistant/frontend/blob/6f27770d88126d9d47a451a6d7e823ecd43d0cba/src/components/ha-wa-dialog.ts#L125-L147

timmo001:
Fixes: #1722

Adds `aria-labelledby` and `aria-describedby` attributes to:

- wa-dialog
- wa-drawer 
- wa-popover

`wa-dialog` and `wa-drawer` have a default label from the title id, which has been added unless overridden by the caller.

### DIFF
--- a/packages/webawesome/src/components/dialog/dialog.ts
+++ b/packages/webawesome/src/components/dialog/dialog.ts
@@ -1,6 +1,7 @@
 import { html, isServer } from 'lit';
 import { customElement, property, query } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
 import { WaAfterHideEvent } from '../../events/after-hide.js';
 import { WaAfterShowEvent } from '../../events/after-show.js';
 import { WaHideEvent } from '../../events/hide.js';
@@ -75,6 +76,13 @@ export default class WaDialog extends WebAwesomeElement {
 
   /** When enabled, the dialog will be closed when the user clicks outside of it. */
   @property({ attribute: 'light-dismiss', type: Boolean }) lightDismiss = false;
+
+  /** The ID of the element that labels the dialog.
+   *  Overrides the default `aria-labelledby="title"`. */
+  @property({ attribute: 'aria-labelledby' }) ariaLabelledby?: string;
+
+  /** The ID of the element that describes the dialog. */
+  @property({ attribute: 'aria-describedby' }) ariaDescribedby?: string;
 
   firstUpdated() {
     if (this.open) {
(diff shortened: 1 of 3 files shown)


packages/webawesome/src/components/dialog/dialog.ts
```diff
@@ -76,6 +77,13 @@ export default class WaDialog extends WebAwesomeElement {
   /** When enabled, the dialog will be closed when the user clicks outside of it. */
   @property({ attribute: 'light-dismiss', type: Boolean }) lightDismiss = false;
 
+  /** The ID of the element that labels the dialog.
+   *  Overrides the default `aria-labelledby="title"`. */
+  @property({ attribute: 'aria-labelledby' }) ariaLabelledby?: string;
+
+  /** The ID of the element that describes the dialog. */
+  @property({ attribute: 'aria-describedby' }) ariaDescribedby?: string;
```
claviska, line 85: This is not a pattern we use in the library. All labels and descriptions are abstracted into `label` and `description` properties, oftentimes with corresponding slots for HTML content.

To the best of my knowledge, native dialogs don't require aria labels as the dialog is announced as a group, and the label is the first thing that gets read by screen readers due to its position.

An argument could be made for adding a `description` property, but I'm not sure that's necessary either as the dialog can contain arbitrary content and descriptions are more common on form controls.

To save yourself the work, please open an issue and ask before proposing API changes.

PR #1734, opened 2025-11-05
wa-icon remove `currentColor` fill on svg

### LINKED ISSUE #1733: wa-icon forcefully applies `currentColor` fill to svg
### Describe the bug

The _Lucide_ icon library is visually broken in the [docs](https://webawesome.com/docs/components/icon#lucide).  

The issue appears to be with this line of code ([`65cb317`](https://github.com/shoelace-style/webawesome/commit/65cb3175afb560ab4c5874877449615f1f971b49#diff-30d43f39de8ce361de97fd4262471db3c52ccb900fb533711abb5f8abe1ecfe8R18)):

https://github.com/shoelace-style/webawesome/blob/3d4b3ada272326f1b57294bb2996c2f914935ab2/packages/webawesome/src/components/icon/icon.css#L28

This code actually nullifies the example given in the [documentation](https://github.com/shoelace-style/webawesome/blob/3d4b3ada272326f1b57294bb2996c2f914935ab2/packages/webawesome/docs/docs/components/icon.md#L428) for the library's mutator function:

> If necessary, a mutator function can be used to mutate the SVG element before rendering. This is necessary for some libraries due to the many possible ways SVGs are crafted. **For example, icons should ideally inherit the current text color via `currentColor`, so you may need to apply `fill="currentColor` or `stroke="currentColor"` to the SVG element using this function.**

The problem is that some Lucide (our company's, etc.) icons apply `currentColor` to `stroke` rather than `fill`. This can be seen in the markup for the [printer](https://cdn.jsdelivr.net/npm/lucide-static@0.16.29/icons/printer.svg) icon.

### To Reproduce

1. Go to https://webawesome.com/docs/components/icon#lucide
2. View the example for the Lucide icon library.
3. Notice that the settings and printer icons are not accurately visualized.

### Screenshots

Image

### Additional information

This issue was alluded to in #1389.

roydukkey:
Fixes #1733

### DIFF
--- a/packages/webawesome/src/components/icon/icon.css
+++ b/packages/webawesome/src/components/icon/icon.css
@@ -25,7 +25,6 @@
 
 svg {
   height: 1em;
-  fill: currentColor;
   overflow: visible;
 
   /* Duotone colors with path-specific opacity fallback */


claviska:
Thanks for the PR. It looks like we needed to remove the style and the attribute. Closing in favor of #1784.

PR #1736, opened 2025-11-06
CSS: Remove duplicate width property from select component

coliff:
Deleted the duplicate 'width: 100%' style from the select component (already defined on line 56).

Also fixed a typo in a comment in the docs CSS.

You should setup Stylelint to prevent these issues :-)


claviska (approved):
Thanks!

PR #1739, opened 2025-11-06
Add Launch Sale Dialog

talbs:
This PR adds a site-wide promotional dialog for the Web Awesome launch sale, offering a 20% lifetime discount on Pro plans. The dialog is shown to non-Pro users and can be dismissed, with the dismissal state stored in localStorage to prevent showing it again.

- Adds a new `hasSiteDialog` flag to control dialog visibility (defaults to `true`)
- Creates a new reusable dialog component with promotional content and CTA buttons
- Implements dialog logic with localStorage persistence and customizable force-show mode for testing

---

image

---

- @claviska, not sure if this needed, but I recall you mentioning this may be wanted to help promote the very limited sale on Web Awesome Pro. Feel free to review if that's the case.


claviska (dismissed):
Perfect! I assume we're using `localStorage` instead of `sessionStorage` to be slightly less annoying., which is fine by me.

claviska (dismissed):

PR #1746, opened 2025-11-10
Add 'lorem' to the language list in cspell.json

coliff:
Lorem ipsum text is used in examples and as such is flagged as typos in VS Code / Cursor. The change in this PR fixes that.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -236,5 +236,6 @@
   "ignoreRegExpList": [
     "(^|[^a-z])sl[a-z]*(^|[^a-z])"
   ],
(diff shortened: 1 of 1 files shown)


claviska:
I think it will be better if we update the docs to use real content instead of lorem ipsum. Adding the entire dictionary will allow a bunch of non-English words that may allow typos that happen to exist in that dictionary to surface.

In the meantime, cSpell can be suppressed in individual files using comments such as:

```html
<!-- cspell:dictionaries lorem-ipsum -->
```

I've opened #1748 to track the work to improve these examples in our docs.

PR #1752, opened 2025-11-11
Add Dependabot configuration for GitHub Actions

coliff:
Some of the GitHub Actions are using old versions. Would be good to keep them up-to-date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: monthly
+    groups:
+      github-actions:
+        patterns:
+          - "*"


claviska:
We prefer to address updates manually. In the [perhaps unlikely but clearly not impossible] event of a supply chain attack, automatic updates can introduce problems for us and our users.

coliff:
Many of them are old versions.. and as for security, to mitigate supply-chain attacks you can set a cooldown period for say 7 days.
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-

For improved security you should use SHA1 pinned versions and `actions/checkout` should have `persist-credentials: false`

claviska:
Thanks for the followup. I'll defer this decision to @KonnorRogers.

PR #1757, opened 2025-11-11
Ensuring styles apply for input labels when using 'slot="label"'

stevematney:


stevematney:
There doesn't appear to be a bug for this in the list, but I noticed that the proper styling doesn't apply on `wa-input` usage when you use `slot="label"` instead of the `label` attribute. This should resolve that there and in other places where the same issue could occur.

stevematney:
This has been rebased on the current version of the `next` branch. 😄

lindsaym-fa:
Thanks, @stevematney! I'm very sorry for the delay on reviewing this, but I plan to prioritize this tomorrow!

stevematney:
> Thanks, @stevematney! I'm very sorry for the delay on reviewing this, but I plan to prioritize this tomorrow!

No problem! Thank you!

packages/webawesome/src/components/input/input.ts
```diff
@@ -349,7 +349,15 @@ export default class WaInput extends WebAwesomeFormAssociatedElement {
       (typeof this.value === 'number' || (this.value && this.value.length > 0));
 
     return html`
-      <label part="form-control-label label" class="label" for="input" aria-hidden=${hasLabel ? 'false' : 'true'}>
+      <label
+        part="form-control-label label"
+        class=${classMap({
+          label: true,
+          'has-slotted': hasLabelSlot,
```
lindsaym-fa:
```suggestion
          'has-slotted': hasLabel,
```

We'll want to check for label content rather than the slot. Otherwise, we miss labels set via attribute, like this:
Image

packages/webawesome/src/components/color-picker/color-picker.ts
```diff
@@ -1284,7 +1284,14 @@ export default class WaColorPicker extends WebAwesomeFormAssociatedElement {
         })}
         part="trigger-container form-control"
       >
-        <div part="form-control-label" class="label" id="form-control-label">
+        <div
+          part="form-control-label"
+          class=${classMap({
+            label: true,
+            'has-slotted': hasLabelSlot,
```
lindsaym-fa:
```suggestion
            'has-slotted': hasLabel,
```

packages/webawesome/src/components/select/select.ts
```diff
@@ -155,7 +155,7 @@ export default class WaSelect extends WebAwesomeFormAssociatedElement {
   /** The select's value. This will be a string for single select or an array for multi-select. */
   @property({ attribute: 'value', reflect: false })
   set value(val: string | string[] | null) {
-    let oldValue = this.value;
+    const oldValue = this.value;
```
lindsaym-fa: These changes from `let` -> `const` don't appear to be related to input label styles, so I'd recommend reverting these.

packages/webawesome/src/components/slider/slider.ts
```diff
@@ -807,7 +807,7 @@ export default class WaSlider extends WebAwesomeFormAssociatedElement {
         id="label"
         part="label"
         for=${this.isRange ? 'thumb-min' : 'text-box'}
-        class=${classMap({ vh: !hasLabel })}
+        class=${classMap({ vh: !hasLabel, 'has-slotted': hasLabelSlot })}
```
lindsaym-fa:
```suggestion
        class=${classMap({ vh: !hasLabel, 'has-slotted': hasLabel })}
```

packages/webawesome/src/styles/component/form-control.styles.ts
```diff
@@ -26,7 +27,6 @@ export default css`
     display: block;
     color: var(--wa-form-control-hint-color);
     font-weight: var(--wa-form-control-hint-font-weight);
-    line-height: var(--wa-form-control-hint-line-height);
```
lindsaym-fa, line 29: Good catch noticing that we're specifying `line-height` twice in this rule! Since this is using the correct token, we'll actually want to keep this line...

packages/webawesome/src/styles/component/form-control.styles.ts
```diff
@@ -26,7 +27,6 @@ export default css`
...
     margin-block-start: 0.5em;
     font-size: var(--wa-font-size-smaller);
     line-height: var(--wa-form-control-label-line-height);
```
lindsaym-fa:
```suggestion
```

...and remove this one. 😊

lindsaym-fa (commented):
Thanks for this, @stevematney!

@claviska and I had talked about this offline and wondered if there was a way to target label parts with slotted content without adding a new class to the components. The closest I could come up with was adding `:is([part~='form-control-label'], [part~='label']) slot[name='label']::slotted(*)` as a selector for the label styles, but that doesn't quite cut it — we want the styles to apply to the label parts when there's slotted content, not the slotted nodes themselves. I'm convinced that the most foolproof approach is to match the pattern we use for hints and add a class when there's a label.

That said, following the existing hint pattern, my assumption is we want the `has-slotted` class to be present whenever _any_ label content is populated. In that case, we'll want to check for label content with `hasLabel` rather than the slot with `hasLabelSlot`.


This definitely begs the question of whether `has-slotted` is the right name for this class — and the same question goes for the `has-slotted` class used for hints, too. The limitedly supported CSS pseudo-class of the same name ([`:has-slotted`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:has-slotted)) only matches when there's a slot-assigned node. Labels set via attribute wouldn't match in that case, so our `has-slotted` class name conflicts with the platform's intention for `:has-slotted`.

@claviska, I'm curious about your thoughts. We could stick with our own status quo and continue the pattern of using `has-slotted`, or we could use something like `has-label` here instead and someday replace the class used for hints with `has-hint`.

packages/webawesome/src/components/select/select.ts
stevematney: I just have `eslint --fix` running on save in my editor with some global config, which is why these got updated (since they're never reassigned), but I can definitely revert this!

packages/webawesome/src/styles/component/form-control.styles.ts
stevematney: I'll get this set correctly!

claviska:
> We could stick with our own status quo and continue the pattern of using has-slotted, or we could use something like has-label here instead and someday replace the class used for hints with has-hint.

To be honest, I'm fine with either as long as we're consistent. I will say that `has-slotted` feels a little weird if it applies even when something isn't slotted, so maybe that's a vote for `has-label`.

stevematney:
> That said, following the existing hint pattern, my assumption is we want the `has-slotted` class to be present whenever _any_ label content is populated. In that case, we'll want to check for label content with `hasLabel` rather than the slot with `hasLabelSlot`.

I'm happy to update this to include the class when there is any label content (slotted or via the attribute), but I'll wait to make any changes to that until we know the direction that class name should take.

For me, `:has-slotted` would obviously be the most preferred approach as a built-in pseudo-class, but I opted not to put it in since its support is currently only in Firefox, as you mentioned. I wanted the effect of my changes to have as limited a scope as possible, and these styles were already applied with content from the attribute, so I went with only adding the new class when we explicitly have _slotted_ content. (And I think you're right that `has-slotted` would be kind of a misnomer if it was applying for all cases of content in the label, as opposed to just slotted content.)

claviska:
I don't think there will be any friction with `has-label` then. No need to hold things up for deliberation :)

stevematney:
Under the new world of `has-label`, the label styles will be applied both for the current `:has(*:not(:empty))` selector, and also for the `.has-label` selector for non-slotted content. If the idea is that `:has-slotted` should _eventually_ be used, it might be good to keep it that way so that, when the `.has-label` functionality is removed, we don't lose the non-slotted selector. If the goal is to be as minimal as possible, I can remove the `:has(*:not(:empty))` selector so `.has-label` is the only one left.

stevematney:
I've updated the PR with the `has-label` changes. If we do want to remove the old selector, I can do that, but I've left it in for now. 😄

lindsaym-fa (approved):
No complaints! Thanks again, @stevematney!

I've created a Pro PR to do the same for `<wa-combobox>`: https://github.com/shoelace-style/webawesome-pro/pull/122  We'll merge this at the same time.

PR #1769, opened 2025-11-13
Gating Launch Sale-Based Banner and Dialog

talbs:
This PR implements gating for launch sale-based promotional content by introducing a `req.stripe.discount.active` flag to control the visibility of both the site-wide dialog and banner, ensuring they only display when the Stripe discount is active and the user doesn't have Pro.

### DIFF
--- a/packages/webawesome/docs/_includes/_banner-wa-launch.njk
+++ b/packages/webawesome/docs/_includes/_banner-wa-launch.njk
@@ -1,5 +1,5 @@
 {% raw %}
-  {%- if not currentUser.hasPro -%}
+  {%- if data.stripeDiscount.active and not currentUser.hasPro -%}
   <div slot="banner" class="banner-wa-launch wa-dark">
     <div class="banner-content wa-split">
       <div class="wa-cluster wa-gap-s">
--- a/packages/webawesome/docs/_includes/_banner-wa-launch.njk
+++ b/packages/webawesome/docs/_includes/_banner-wa-launch.njk
@@ -1,5 +1,5 @@
 {% raw %}
-  {%- if data.stripeDiscount.active and not currentUser.hasPro -%}
+  {%- if req.stripe.discount.active and not currentUser.hasPro -%}
   <div slot="banner" class="banner-wa-launch wa-dark">
     <div class="banner-content wa-split">
       <div class="wa-cluster wa-gap-s">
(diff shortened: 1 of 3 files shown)


packages/webawesome/docs/_includes/_dialog-wa-launch.njk
```diff
@@ -1,97 +1,100 @@
...
 
+        // Initialize when ready
+        if (customElements.get('wa-dialog')) {
+          initDialog();
+        } else {
+          document.addEventListener('wa-discovery-complete', initDialog, { once: true });
+          // Fallback timeout in case wa-discovery-complete doesn't fire
+          setTimeout(initDialog, 100);
+        }
+      })();
```
KonnorRogers:
```suggestion
      customElements.whenDefined("wa-dialog").then(() => {
        initDialog()
      })
```

KonnorRogers (approved):
seem good to me, just a minor note about a better way to track if `<wa-dialog>` has upgraded

packages/webawesome/docs/_includes/_banner-wa-launch.njk
```diff
@@ -1,5 +1,5 @@
 {% raw %}
-  {%- if not currentUser.hasPro -%}
+  {%- if data.stripeDiscount.active and not currentUser.hasPro -%}
```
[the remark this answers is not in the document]
KonnorRogers: don't listen to copilot, its drunk. nunjucks automatically does safe navigation.

KonnorRogers: so its not obvious, but `data.stripeDiscount` is for the `purchase` page where the discount may not be the same as the site-wide discount.

This should check: 


```njk
req.stripe.discount.active
```

I wonder if the naming should change to like `req.stripe.activePromotion` or something 🤔

KonnorRogers (changes requested):
one minor note about site wide discount vs page discount

packages/webawesome/docs/_includes/_banner-wa-launch.njk
talbs: Thanks! I think `req.stripe.discount.active` is fine, but if you wanna change the name, go for it. I'll use `req.stripe.discount.active` for now.

KonnorRogers (approved):

Issue #1483, opened 2025-09-22
Cancelling wa-hide on wa-dropdown hangs browser


medmunds:
### Describe the bug
If you call event.preventDefault() on a wa-hide event from a wa-dropdown, the browser tab hangs.

### To Reproduce

1. Create a wa-dropdown, and add a wa-hide handler that calls preventDefault()—e.g.:
    ```js
    document.querySelector('wa-dropdown')
      .addEventListener('wa-hide', event => {
        event.preventDefault();
      });
    ```
2. Click to open the dropdown, then click again to hide it.

Results: The browser tab is hung.
    
### Demo

https://codepen.io/mvedmunds/pen/vENoWyK?editors=1000

### Browser / OS
WebAwesome 3.0.0-beta.5
Chrome, Firefox, Safari

### Additional information
It seems to get stuck repeatedly sending wa-show events after the cancelled wa-hide.

Cancelling wa-hide on a dropdown isn't documented, but there seems to be [code that intends to handle](https://github.com/shoelace-style/webawesome/blob/efbc4045243261ce244e27369a988fab8d6a8080/packages/webawesome/src/components/dropdown/dropdown.ts#L257-L260) it.

Context: discussion #1452. This is not blocking me; just didn't want to lose the info if this is a real bug.

JRJurman:
> Cancelling wa-hide on a dropdown isn't documented, but there seems to be code that intends to handle it.

~This is actually documented as something that should be possible (see [Checkable Items](https://webawesome.com/docs/components/dropdown#checkable-items))~

I'm actually running into this issue, trying to create a combobox (using an input inside a dropdown).

🤦 just re-read the docs and realize it says `wa-select` is the event to cancel, not `wa-hide`. That appears to work as expected, but I still need `wa-hide` for my own components 😓

claviska:
Fixed in #1774. Thanks @JRJurman!

PR #1774, opened 2025-11-16
Fix: Allow wa-hide to be cancelled in WaDropdown

JRJurman:
## Summary

Related issue: #1483 

Previously, when cancelling the `wa-hide` event for a `<wa-dropdown>`, developers would see the browser hang on an infinite loop. This was caused by constant flipping of the `this.open` property, set as `true` in [dropdown.ts:257-260](https://github.com/shoelace-style/webawesome/blob/5a6ce1c4d397e5e64c9ed513e2fc2d18fe142184/packages/webawesome/src/components/dropdown/dropdown.ts#L257-L260) (where we handle `preventDefault` in `hideMenu`), and then set as `false` in [dropdown.ts:230](https://github.com/shoelace-style/webawesome/blob/5a6ce1c4d397e5e64c9ed513e2fc2d18fe142184/packages/webawesome/src/components/dropdown/dropdown.ts#L230) (where we close all dropdowns, including `this`).

### Why do we need this?

When trying to implement a combobox (using an input as a trigger), we'd like the ability to better control the showing and hiding of the dropdown menu. That requires the ability to specifically cancel `wa-hide` (not just `wa-select`).

### Solutions

There were many different ways to solve this, including:
1. checking and skipping `this` when closing dropdowns in [dropdown.ts:230](https://github.com/shoelace-style/webawesome/blob/5a6ce1c4d397e5e64c9ed513e2fc2d18fe142184/packages/webawesome/src/components/dropdown/dropdown.ts#L230)
2. keeping track of the cancellation as a dedicated (private) variable
3. removing the `.open` side-effect in the [hideMenu](https://github.com/shoelace-style/webawesome/blob/5a6ce1c4d397e5e64c9ed513e2fc2d18fe142184/packages/webawesome/src/components/dropdown/dropdown.ts#L258) and [showMenu](https://github.com/shoelace-style/webawesome/blob/5a6ce1c4d397e5e64c9ed513e2fc2d18fe142184/packages/webawesome/src/components/dropdown/dropdown.ts#L226) handlers
4. Gating inside the `showMenu` and `hideMenu` calls, checking if the menu is already closed or open.
5. Gating the `showMenu` and `hideMenu` calls in the [updated](https://github.com/shoelace-style/webawesome/blob/5a6ce1c4d397e5e64c9ed513e2fc2d18fe142184/packages/webawesome/src/components/dropdown/dropdown.ts#L114-L119) function

In this PR, I've opted for a combination of the last two options. This feels the most semantically correct (we only want to call `showMenu` and `hideMenu` when we intend to trigger those functions). I'm happy to implement one of the other options, or explore a different solution here if that would be more desirable.

## Recording

Below is a recording using a modified version of the `<wa-dropdown>` docs.

<details>
<summary>Code Snippet (modified)</summary>

```html
<div class="dropdown-checkboxes">
  <wa-dropdown>
    <wa-button slot="trigger" with-caret>View</wa-button>

    <wa-dropdown-item type="checkbox" value="canvas" checked>Show canvas</wa-dropdown-item>
    <wa-dropdown-item type="checkbox" value="grid" checked>Show grid</wa-dropdown-item>
    <wa-dropdown-item type="checkbox" value="source">Show source</wa-dropdown-item>

    <wa-divider></wa-divider>

    <wa-dropdown-item value="preferences">Preferences…</wa-dropdown-item>
  </wa-dropdown>
</div>

<script>
  const container = document.querySelector('.dropdown-checkboxes');
  const dropdown = container.querySelector('wa-dropdown');

  dropdown.addEventListener('wa-select', event => {
    if (event.detail.item.type === 'checkbox') {
      // Checkbox
      console.log(event.detail.item.value, event.detail.item.checked ? 'checked' : 'unchecked');
    } else {
      // Not a checkbox
      console.log(event.detail.item.value);
    }
  });

  // NEW EVENT LISTENERS FOR TESTING wa-hide
  dropdown.addEventListener('wa-hide', event => {
    event.preventDefault();
  });
  // (cancelling wa-select is still required)
  dropdown.addEventListener('wa-select', event => {
    event.preventDefault();
  })
</script>
```

</details>

<video aria-label="a user's cursor clicks on a button to open a menu, and then clicks on a checklist selection. The menu stays open, and on clicking the button control and the background, the menu stays open." src="https://github.com/user-attachments/assets/02ba9efa-3d41-4ecc-9abe-4a87f7d33ae7"></video>

## Testing

Aside from manual testing, this PR adds the following unit tests:
1. `<wa-dropdown> should not be open by default`
2. `<wa-dropdown> should fire a single show/after-show and hide/after-hide in normal open/close flow`
3. `<wa-dropdown> should fire a single show/after-show and hide/after-hide when wa-hide event is cancelled`

## Other changes

There are still some semantic oddities with the underlying code here. In an effort to reduce the total number of lines changed, I've opted not to only make the changes required to fix the bug here, however here are some other pieces that are probably worth fixing as a follow up (with potentially more dedicated testing).

1. Setting `this.open = true` in the `hideMenu` `preventDefault` handler assumes that the menu was already open. This isn't always the case (e.g. when the component is first upgraded, `open` is set to `undefined`, and triggers a `hideMenu` call). We should either remove these, or account for whatever the previous `open` value actually was. This is somewhat handled in the `updated` function call now, but if we ever expose `hideMenu` or `showMenu`, this could cause unexpected behaviors.
2. `closeAllSubmenus` is called regardless of whether or not the `wa-hide` event was cancelled. This should almost certainly be moved to inside `hideMenu`.
3. `showMenu` still closes all dropdowns (including `this`) and then reopens itself. It might be worth checking and skipping when `dropdown === this`.
4. We dynamically sync `aria-expanded` in syncAriaAttributes for `nativeButton`, but not the underlying `this.menu`

### DIFF
--- a/packages/webawesome/src/components/dropdown/dropdown.test.ts
+++ b/packages/webawesome/src/components/dropdown/dropdown.test.ts
@@ -1,9 +1,119 @@
-import { expect, fixture, html } from '@open-wc/testing';
+import { aTimeout, expect, fixture, html, waitUntil } from '@open-wc/testing';
+import sinon from 'sinon';
+import type WaDropdown from './dropdown.js';
 
 describe('<wa-dropdown>', () => {
   it('should render a component', async () => {
     const el = await fixture(html` <wa-dropdown></wa-dropdown> `);
 
     expect(el).to.exist;
+  });
+
+  it('should not be open by default', async () => {
+    const el = await fixture<WaDropdown>(html`
+      <wa-dropdown>
+        <wa-button slot="trigger">Dropdown</wa-button>
+        <wa-dropdown-item>One</wa-dropdown-item>
+      </wa-dropdown>
+    `);
+
+    await el.updateComplete;
+    await aTimeout(200);
+
+    expect(el.open).to.be.false;
+  });
+
+  it('should fire a single show/after-show and hide/after-hide in normal open/close flow', async () => {
+    const el = await fixture<WaDropdown>(html`
+      <wa-dropdown>
+        <wa-button slot="trigger">Dropdown</wa-button>
+        <wa-dropdown-item>One</wa-dropdown-item>
+        <wa-dropdown-item>Two</wa-dropdown-item>
+      </wa-dropdown>
+    `);
+
+    // setup spies to track how often we see different show/hide events
+    const showSpy = sinon.spy();
+    const afterShowSpy = sinon.spy();
+    const hideSpy = sinon.spy();
+    const afterHideSpy = sinon.spy();
+
+    el.addEventListener('wa-show', showSpy);
+    el.addEventListener('wa-after-show', afterShowSpy);
+    el.addEventListener('wa-hide', hideSpy);
+    el.addEventListener('wa-after-hide', afterHideSpy);
+
+    // open the dropdown by triggering a click on the trigger
+    const trigger = el.querySelector<HTMLElement>('[slot="trigger"]')!;
+    trigger.click();
+
+    await waitUntil(() => showSpy.calledOnce);
+    await waitUntil(() => afterShowSpy.calledOnce);
+
+    expect(showSpy.callCount).to.equal(1);
+    expect(afterShowSpy.callCount).to.equal(1);
+
+    expect(el.open).to.be.true;
+
+    // close the dropdown by clicking the trigger again
+    trigger.click();
+
+    await waitUntil(() => hideSpy.calledOnce);
+    await waitUntil(() => afterHideSpy.calledOnce);
+
+    expect(hideSpy.callCount).to.equal(1);
+    expect(afterHideSpy.callCount).to.equal(1);
+
+    expect(el.open).to.be.false;
+  });
+
+  it('should fire a single show/after-show and hide/after-hide when wa-hide event is cancelled', async () => {
+    const el = await fixture<WaDropdown>(html`
+      <wa-dropdown>
+        <wa-button slot="trigger">Dropdown</wa-button>
+        <wa-dropdown-item>One</wa-dropdown-item>
+        <wa-dropdown-item>Two</wa-dropdown-item>
+      </wa-dropdown>
+    `);
+
+    // setup spies to track how often we see different show/hide events
+    const showSpy = sinon.spy();
+    const afterShowSpy = sinon.spy();
+    const hideSpy = sinon.spy();
+    const afterHideSpy = sinon.spy();
+
+    el.addEventListener('wa-show', showSpy);
+    el.addEventListener('wa-after-show', afterShowSpy);
+
+    // Intercept wa-hide and prevent it
+    el.addEventListener('wa-hide', event => {
+      event.preventDefault();
+      hideSpy(event);
+    });
+
+    el.addEventListener('wa-after-hide', afterHideSpy);
+
+    // open the dropdown by triggering a click on the trigger
+    const trigger = el.querySelector<HTMLElement>('[slot="trigger"]')!;
+    trigger.click();
+
+    await waitUntil(() => showSpy.calledOnce);
+    await waitUntil(() => afterShowSpy.calledOnce);
+
+    expect(showSpy.callCount).to.equal(1);
+    expect(afterShowSpy.callCount).to.equal(1);
+
+    expect(el.open).to.be.true;
+
+    // click on the trigger (which should do nothing to the open state)
+    trigger.click();
+
+    await waitUntil(() => hideSpy.calledOnce);
+
+    expect(hideSpy.callCount).to.equal(1);
+    // after-hide should not have been called if hide is cancelled
+    expect(afterHideSpy.callCount).to.equal(0);
+
+    expect(el.open).to.be.true;
   });
 });
--- a/packages/webawesome/src/components/dropdown/dropdown.test.ts
+++ b/packages/webawesome/src/components/dropdown/dropdown.test.ts
@@ -9,9 +9,9 @@ describe('<wa-dropdown>', () => {
     expect(el).to.exist;
   });
 
-  it('should not be open by default', async () => {
+  it('should respect the open attribute when included', async () => {
     const el = await fixture<WaDropdown>(html`
-      <wa-dropdown>
+      <wa-dropdown open>
         <wa-button slot="trigger">Dropdown</wa-button>
         <wa-dropdown-item>One</wa-dropdown-item>
       </wa-dropdown>
@@ -20,7 +20,7 @@ describe('<wa-dropdown>', () => {
     await el.updateComplete;
     await aTimeout(200);
 
-    expect(el.open).to.be.false;
+    expect(el.open).to.be.true;
   });
 
   it('should fire a single show/after-show and hide/after-hide in normal open/close flow', async () => {
--- a/packages/webawesome/src/components/dropdown/dropdown.ts
+++ b/packages/webawesome/src/components/dropdown/dropdown.ts
@@ -109,6 +109,18 @@ export default class WaDropdown extends WebAwesomeElement {
 
   async updated(changedProperties: PropertyValues) {
     if (changedProperties.has('open')) {
+      const previousOpen = changedProperties.get('open');
+      // check if the previous value is the same
+      // (if they are, do not trigger menu showing / hiding)
+      if (previousOpen === this.open) {
+        return;
+      }
+      // check if we are changing from undefined to false
+      // (if we are, we can skip menu hiding)
+      if (previousOpen === undefined && this.open === false) {
+        return;
+      }
+
       this.customStates.set('open', this.open);
 
       if (this.open) {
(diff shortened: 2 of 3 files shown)


packages/webawesome/src/components/dropdown/dropdown.test.ts
```diff
@@ -1,9 +1,119 @@
...
+import type WaDropdown from './dropdown.js';
 
 describe('<wa-dropdown>', () => {
   it('should render a component', async () => {
     const el = await fixture(html` <wa-dropdown></wa-dropdown> `);
 
     expect(el).to.exist;
   });
+
+  it('should not be open by default', async () => {
```
JRJurman: I wasn't actually able to get this test to fail, so I'm happy to remove it if it feels redundant. This was an attempt to show the scenario where the change of `open` from `undefined` to `false` in `updated()` caused the menu to start opened (a side-effect of the `preventDefault` setting `this.open` to `true` - part of the original code). 

This doesn't happen in the testing environment because the component definition already exists by the time this fixture is created.

It occurs to me now that there's a way where we could create a new component tag (e.g. `wa-dropdown-for-testing`) and then upgrade as part of the test. If that feels worthwhile, I could make that test, either as part of this PR, or as a separate PR.

packages/webawesome/src/components/dropdown/dropdown.ts
```diff
@@ -227,6 +239,12 @@ export default class WaDropdown extends WebAwesomeElement {
       return;
     }
 
+    // if this dropdown is already open, do nothing
+    // (this can happen when wa-hide was cancelled)
+    if (this.popup.active) {
+      return;
+    }
```
JRJurman, line 246: I don't love checking `this.popup.active` directly. I'm happy to use alternative properties or checks here, but this seemed the most direct, and simpliest way to validate if the menu was already opened.

packages/webawesome/src/components/dropdown/dropdown.ts
```diff
@@ -109,6 +109,18 @@ export default class WaDropdown extends WebAwesomeElement {
...
+      // check if the previous value is the same
+      // (if they are, do not trigger menu showing / hiding)
+      if (previousOpen === this.open) {
+        return;
+      }
+      // check if we are changing from undefined to false
+      // (if we are, we can skip menu hiding)
+      if (previousOpen === undefined && this.open === false) {
+        return;
+      }
```
JRJurman, line 122: as mentioned above, when a component is upgraded, we see `open` change from `undefined` to `false`, which would trigger the `hideMenu`, which would then be cancelled, which would set `this.open = true`. 

All of that would result in the menu defaulting open. In order to prevent this, we check here the previous and open state.

JRJurman: I realize here that I'm checking if `this.open` is specifically `false`, and we could instead check `!this.open`. I feel like that is semantically ambigious, especially since we're doing a specific check for `previousOpen === undefined`, and I don't want to conflate the case that is `undefined` -> `undefined`.

packages/webawesome/src/components/dropdown/dropdown.ts
```diff
@@ -109,6 +109,18 @@ export default class WaDropdown extends WebAwesomeElement {
 
   async updated(changedProperties: PropertyValues) {
     if (changedProperties.has('open')) {
+      const previousOpen = changedProperties.get('open');
+      // check if the previous value is the same
+      // (if they are, do not trigger menu showing / hiding)
+      if (previousOpen === this.open) {
+        return;
+      }
```
JRJurman, line 117: this is the specific check that skips `hideMenu` or `showMenu` if we already are in the state we want to be in.

JRJurman: I also manually validated that `<wa-dropdown>` will start open if the attribute is set on the tag (e.g. `<wa-dropdown open>`). If we were worried that that wouldn't happen, I'm happy to write a test around that as well.

packages/webawesome/src/components/dropdown/dropdown.test.ts
```diff
@@ -1,9 +1,119 @@
...
+      </wa-dropdown>
+    `);
+
+    await el.updateComplete;
+    await aTimeout(200);
+
+    expect(el.open).to.be.false;
+  });
+
+  it('should fire a single show/after-show and hide/after-hide in normal open/close flow', async () => {
```
JRJurman, line 26: this test passed before and after this PR. It is the baseline behavior that validates we see all the events we expect to see when opening and closing the menu.

packages/webawesome/src/components/dropdown/dropdown.test.ts
```diff
@@ -1,9 +1,119 @@
...
+    await waitUntil(() => hideSpy.calledOnce);
+    await waitUntil(() => afterHideSpy.calledOnce);
+
+    expect(hideSpy.callCount).to.equal(1);
+    expect(afterHideSpy.callCount).to.equal(1);
+
+    expect(el.open).to.be.false;
+  });
+
+  it('should fire a single show/after-show and hide/after-hide when wa-hide event is cancelled', async () => {
```
JRJurman, line 70: This test validates the fix. Before the changes in this PR, this test would error in the test run (hanging in browser execution, and failing in headless runs after 2 minutes). After the PR changes, this test passes immediately in both contexts.

packages/webawesome/src/components/dropdown/dropdown.test.ts
JRJurman: Okay, I was able to make a dedicated test that validates the behavior on-upgrade (fails on the `next` branch, but passes on this PR). Let me know if it would be worthwhile to include that here as part of this PR, or if this kind of interaction is beyond the scope of support here (or if it should be its own PR):

https://gist.github.com/JRJurman/6469216b460b8faf2d9aeeea05d60a47

packages/webawesome/src/components/dropdown/dropdown.ts
claviska: > If we were worried that that wouldn't happen, I'm happy to write a test around that as well.

If you have the time, that would be great. Thanks!

packages/webawesome/src/components/dropdown/dropdown.ts
claviska: It's a public property, I have no concerns :)

packages/webawesome/src/components/dropdown/dropdown.ts
claviska: I appreciate the extra info in these comments. It's super helpful! Thanks again

packages/webawesome/src/components/dropdown/dropdown.test.ts
claviska: We can probably remove this one. It will be pretty obvious to anyone testing if they're all open by default.

claviska (dismissed):
Solid PR, thanks for tackling this!

I left some inline comments, but nothing that blocks this from getting merged. The only thing we need to add is a changelog entry, which I'm happy to do.

packages/webawesome/src/components/dropdown/dropdown.test.ts
JRJurman: Sounds good, I can remove this 👍

packages/webawesome/src/components/dropdown/dropdown.ts
JRJurman: Yeah! I'm happy to add a test here to verify that behavior 👍

JRJurman:
@claviska - tests updated, and I added a changelog. Let me know if there are any other changes you want included here, and I'm happy to put those together 😄

JRJurman:
Also, I just realized a good chunk of the PR description was missing because I forgot an ending video tag when I added the aria-label 🤦 - well, none of that was super critical, but there's more details there if anyone is interested 😅

claviska:
> This is somewhat handled in the updated function call now, but if we ever expose hideMenu or showMenu, this could cause unexpected behaviors.

No worries here...we want to keep these private and control things with the attribute/property.

> closeAllSubmenus is called regardless of whether or not the wa-hide event was cancelled. This should almost certainly be moved to inside hideMenu.

Agree. Do you want this to land in a subsequent PR or add it here?

> showMenu still closes all dropdowns (including this) and then reopens itself. It might be worth checking and skipping when dropdown === this.

Good catch. I imagine this will cause a visual blip, so I'm fine with this change.

> We dynamically sync aria-expanded for this element, but not the underlying menu

I might not be understanding this, but the menu's `aria-expanded` attribute is synced in `syncAriaAttributes()`.

JRJurman:
> I might not be understanding this, but the menu's aria-expanded attribute is synced in syncAriaAttributes().

@claviska - sorry, I should have pointed directly here - in `syncAriaAttributes()`, at the very end, we dynamically set `nativeButton` based on `this.open`, but not `this.menu`.

As for the other two things (`closeAllSubmenus` and `dropdown === this`) - I don't know that we absolutely need those for the moment (I didn't aggressively test the dropdown in the configurations that would be impacted by these). I'm happy to make that change here but it wouldn't be until late today / tomorrow. I think I'd prefer making those changes as separate PR(s), when we actually have a minimal repo to show a failure around / build a test for (rather than trying to find one now).

claviska:
Thanks for the fantastic PR, @JRJurman! 🫡

Issue #1775, opened 2025-11-17
Docs are missing a page title and `og:...` meta tags


0warie:
### Describe the bug

When navigating to **any** of the pages in the documentation, the page title is always just the URL path.

### Screenshots
<img width="1100" height="181" alt="Image" src="https://github.com/user-attachments/assets/98cd60fd-1627-44ce-94a6-802b412c1010" />

### Browser / OS
 - OS: Windows 11
 - Browser: Firefox
 - Browser version: 145.0 (64-bit)

0warie:
Making a pr for this now

claviska:
Fixed in #1776

PR #1750, opened 2025-11-10
Revert Custom Last Modified Utility

talbs:
This work undoes the addition of the custom `getLastModifiedISO` utility added in 6041d54a428c8d97e8c4cff90cec08e2bcd9f164 in favor of a ~~better and more performant~~ static solution.

PR #1776, opened 2025-11-17
fix: re-add page titles and metadata

0warie:
related issue: #1775 

Currently the page title on all of the docs pages is just the page's URL (or part of it). For example:
- https://webawesome.com/docs/components/icon -> `https://webawesome.com/docs/`
- https://webawesome.com/docs/color-palettes -> `https://webawesome.com/docs/color-palettes`

Doing some quick investigating, this was caused by: https://github.com/shoelace-style/webawesome/pull/1750.

In the process of removing the "Last Modified Utility", the entire `addGlobalData` method was removed which also contains the logic for adding the composing the page title and meta tags.


talbs (approved):
Thanks so much for noticing and resolving this, @0warie!

Your changes look good and bring back that needed title and Open Graph functionality - appreciate it!

claviska (approved):

PR #1779, opened 2025-11-17
Fix card slot detection

claviska:
Adds all detectable slots to the `HasSlotController` so they respond correctly when items are added/removed.

Solves the issue described in #1450.


talbs (approved):
👀 good. Getting 🎁🎁🎁.

PR #1788, opened 2025-11-18
fix(wa-spinner): implement --track-width support

nguyenyou:
The CSS variable --track-width is documented as a supported option, but the current implementation doesn’t apply it. This PR adds proper support for --track-width so the component behaves as documented.


nguyenyou:
Duplicated https://github.com/shoelace-style/webawesome/pull/1316

PR #1791, opened 2025-11-18
Simplify spinner internals and fix track width bug

### LINKED ISSUE #1317: Spinner's track-width css property is not being applied to track and indicator width
The track width CSS property is not being applied to the track and indicator width.

PR - https://github.com/shoelace-style/webawesome/pull/1316

claviska:
This is my best attempt to fix #1317. I tried numerous methods to make it work with the SVG, but competing with the font size scale means a nasty calculation that worked in Chrome and Safari, but not Firefox.

In this PR, I removed the SVG and replaced the stroke animation with a simpler spin. This _does_ change the visuals a bit (no more rounded end caps on the indicator and the animation is simpler), but I wasn't able to get it done any other way.

@lindsaym-fa we don't have to merge this one, but I don't see another way to solve the problems mentioned in #1316 with the existing SVG. I'm not saying there _isn't_ a way — but I haven't been able to find a cross-browser solution yet. Instead of scrapping all the work, I'm submitting this as an option to consider.


claviska:
Closing as stale.

PR #1793, opened 2025-11-18
Define and import cascade layers

lindsaym-fa:
> _One `@` to rule them all,_
> _One `@` to find them,_
> _One `@` to bring them all,_ 
> _and in the darkness bind them._

Currently, we define all CSS cascade layers in `webawesome.css` to make it easy for folks to use themes, native styles, and CSS utilities together and ensure each has the right precedence. Folly, it turns out. As soon as we take a more modular approach to styles and exclude `webawesome.css` in favor of only the desired CSS features — as we do in Projects — we get issues like [this one reported in Discord](https://discord.com/channels/739437527907303535/1436063593069019257), where `wa-utilities` styles don't necessarily have precedence over `wa-native` styles. Without an `@layer` rule to define precedence at the outset, browsers will give layers precedence in the order they appear in code.

This work establishes a `layers.css` file to serve as a single source of truth for cascade layer precedence and imports this stylesheet in:
- All palettes (via `/styles/color/palettes/base.css`)
- All themes (via each `/styles/themes/{name}.css`)
- `native.css`
- `utilities.css`

---

@claviska, I opted to import `layers.css` in just cornerstone stylesheets. E.g., `utilities.css` includes an `@import` for `layers.css`, but the files that utilities.css itself imports — like `utilities/fouce.css` and `utilities/visually-hidden.css` — do not. Let me know if you feel this is sufficient or if we'd prefer to be ultra-safe and add the import to those lower-level stylesheets too.


claviska (dismissed):
Nice work! This looks like we discussed! :shipit:

claviska (approved):

PR #1794, opened 2025-11-19
Add spellcheck GitHub Action workflow

coliff:
- Uses cSpell. There's already a `cspell.yml`config in the project root so the GitHub Action will use that.
- Uses SHA1 pinned versions for improved security.
- Only checks for changed files.

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -1,0 +1,28 @@
+name: "Check spelling"
+
+on:
+  push:
+    branches: [main]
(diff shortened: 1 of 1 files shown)


claviska:
We use cSpell loosely, and we don't want to start seeing blocking errors in the CI for spelling issues. We really do appreciate the eagerness to help, but opening a discussion before submitting PRs will save everyone time :)

coliff:
I suggest adding this then to prevent it blocking/failing then:

`strict: false`

PR #1798, opened 2025-11-19
Fix tree-item initial RTL check

wendevlin:
If you are in rtl direction and have a tree-item under a tree item the initial arrow was the wrong direction.

### DIFF
--- a/packages/webawesome/src/components/tree-item/tree-item.ts
+++ b/packages/webawesome/src/components/tree-item/tree-item.ts
@@ -243,7 +243,7 @@ export default class WaTreeItem extends WebAwesomeElement {
   }
 
   render() {
-    const isRtl = this.localize ? this.localize.dir() === 'rtl' : this.dir === 'rtl';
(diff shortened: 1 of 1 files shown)


packages/webawesome/src/components/tree-item/tree-item.ts
```diff
@@ -243,7 +243,7 @@ export default class WaTreeItem extends WebAwesomeElement {
   }
 
   render() {
-    const isRtl = this.hasUpdated ? this.localize.dir() === 'rtl' : this.dir === 'rtl';
+    const isRtl = this.localize ? this.localize.dir() === 'rtl' : this.dir === 'rtl';
```
KonnorRogers: 🤔 this will break in SSR environments, and probably cause a hydration mismatch. May need to queue up a 2nd render to fix this

claviska:
```suggestion
    const isRtl = this.localize.dir() === 'rtl';
```

claviska (approved):

PR #1804, opened 2025-11-20
🚧 Support PostHog Analytics

talbs:


talbs:
Closing this as I think we can handle this all in `webawesome-app` instead.

PR #1815, opened 2025-11-21
Bug fix: el.form = 'foo' no longer sets the el.form property

KonnorRogers:
`submit-on-enter` was broken for me when using `<wa-input form="my-form">`

The reason is that it was setting `waInput.form` to equal "foo" instead of returning the `HTMLFormElement`.

This also breaks platform expectations of what an `el.form` should be.


partzkoh18-rgb (approved):

partzkoh18-rgb (approved):

packages/webawesome/src/components/button/button.ts
```diff
@@ -135,24 +134,23 @@ export default class WaButton extends WebAwesomeFormAssociatedElement {
...
-    button.style.height = '0';
-    button.style.clipPath = 'inset(50%)';
-    button.style.overflow = 'hidden';
-    button.style.whiteSpace = 'nowrap';
+    button.style.position = 'absolute !important';
+    button.style.width = '0 !important';
+    button.style.height = '0 !important';
+    button.style.clipPath = 'inset(50%) !important';
+    button.style.overflow = 'hidden !important';
+    button.style.whiteSpace = 'nowrap !important';
```
claviska, line 152: `!important` is interesting...were you seeing a visual blip?

packages/webawesome/src/components/button/button.ts
```diff
@@ -135,24 +134,23 @@ export default class WaButton extends WebAwesomeFormAssociatedElement {
 
   private constructLightDOMButton() {
     const button = document.createElement('button');
+
+    for (const attribute of this.attributes) {
+      button.setAttribute(attribute.name, attribute.value);
+    }
```
claviska, line 144: I'm fine with this, but I'm curious if we were missing any attributes in the array below.

claviska (dismissed):
Questions asked out of curiosity, but nothing blocking :)

packages/webawesome/src/components/button/button.ts
KonnorRogers: @claviska my theory was this may copy over a `style=""` attribute, perhaps I should just skip that instead 🤔

packages/webawesome/src/components/button/button.ts
KonnorRogers: We weren't, but the problem was sometimes I would be checking a `data-*` attribute in a form submission, and the data attribute was missing because it was on the `<wa-button>`, and not the constructed button which gets tagged as the submitter. For example:

```js
document.addEventListener("submit", (e) => {
  if (e.submitter.getAttribute("data-foo")) {}
})
```

PR #1824, opened 2025-11-25
add rotate, flip and animations to wa-icon (#1457)

### LINKED ISSUE #1457: Add rotate, flip, and FA animations to `<wa-icon>`
The subject says it all.

TUEgemenKancal:
* add properties and styles for rotating, flipping and animations to icon.ts and icon.css
* add documentation for new features to icon.md
* add several tests to icon.test.ts

### DIFF
--- a/packages/webawesome/src/components/icon/icon.styles.ts
+++ b/packages/webawesome/src/components/icon/icon.styles.ts
@@ -6,6 +6,7 @@ export default css`
     --primary-opacity: 1;
     --secondary-color: currentColor;
     --secondary-opacity: 0.4;
+    --rotate-angle: 0deg;
 
     box-sizing: content-box;
     display: inline-flex;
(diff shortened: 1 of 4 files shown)


claviska:
Internal discussion copied here for reference:

> There definitely are a lot of custom properties introduced, but they're all consistent with the custom properties FA offers: https://docs.fontawesome.com/web/style/animate
>
> FA treats their custom props as global, so I wonder if we could consolidate some of the overlap. e.g., `--beat-scale` and `--beat-fade-scale` could probably just be a single `--beat-scale` since it's intuitive enough that this would apply to any animation with the keyword "beat," including "beat-fade".
> 
> Though I'm only seeing that overlap in two custom prop pairs — `--beat-scale` / `--beat-fade-scale` and `--fade-opacity` / `--beat-fade-opacity` — so that doesn't do much to reduce the number of custom props.
>
> Do we have any concerns with discrepancies between `<wa-animation>` and the FA icon animation API? They use the same words to mean different things, and `<wa-animation>` has more preset animation options, but not the same level of customization as FA icons.

Having worked on the FA animations in the past, I'm curious what you think @talbs. Mind giving this PR a review when you have a chance? 🙏

packages/webawesome/src/components/icon/icon.styles.ts
```diff
@@ -41,4 +42,217 @@ export default css`
...
+
+  :host([effect='spin']) {
+    animation-name: spin;
+    animation-delay: var(--animation-delay, 0s);
+    animation-direction: var(--animation-direction, normal);
+    animation-duration: var(--animation-duration, 2s);
+    animation-iteration-count: var(--animation-iteration-count, infinite);
+    animation-timing-function: var(--animation-timing, linear);
+  }
+
```
talbs, line 125: FYI, If we're aiming for parity with Font Awesome's animations, we're missing [`spin-reverse`](https://github.com/FortAwesome/Font-Awesome/blob/7.x/css/all.css#L249).

☝️ I'll add this as part of a follow-up PR.

packages/webawesome/src/components/icon/icon.ts
```diff
@@ -84,6 +107,15 @@ export default class WaIcon extends WebAwesomeElement {
   /** The name of a registered custom icon library. */
   @property({ reflect: true }) library = 'default';
 
+  /** Sets the rotatiton degree of the icon */
```
talbs, line 110:
```suggestion
  /** Sets the rotation degree of the icon */
```

NIT: Small typo here.

☝️ I'll tackle this as part of a follow-up PR.

packages/webawesome/src/components/icon/icon.ts
```diff
@@ -22,6 +22,8 @@ interface IconSource {
   fromLibrary: boolean;
 }
 
+export type IconEffect = 'beat' | 'fade' | 'beat-fade' | 'bounce' | 'flip' | 'shake' | 'spin' | 'spin-pulse';
```
talbs, line 25: After a team discussion, we'll want to use the term `animation` in place of `effect`. 

☝️ I'll tackle this as part of a follow-up PR.

packages/webawesome/src/components/icon/icon.ts
```diff
@@ -38,6 +40,27 @@ interface IconSource {
  * @cssproperty [--primary-opacity=1] - Sets a duotone icon's primary opacity.
  * @cssproperty [--secondary-color=currentColor] - Sets a duotone icon's secondary color.
  * @cssproperty [--secondary-opacity=0.4] - Sets a duotone icon's secondary opacity.
+ * @cssproperty [--animation-delay=0] Sets when the animation will start
+ * @cssproperty [--animation-direction=normal] Defines whether or not the animation should play in reverse on alternate cycles
+ * @cssproperty [--animation-duration=1s] Defines the length of time that an animation takes to complete one cycle
+ * @cssproperty [--animation-iteration-count=infinite] Defines the number of times an animation cycle is played
+ * @cssproperty [--animation-timing] Describes how the animation will progress over one cycle of its duration
+ * @cssproperty [--beat-scale] Set max value that an icon will scale
```
talbs, line 48: NIT: For single animation-specific properties, we'll want to note the specific animation in their descriptions.

☝️ I'll tackle this as part of a follow-up PR.

talbs (approved):
@TUEgemenKancal, thanks for the Awesome (small pun intended) contribution!

This is solid work that brings the icon component closer to having parity with Font Awesome's Styling Toolkit.

As @claviska noted, I led Font Awesome's animations and styling utilities. From that perspective, this port looks accurate. 

I noted a few places where some follow-ups are needed. **But I'll tackle those in a subsequent PR** so that we can get your good work landed. 

Please pass along my thanks and kudos to any peers who were also involved in this work. We appreciate you, them, and the help!

PR #1831, opened 2025-11-26
wa-dialog docs - clarify prevent dialog closing documentation

ajmchambers:
Updated the dialog documentation to match newer default behaviour (non light-dismiss) and to describe the example more accurately.


claviska (approved):
Thank you!

Issue #106, opened 2024-05-13
Disabled link buttons aren't actually disabled


claviska:

claviska:
fixed in c4a084ee41d0dca6e6cf6125ccc2cba7d9acdaca

PR #1848, opened 2025-12-03
Support disabled link buttons

roydukkey:
#106 doesn't give much context. I assume effort was allocated to higher priority items, nevertheless, this does seem to cover support for disabled link buttons.

### DIFF
--- a/packages/webawesome/src/components/button/button.ts
+++ b/packages/webawesome/src/components/button/button.ts
@@ -72,7 +72,7 @@ export default class WaButton extends WebAwesomeFormAssociatedElement {
   /** Draws the button with a caret. Used to indicate that the button triggers a dropdown menu or similar behavior. */
   @property({ attribute: 'with-caret', type: Boolean, reflect: true }) withCaret = false;
 
-  /** Disables the button. Does not apply to link buttons. */
+  /** Disables the button. */
   @property({ type: Boolean }) disabled = false;
 
   /** Draws the button in a loading state. */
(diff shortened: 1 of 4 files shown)


packages/webawesome/src/components/button/button.ts
```diff
@@ -235,7 +235,9 @@ export default class WaButton extends WebAwesomeFormAssociatedElement {
 
   /** Simulates a click on the button. */
   click() {
-    this.button.click();
+    if (!this.disabled) {
+      this.button.click();
+    }
```
claviska: This only blocks calls to the button's `click()` method.

claviska (changes requested):
I think the main requirements for this are:

- The link button should show a `not-allowed` cursor like disabled buttons currently do
- The user must not be able to activate the link through any combination clicking or CMD/SHIFT+clicking or using the keyboard

I'd probably suggest adding `inert` to the internal `<a>` to achieve this, but we also need to make sure the host element isn't dispatching a `click` event when this happens.

packages/webawesome/src/components/button/button.ts
roydukkey: 227236b

claviska (approved):
I'm going to merge this as-is, but it looks like `inert` actually _isn't_ the best for native parity. When `inert` is applied, screen reader virtual cursors skip the element entirely. I'm going to revert that line and prevent the click with JS instead so the experience is the same for screen readers.

Other notes to myselt:

- The `aria-disabled` attribute should only be applied to links, not button since it already has `disabled`
- Add changelog entry

claviska:
See https://github.com/shoelace-style/webawesome/pull/2006 for the follow up. Thanks for helping to get this over the finish line, and apologies for the misdirection. Once I tested `inert` I remembered it makes it invisible to assistive devices as well, which isn't what we want :)

PR #1863, opened 2025-12-09
Expose `--indicator-linecap` for progress-ring

roydukkey:
Adds feature suggested in #1521.

Screenshot 2025-12-09 at 13 46 02

### DIFF
--- a/packages/webawesome/docs/docs/components/progress-ring.md
+++ b/packages/webawesome/docs/docs/components/progress-ring.md
@@ -27,6 +27,14 @@ Use the `--track-width` and `--indicator-width` custom properties to set the wid
 <wa-progress-ring value="50" style="--track-width: 6px; --indicator-width: 12px;"></wa-progress-ring>
 ```
 
+### Indicator Linecap
+
+Use the `--indicator-linecap` custom property to set shape at the ends of the progress ring's indicator.
+
+```html {.example}
+<wa-progress-ring value="10" style="--track-width: 50%; --indicator-width: 50%; --indicator-linecap: butt"></wa-progress-ring>
+```
+
 ### Colors
 
 To change the color, use the `--track-color` and `--indicator-color` custom properties.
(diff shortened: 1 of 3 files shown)


claviska:
I wonder if we should make the public API something like:

```html
<wa-progress-ring appearance="ring | pie">
```

It's not obvious to me that setting `--indicator-linecap` would result in a pie chart-style effect, so I'm wondering if it's too low level. @lindsaym-fa any thoughts on this one?

lindsaym-fa:
@claviska I agree that `--indicator-linecap` isn't the clearest option for a pie-style progress ring, but ultimately, I like its simplicity and usefulness beyond just creating a pie shape. I can see the use case for flat line caps with a typical progress ring too — not just pie. (Aside: I think `--indicator-linecap` would be awesome to have for `<wa-progress-bar>` too, though obviously there would be refactoring work to support that.)

`appearance="ring | pie"` definitely makes the pie shape easier, but it makes me question why we wouldn't have a single `<wa-progress>` component that supports `appearance="bar | ring | pie"` in that case. There's as much potential overlap between bar and ring as there is between ring and pie, as I see it. That'd be a breaking change and a larger discussion for some other time, but it gives me pause on if `<wa-progress-ring>` with `appearance="ring | pie"` is the right call.

Exposing `--indicator-linecap` seems to have the lowest impact on the API with the highest flexibility, though I definitely admit it's more error-prone for creating a pie shape than I'd prefer. Case in point: Setting `--track-width` and `--indicator-width` to 50% as shown in the example isn't a reliable way to create the shape since Safari doesn't get along with percentage values for those properties.

Here's how the example renders in Safari:
Screenshot 2025-12-29 at 2 49 35 PM

Setting those properties to something like `calc(var(--size) / 2)` would work great here instead. I'd recommend showing two examples in the docs for this: one with two progress rings showing the default `round` and `butt` linecaps as an unadulterated demonstration of the options, and a separate example showing how a combination of `--track-width`, `indicator-width`, and `--indicator-linecap` can be used to create a pie shape.

claviska:
One more thought...we could move bar, ring, and pie into a single `<wa-progress appearance="...">` and deprecate alias `<wa-progress-ring>` in non-breaking way if we wanted to move in that direction.

claviska:
I think we want to move towards a single `<wa-progress>` element. For now, I've added two parts to the SVG circles that allow you to achieve the output you're looking for. See #1954 for details.

PR #1864, opened 2025-12-09
Add dock panel

timkpaine:
Ref: https://github.com/shoelace-style/webawesome/discussions/1834

Also fixes a small issue in the markdown processing that I ran into with the docs


claviska:
This is a really cool prototype! Thanks for submitting it. I think we'd like to see a more declarative approach where the panels are in the DOM. We typically reserve JS-based APIs for things such as charts and data grid where it would be less ergonomic and performant to mold the data into a DOM structure.

That said, this is a really inspiring demo and, when we get to a dock panel-style component, we'll definitely draw some inspiration from this!

In the meantime, I'm going to close it out to keep our PR list up to date, but we can still reference it and access it.

Cheers!

PR #1865, opened 2025-12-10
Revert `.css` => `.styles.ts`

### LINKED ISSUE #1812: CSS Importing technique is defeating all stylesheet sharing
### Describe the bug

Shared stylesheets are duplicated in every module that imports them. This increases memory overhead and reduces performance of the elements.

Styles in components are imported with non-standard CSS imports, which are provided as a string via a build tool.  Example:

https://github.com/shoelace-style/webawesome/blob/09c8f02f794c05852115b6c1c095011c971256ed/packages/webawesome/src/components/button/button.ts#L16

These styles are then converted into a stylesheet using `unsafeCSS()`:

https://github.com/shoelace-style/webawesome/blob/09c8f02f794c05852115b6c1c095011c971256ed/packages/webawesome/src/internal/webawesome-element.ts#L29-L32

This means that every time a CSS file is used in a component definition it is parsed into a unique stylesheet object, instead reusing one stylesheet object across all definitions like you would get with Lit's `css` tag or native CSS modules.

### To Reproduce

1. Go to a page with multiple WA components, like https://webawesome.com/docs/components/tab-group/
2. Open devtools, select the Memory tab, and take a heap snapshot
3. Search for CSSStyleSheet in the Filter by class input
4. See **224** individual CSSStyleSheet objects.

There should only be one CSSStyleSheet for each .css file imported into the components, but what we're seeing is one CSSStyleSheet for each use in the static `css` array. Files like `host.css` are duplicated for every component class.

### Additional information

The solution is to either convert the .css files to JS modules and use Lit's `css` tag in the imported modules, or to use native CSS Module syntax with a build step that supports them, since native modules produce a shared stylesheet. One such build tool is my Rollup plugin: https://www.npmjs.com/package/rollup-plugin-css-modules There may be others for other build tools.

claviska:
This PR fixes #1812 by reverting the component CSS files to Lit's preferred syntax. Since this updates the base class, I'll update pro after this is merged.


lindsaym-fa (changes requested):
`styles/utilities/size.css` and `styles/utilities/variants.css` were both imported in native styles, so these are the results we're getting for native buttons and such now:

Image

I'd recommend:
- Moving `size.styles.ts` and `variants.styles.ts` to the `styles/component/` directory
- Restoring `size.css` and `variants.css` in the `styles/utilities/` directory and removing the `:host()` selector for each rule in these files

Does that make sense to you, @claviska? (And let me know if you'd like me to make those changes.)

claviska:
@lindsaym-fa nice catch. It looks like the visually hidden utilities were also affected so I updated those as well. Let me know if you see anything else!

claviska:
Note: I also removed the `wa-utilities` layer from the component CSS in [d75fc9e](https://github.com/shoelace-style/webawesome/pull/1865/commits/d75fc9e1fa385fc8bf2c720caa469660c4725f86).

lindsaym-fa (commented):
One small comment. All else looks good!

packages/webawesome/src/styles/utilities/visually-hidden.css
```diff
@@ -1,16 +1,14 @@
-@layer wa-utilities {
```
lindsaym-fa, line 1: This stylesheet should still have the `@layer` rule, right?

claviska: 🦅 👁️

claviska: Fixed in [1be9cbd](https://github.com/shoelace-style/webawesome/pull/1865/commits/1be9cbd51b1f1ebc97ce36a907146d9327f43765)

lindsaym-fa (approved):

PR #1895, opened 2025-12-19
porting shoelace framework docs to web awesome (and CEM goodies)

### LINKED ISSUE #1645: "name" attribute is missing from WaInputProps type
### Describe the bug
See title. It's also missing from the [`wa-input` documentation](https://webawesome.com/docs/components/input#attributes-and-properties).

### To Reproduce

1. Given this TypeScript code:
```ts
<wa-input type="text" label="Name" name="name"></wa-input>
```
2. Observe following error:
```
Type '{ type: "text"; label: string; name: string; }' is not assignable to type 'Partial<WaInputProps & BaseProps<WaInput>>'.
  Property 'name' does not exist on type 'Partial<WaInputProps & BaseProps<WaInput>>'.ts(2322)
```

### Additional information
@KonnorRogers said: "may need to update the manifest generator. may not be searching up the inheritance chain."

KonnorRogers:
Big shoutout to @break-stuff for making this PR as good as it is and the work in the CEM Toolkit.

Fixes #1645


break-stuff:
I'll take a look and see if I can get those other things working for you.

Technically you may not need the VS Code and JetBrains plugin since the language server is better.

packages/webawesome/scripts/make-react.js
```diff
@@ -22,6 +22,7 @@ const components = getAllComponents(metadata);
 const index = [];
 
 for await (const component of components) {
+  if (!component.tagName) { continue }
```
claviska, line 25: Why wouldn't a tag name exist? Could probably use a comment here.

packages/webawesome/custom-elements-manifest.js
```diff
@@ -22,11 +26,23 @@ function replace(string, terms) {
...
+  overrideModuleCreation({ts, globs}) {
+    const program = getTsProgram(ts, globs, "tsconfig.json");
+    return program
+      .getSourceFiles()
+      .filter((sf) => globs.find((glob) => sf.fileName.includes(glob)));
+  },
+
   plugins: [
+    typeParserPlugin({
+    }),
```
claviska:
```suggestion
    typeParserPlugin(),
```

claviska (dismissed):
No blockers.

packages/webawesome/scripts/make-react.js
KonnorRogers: WebAwesomeFormAssociated and WebAwesomeElement do not have a tagName, but extend HTMLElement / LitElement

claviska: Got it. These get passed in from the CEM but are just base classes.

Issue #1905, opened 2025-12-27
TypeError: Cannot read properties of undefined (reading 'clientX') in DraggableElement.handleDragStop


medmunds:
### Describe the bug
[DraggableElement.handleDragStop](https://github.com/shoelace-style/webawesome/blob/2b69fa74a57f00bb11a07e2393fb7adf9737f869/packages/webawesome/src/internal/drag.ts#L107-L109) can cause a TypeError trying to access the empty event.touches array found in most touchend events:

```text
TypeError: Cannot read properties of undefined (reading 'clientX')
    at this.handleDragStop (../../node_modules/@awesome.me/webawesome/dist/chunks/chunk.OWHBNFYI.js:49:61)
    at sentryWrapped (../../node_modules/@sentry/browser/build/npm/esm/prod/helpers.js:93:17)
```

```typescript
// webawesome/src/internal/drag.ts
  private handleDragStop = (event: PointerEvent | TouchEvent) => {
    const clientX = 'touches' in event ? event.touches[0].clientX : (event as PointerEvent).clientX;
                                               ^^^^^^^^^^^^^^^^^^
```

### To Reproduce
Unclear exactly how to reproduce (reported via Sentry). I've managed to reproduce it myself (once) on an Android phone by touch-dragging the thumb of a horizontal wa-slider within a vertically scrolling area, and simultaneously tapping my other fingers elsewhere on the screen. But I don't have exact steps.

The code above is clearly incorrect: in a single-finger touchend event, `touches` will be empty because it includes only the fingers that are still down. (`changedTouches` would be the better choice.)

handleDragStop is installed for both pointerup and touchend events. Since this code isn't constantly throwing TypeErrors on touch devices, I suspect maybe pointerup is usually delivered before touchend. When handleDragStop is invoked on pointerup it removes itself for both events, and the invalid `touches[0]` access never runs. The trick to reproduce would be getting a touchend before (or instead of) a pointerup.

### Browser / OS
\@awesome.me/webawesome 3.1.0

At least these:
 - Chrome Mobile 142.0 on Android
 - Chrome Mobile 143.0.7499.146 on Android 16, Pixel 7

### Additional information
The handleDragStop code could avoid the TypeError by examining event.changedTouches instead.

There are a couple of other likely related issues in DraggableElement:
[...]

PR #1908, opened 2025-12-29
Fix drag bug

claviska:
Fixes the drag bug described in #1905. (This utility is used by a number of components.)

@medmunds Can you confirm this fixes it on Android?


medmunds:
> @medmunds Can you confirm this fixes it on Android?

@claviska I'll try, but I still don't really know reliable steps to reproduce.

By inspection, it should prevent the TypeError on touchend. If you wanted to be extra safe you might add a null check on the array access: `event.changedTouches[0]?.clientX ?? 0`. (Especially borrowing the dragStop code for cancel: I *think* touchcancel should always have at least one changedTouches item, but I can't find that documented anywhere.)

Incidentally, I strongly suspect the DraggableElement code may not behave as expected when multiple touches are involved, per the MDN [TouchEvent.touches](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches#:~:text=Note%3A%20Touches%20inside%20the%20array%20are%20not%20necessarily%20ordered%20by%20order%20of%20occurrences%20(the%20i%2Dth%20element%20in%20the%20array%20being%20the%20i%2Dth%20touch%20that%20happened).%20You%20cannot%20assume%20a%20specific%20order.%20To%20determine%20the%20order%20of%20occurrences%20of%20the%20touches%2C%20use%20the%20touch%20object%20IDs.) docs:

> **Note:** Touches inside the array are not necessarily ordered by order of occurrences (the i-th element in the array being the i-th touch that happened). You cannot assume a specific order. To determine the order of occurrences of the touches, use the touch object IDs.

(But I don't have a specific bug report for that.)

claviska:
> Incidentally, I strongly suspect the DraggableElement code may not behave as expected when multiple touches are involved

I think that's fine, as the utility intends to only track simple (i.e. one pointer) drags. If we were to expand it, we'd definitely want to account for multiple pointers.

PR #1913, opened 2025-12-30
Fix popover and dropdown bugs

### LINKED ISSUE #1910: InvalidStateError: Failed to execute 'showPopover' on 'HTMLElement': Invalid on disconnected popover elements.
### Describe the bug
If a wa-dropdown is removed while it is in the process of opening, [WaPopup.start tries to call showPopover](https://github.com/shoelace-style/webawesome/blob/2b69fa74a57f00bb11a07e2393fb7adf9737f869/packages/webawesome/src/components/popup/popup.ts#L293) on a now-disconnected element, resulting in:

```text
chunk.VIY2CCPQ.js:1644 Uncaught (in promise) InvalidStateError: Failed to execute 'showPopover' on 'HTMLElement': Invalid on disconnected popover elements.
    at WaPopup.start (chunk.VIY2CCPQ.js:1644:29)
    at WaPopup.updated (chunk.VIY2CCPQ.js:1610:14)
    at WaPopup._$AE (chunk.HAXJJGQX.js:248:128)
    at WaPopup.performUpdate (chunk.HAXJJGQX.js:243:16)
    at WaPopup.scheduleUpdate (chunk.HAXJJGQX.js:221:17)
    at WaPopup._$EP (chunk.HAXJJGQX.js:217:21)
```

### To Reproduce
I haven't been able to find the real-world steps that would reproduce this. (I think it involves some hard-to-replicate timing.)

But this code reproduces it reliably:
```typescript
dropdown.open = true;
dropdown.remove();
```

### Demo

https://codepen.io/mvedmunds/pen/myEJqoa?editors=0010

### Browser / OS
any

Web Awesome 3.1.0

### Additional information
Suggested fix (also protects against `@query popup` returning null): 

```patch
webawesome/src/components/popup/popup.ts

   private start() {
     // We can't start the positioner without an anchor
-    if (!this.anchorEl || !this.active) {
+    if (!this.anchorEl || !this.active || !this.isConnected) {
       return;
     }
 
-    this.popup.showPopover?.();
+    this.popup?.showPopover?.();

     this.cleanup = autoUpdate(this.anchorEl, this.popup, () => {
       this.reposition();
     });
   }
```

### LINKED ISSUE #1911: TypeError: Cannot read properties of null (reading 'assignedElements' / 'hidePopover')
### Describe the bug

WaPopup and WaDropdown (and likely other components) don't check for null `@query` property results, which can cause TypeErrors.

I've seen several Sentry reports about this in those two components, likely involving shadow DOM ownership bugs in older browser versions.

### To Reproduce

I'm not sure how users are hitting this. The linked codepen below emulates what I suspect are older browser bugs that reveal the problem.

Both [dropdown](https://github.com/shoelace-style/webawesome/blob/2b69fa74a57f00bb11a07e2393fb7adf9737f869/packages/webawesome/src/components/dropdown/dropdown.ts#L142) and [popup](https://github.com/shoelace-style/webawesome/blob/2b69fa74a57f00bb11a07e2393fb7adf9737f869/packages/webawesome/src/components/popup/popup.ts#L300-L303) include code that doesn't check for null `@query` values:

```typescript
// webawesome/src/components/dropdown/dropdown.ts
  @query('slot:not([name])') defaultSlot: HTMLSlotElement;
  // ...
  private getItems(includeDisabled = false): WaDropdownItem[] {
    const items = this.defaultSlot
                      ^^^^^^^^^^^^
      .assignedElements({ flatten: true })
      .filter(el => el.localName === 'wa-dropdown-item') as WaDropdownItem[];
```

```typescript
// webawesome/src/components/popup/popup.ts
  @query('.popup') popup: HTMLElement;
  // ...
  private async stop(): Promise<void> {
    return new Promise(resolve => {
      this.popup.hidePopover?.();
          ^^^^^^
```

Here are the stack traces from Sentry:

```text
TypeError: Cannot read properties of null (reading 'assignedElements')
    at WaDropdown.getItems (../../node_modules/@awesome.me/webawesome/dist/chunks/chunk.WNWBLNT7.js:265:36)
    at WaDropdown.closeAllSubmenus (../../node_modules/@awesome.me/webawesome/dist/chunks/chunk.WNWBLNT7.js:301:24)
    at WaDropdown.disconnectedCallback (../../node_modules/@awesome.me/webawesome/dist/chunks/chunk.WNWBLNT7.js:234:10)
    at HeadMatter.restoreHead (../../src/head-matter.ts:76:12) ...
```

```text
TypeError: Cannot read properties of null (reading 'hidePopover')
    at <anonymous> (../../node_modules/@awesome.me/webawesome/dist/chunks/chunk.7AACWKNJ.js:177:18)
    at new Promise (<anonymous>)
    at WaPopup.stop (../../node_modules/@awesome.me/webawesome/dist/chunks/chunk.7AACWKNJ.js:176:12)
    at WaPopup.disconnectedCallback (../../node_modules/@awesome.me/webawesome/dist/chunks/chunk.7AACWKNJ.js:130:10)
    at HeadMatter.restoreHead (../../src/head-matter.ts:76:12) ...
```

(HeadMatter is my own component that manipulates DOM in document.head following a slotchange. It does *nothing* involving wa-dropdown or wa-popup. But it does have a *sibling* wa-dropdown component (which it doesn't touch), making me suspect browser bugs with confusion over shadow DOM ownership.)

### Demo

https://codepen.io/mvedmunds/pen/PwzqPZB?editors=0010

(Just to be clear: this codepen deliberately manipulates the DOM in invalid ways in an attempt to simulate presumed browser bugs. I'm not suggesting it's a valid use case, but it does exactly reproduce the stack traces recorded by Sentry.)

### Browser / OS

Web Awesome 3.1.0

Problems recorded in (exact versions unknown due to user-agent reduction):
- Chrome >= 131.0 on Windows >= 10
- Chrome >= 123.0 on Linux

### Additional information

Besides the odd behavior above, `@query` can legitimately return null if accessed before the first render. The most general fix would probably be to mark all query properties as possibly null and then address the resulting TypeScript complaints. E.g.:

```patch
-  @query('.popup') popup: HTMLElement;
+  @query('.popup') popup?: HTMLElement;
 // ...
-       this.popup.hidePopover?.();
+       this.popup?.hidePopover?.();
```

claviska:
- Fixes #1910
- Fixes #1911


packages/webawesome/src/components/dropdown-item/dropdown-item.ts
```diff
@@ -219,7 +219,7 @@ export default class WaDropdownItem extends WebAwesomeElement {
       await animateWithClass(this.submenuElement, 'hide');
       this.submenuElement.hidden = true;
       this.submenuElement.removeAttribute('data-visible');
-      this.submenuElement.hidePopover();
+      this.submenuElement.hidePopover?.();
     }
```
medmunds: this.submenuElement could become null during the await animateWithClass. (The problem isn't that submenuElement still exists but has lost its hidePopover method; the problem is querySelector('#submenu') no longer finds that element after some DOM cleanup or corruption. And any async call can let that change slip into code that otherwise looks safe. Or at least, that's what the Sentry logs seemed to show.)

claviska: Ah, `submenuElement` is only available when a submenu actually exists. Good catch!

claviska: Actually, I think this is fine since there's a guard right above this that exists when the submenu doesn't exist.

claviska: Cancel. I missed the `await`. Fixed in f3be2c4

PR #1915, opened 2025-12-30
Ensure consistent input heights in flex/grid containers

### LINKED ISSUE #1876: Inconsistent height between input and select when using wa-grid
### Describe the bug
* Framework: Vue.js 3
* Build tool: Vite 7

When an input and a select are aligned on the same row using the wa-grid layout, their heights are inconsistent.

* The input has a height of 46px
* The select has a height of 43px

This visual mismatch breaks alignment and consistency in forms.

### To Reproduce

1. Create a layout using wa-grid
2. Place an input and a select on the same row
3. Observe their rendered heights

### Demo

[Playground](https://stackblitz.com/edit/vitejs-vite-ukfod4qs?file=src%2FApp.vue)

### Screenshots

Image

### Browser / OS
 - OS: Mac 26.1
 - Browser: Chrome (but same bug in Firefox)
 - Browser version: 143

### LINKED ISSUE #2004: wa-input and wa-select has different heights
### Describe the bug
A default wa-input and wa-select has different heights.

### To Reproduce

1. Add a wa-input and wa-select in horisontal layout
2. See the height difference.

### Demo

https://codepen.io/fredrik-sogaard/pen/dPXeOwz?editors=1000

Image

lindsaym-fa:
Fixes two styles that caused inconsistent input, select, and combobox heights in flex/grid containers (example in #1876):

1. Adds `display: contents` to `form-control` parts to ignore the wrapper for styling and instead expose its child elements (label, input, and hint) to the flex rules applied to the host element
2. Removes `flex: auto` from `<wa-input>`'s internal `.text-field` to ensure the field always respects its given `height`

---

### 'Shots

**Before**
Screenshot 2025-12-30 at 4 31 37 PM

**After**
Screenshot 2025-12-30 at 4 30 46 PM

---

### Explanation

There's a bit of a complicated interaction of CSS layout models here. Here's what I understand to be true:

`form-control.styles.ts` applies `display: flex` and `flex-direction: column` to the host element for every form control to consistently arrange labels, inputs, and hints. However, some components like `<wa-select>`, `<wa-combobox>`, and `<wa-color-picker>` have an additional internal `part="form-control"` wrapper, which prevents wrapped labels, inputs, and hints from being included in the host's flex layout.

Notably, this means that labels follow a typical block document flow and rest on the text baseline — whereas flexbox alignment isn't based on text lines — causing a smidge of extra whitespace above the label. With this extra whitespace, components like `<wa-select>` have a slightly inflated height compared to `<wa-input>`.

Meanwhile, `flex: auto` is applied to the text field in `<wa-input>`, and this shorthand assumes `flex-grow: 1`. This means that, given free space in a flex container — whether that space is vertical or horizontal — the text field will grow to occupy the free space.

And, the final piece of the puzzle: utilities like `.wa-grid` assume `align-items: stretch`, which cause the grid items within to stretch vertically to fill the entire available height for the grid row. The same is true for flex containers with `align-items: stretch`.

So, with `<wa-select>`'s modicum of additional height, `<wa-input>`'s growable text field, and `.wa-grid`'s directive for stretchy grid items, `<wa-input>`'s text field stretches a few pixels so that the host element matches the height of its neighboring `<wa-select>`.

### DIFF
--- a/packages/webawesome/src/components/input/input.styles.ts
+++ b/packages/webawesome/src/components/input/input.styles.ts
@@ -6,7 +6,6 @@ export default css`
   }
 
   .text-field {
-    flex: auto;
     display: flex;
     align-items: stretch;
     justify-content: start;
--- a/packages/webawesome/src/styles/component/form-control.styles.ts
+++ b/packages/webawesome/src/styles/component/form-control.styles.ts
@@ -1,7 +1,8 @@
 import { css } from 'lit';
 
 export default css`
-  :host {
+  :host,
+  [part~='form-control'] {
     display: flex;
     flex-direction: column;
   }
--- a/packages/webawesome/src/styles/component/form-control.styles.ts
+++ b/packages/webawesome/src/styles/component/form-control.styles.ts
@@ -1,10 +1,14 @@
 import { css } from 'lit';
 
 export default css`
-  :host,
-  [part~='form-control'] {
+  :host {
     display: flex;
     flex-direction: column;
+  }
+
+  /* Treat wrapped labels, inputs, and hints as direct children of the host element */
+  [part~='form-control'] {
+    display: contents;
   }
 
   /* Label */


packages/webawesome/src/components/input/input.styles.ts
```diff
@@ -6,7 +6,6 @@ export default css`
   }
 
   .text-field {
-    flex: auto;
```
lindsaym-fa, line 9: Here's a different (and especially egregious) example of the problem that removing this line is meant to fix. These screenshots show how `<wa-input>` changes in height with a taller component like `<wa-card>` as a neighboring grid item:
**Before**
Image
**After**
Image

claviska (approved):

PR #1923, opened 2026-01-05
chore(translation): update persian translation

alimd:
As a native Persian speaker, I identified and addressed the shortcomings of the Persian translation, striving to enhance its comprehensibility and appropriateness within my native language.


claviska (approved):
Thank you!

PR #1924, opened 2026-01-06
Add name property to input component

ouvreboite:
Should solve https://github.com/shoelace-style/webawesome/issues/1645


KonnorRogers:
Thanks for the PR, unfortunately it kind of puts a band-aid on the much bigger problem of inheritance not working as expected.

this is actually solved in this PR:

https://github.com/shoelace-style/webawesome/pull/1895

in particular this line here:

https://github.com/shoelace-style/webawesome/pull/1895/changes#diff-2c4d3ab7adcdf52782e57ffc7acdc16fcee6a5c1f3966f1e61d709d2d8e768e8R29-R30

PR #1928, opened 2026-01-09
Remove extra period in tooltip docs

alexhrao:
This removes the second period in the docs, which gets reflected on the website

I didn't create an issue for this first because of how trivial it is, which seems to be allowed in the contribution guidelines. That being said if I should open one anyway, please let me know and I'd be happy to do so


claviska (approved):
Thanks!

PR #1931, opened 2026-01-09
Add missing `.wa-gap-4xl` utility class

lindsaym-fa:
Adds missing `.wa-gap-4xl` utility to account for our existing `--wa-space-4xl` design token. (_D'oh!_)


talbs (approved):
tgs-so-good

---

Great to have this - thanks for adding it. I'll scour our site/app to see where we can use it instead of any bespoke custom CSS rules.

PR #1932, opened 2026-01-10
Added a fix to support v-model within Vue for the Input component.

deathlynocturnal:
After messing about with Vue, I have added a tweak to the `handleInput` code within the Input component, this seems to have been done with the Textarea component, but not the Input. I have tested this by amending my local code and Vue's `v-model` works as expected.


deathlynocturnal:
I am going to close this pull request, because my environment was not working with `v-model` but for some reason I can't replicate that anymore. I will re-open if I can get it to happen again!

PR #1930, opened 2026-01-09
Add `justify-content` utility classes

lindsaym-fa:
Adds `justify-content` CSS utilities for controlling how space is distributed along the main axis in flex and grid containers. 

- Adds six new utility classes for `justify-content` (start, end, center, space-around, space-between, space-evenly)
  - Note: I've opted to exclude `justify-content: stretch` since it has no useful effect when used alongside our existing layout utilities. In flex containers, it behaves the same as `start`, and because `.wa-grid` already sizes grid items to fill the available space along the main axis, it has no effect. That said, I'm open to adding it for completeness and usefulness in systems outside of WA.
- Updates existing layout utility classes to give their default `justify-content` properties zero specificity
- Adds Justify Content documentation with visual previews
- Updates Align Items documentation for improved clarity and consistency

PR #1935, opened 2026-01-12
Abstract + Visually Sync Up Styling in Layout Docs Examples

talbs:
This work is a follow up to https://github.com/shoelace-style/webawesome/pull/1930 and aims to visually sync up the styling we use in examples that explain our Layout-based styling utilities. 

It also abstracts some of the common styling values into CSS custom properties for easier and more consistent use across each doc.


lindsaym-fa (approved):
Great improvement! Thanks!

PR #1936, opened 2026-01-12
Set default cursor for empty `<wa-tree-item>` expand buttons

leonvogt:
I've noticed that hovering over a `<wa-tree-item>` with no children still results in a pointer cursor. This PR addresses that issue.

**Before**

https://github.com/user-attachments/assets/785cdcee-1d05-446e-b807-2dca4e073f78

---

**After**

https://github.com/user-attachments/assets/6b5528b4-666e-480b-96a6-8dd86f3e8550


claviska (approved):
Thanks for the PR! Great attention to detail. I'll submit a changelog for this.

PR #1938, opened 2026-01-13
Unlist unreleased utilities from the docs

lindsaym-fa:
This PR  hides unreleased utilities from the documentation that are pending a next minor version release.
- Marked the "Justify Content" utility page as unpublished and unlisted to hide it from production builds
- Commented out the wa-gap-4xl entry in the Gap utility documentation
- Removed the "Justify Content" link from the sidebar navigation


talbs (approved):
Looks good!

PR #1943, opened 2026-01-13
Apply flexbox by default with `wa-align-items-*` and `wa-justify-content-*` utils

lindsaym-fa:
Because the `align-items` and `justify-content` properties don't apply to block-level boxes, our `wa-align-items-*` and `wa-justify-content-*` utility classes don't have any effect on elements outside of flexbox, grid, or multicol formatting contexts.

To ensure these utilities can be used on their own and still produce meaningful results, this PR adds `display: flex` to each of these classes with 0 specificity (using `:where()`). We already take the same approach with our `wa-gap-*` utilities.


claviska (dismissed):

talbs (dismissed):
This seems worth it to me and is well-scoped. I say go for it!

Also, 💯 x 💯 for this PR's branch name! Instant 🏆!

PR #1946, opened 2026-01-13
Add missing layout utility descriptions

lindsaym-fa:
The descriptions for each of our layout utility classes seem to have gone missing at some point. This PR reintroduces these much-needed descriptions to the docs.


talbs (approved):
Thanks for bringing these back!

PR #1958, opened 2026-01-15
Support selecting frame document source in code examples

kkoop64:
## Description
Support `data-select-src` on `<wa-zoomable-frame>` in code examples to make the relevant code visible/copyable.
* A boolean attribute shows the full source document.
* A string attribute shows the source document elements matching the given CSS selectors.
* Works with `src` and `srcdoc` attributes.

### Current Behavior

Screenshot 2026-01-15 at 6 41 37 PM

#### Select full document source:

```html {.example}
<wa-zoomable-frame
  srcdoc="<!doctype html>
<html>
  <body>
    <h1>Hello, World!</h1>
    <p>This is inline content.</p>
  </body>
</html>"
  data-select-src
></wa-zoomable-frame>
```

Screenshot 2026-01-15 at 6 40 23 PM

#### Select `<body>` from source to focus on the relevant code:

```html {.example}
<wa-zoomable-frame
  srcdoc="<!doctype html>
<html>
  <body>
    <h1>Hello, World!</h1>
    <p>This is inline content.</p>
  </body>
</html>"
  data-select-src="body"
></wa-zoomable-frame>
```

Screenshot 2026-01-15 at 6 39 11 PM

See `src`-based examples in https://github.com/shoelace-style/webawesome-pro/pull/127.

### DIFF
--- a/packages/webawesome/docs/_transformers/code-examples.js
+++ b/packages/webawesome/docs/_transformers/code-examples.js
@@ -1,5 +1,5 @@
+import { readFileSync } from 'fs';
 import { parse } from 'node-html-parser';
-import { v4 as uuid } from 'uuid';
 import { copyCode } from './copy-code.js';
 import { highlightCode } from './highlight-code.js';
 
@@ -25,7 +25,8 @@ export function codeExamplesTransformer(options = {}) {
       const hasButtons = !code.classList.contains('no-buttons');
       const isOpen = code.classList.contains('open') || !hasButtons;
       const noEdit = code.classList.contains('no-edit');
-      const id = `code-example-${uuid().slice(-12)}`;
+      const uuid = crypto.randomUUID();
+      const id = `code-example-${uuid.slice(-12)}`;
       let preview = pre.textContent;
 
       const langClass = [...code.classList.values()].find(val => val.startsWith('language-'));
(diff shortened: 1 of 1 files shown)


packages/webawesome/docs/_transformers/code-examples.js
```diff
@@ -38,7 +39,37 @@ export function codeExamplesTransformer(options = {}) {
...
+            // Fix parse() formatting of wrapped first attributes and reduce
+            // indentation to match the least-indented line for each fragment
+            const html = fragment.outerHTML
+              .replace(/<([^\s>]+)(\s{2,})(?=[^\s>])/g, (_, tag, spaces) => `<${tag}\n${spaces.slice(1)}`)
+              .split('\n');
+            const lastLine = html[html.length - 1] || '';
+            const indent = new RegExp(`^${lastLine.match(/^\s*/)?.[0] ?? ''}`);
+            source = `${i ? source : ''}${html.map(line => line.replace(indent, '')).join('\n')}\n\n`;
+          });
+        }
```
kkoop64, line 69: If we only ever want to show the full document source, then we can eliminate half this logic. However, I found the extra functionality compelling enough to submit for review. Let me know what you think!

claviska (approved):
I have no concerns about the code, but I don't think I'm seeing the big picture. What does this allow us to do moving forward? (Just for my knowledge! Apologies if ya'll have discussed this already.)

lindsaym-fa:
@claviska I'll jump in here. 😄

> What does this allow us to do moving forward?

Currently, our full-page patterns exist as standalone HTML files. When we present these full-page patterns in the docs, we (rightfully, IMO) use `<wa-zoomable-frame>` to show them off. However, that doesn't allow the source code to be easily copy-and-pasteable, one of the key promises of patterns. Folks instead have to open the patterns in a new tab (which we provide a link for) and copy the markup from the DOM.

As we're adding more full-page patterns in the current split, we wanted to prioritize making sure that the code is easier to grab. We have this in the pitch:
> **Ensure the Markup is Gettable**
..._without_ having to reach into the DOM. Because our page-level patterns get wrapped in zoomable frames that reference the relative path of the file, the markup for the pattern isn't easily reached. 
> 
> We could easily go overboard here, but for now, let's think about adding a download button for each pattern that downloads the corresponding HTML file, or a standalone copy button that grabs the markup.
> 
> Other proposed solutions are welcome, as long as folks can get the code for the pattern in a few clicks.

This solution — allowing us to opt-in to showing the `src` of  `<wa-zoomable-frame>` — gives us some cool benefits:

1. We can leverage our existing code example component (with its built-in resizing, expandable code, and copy button) to show a preview alongside code that is actually useful for folks
2. With the optional string attribute, we can isolate the code to just the pattern itself rather than including all of the HTML boilerplate from the original HTML file

claviska:
Ahh, thanks for the clarification! That's what I get for being sick last week when I'm sure you folks discussed this 😄

talbs (approved):
I can't speak to the code diff in detail, but this works as intended on the UI side of things - I was able to view source code scoped to an identified element and copy that code as well.

A jorb well done, @kkoop64!

PR #1960, opened 2026-01-16
Fix Pro+ family/variants in `<wa-icon>` 

### LINKED ISSUE #1880: Update `<wa-icon>` to FA 7.1.0 and add support for Utility icon packs
The FA Pro+ 7.1.0 icon packs for Utility, Utility Fill, and Utility Duo aren't currently supported in `<wa-icon>`.

- [ ] Update the default library in `<wa-icon>` to 7.1.0
- [ ] Add support for `family="utility"`
- [ ] Add support for available Utility variants: `variant="semibold"`, `variant="duo-semibold"`, and `variant="fill-semibold"`

### LINKED ISSUE #1951: `wa-icon` misinterprets some Pro+ families suffixes as variants
The <wa-icon> component currently misinterprets the name of some Font Awesome Pro+ icon packs when they include multiple families such as duo or fill.

In Font Awesome:
	•	Pro+ family names sometimes include qualifiers like `jelly`, `jelly-duo`, `jelly-fill`
	•	In `FA` the variant maps 1-to-1 with what `FA` calls a style (e.g. regular, solid)

In Web Awesome today:
- `family="jelly-duo" variant="regular"` is NOT recognized
- This leads to incorrect asset URLs being generated and icons failing to load

Instead, Web Awesome expects `family="jelly" variant="duo-regular"`, but that is not the correct mapping to the Font Awesome family/style structure, because there is no "style/variant" named "duo-regular".

This is the correct Font Awesome mapping:
`<wa-icon name="house" family="jelly-duo" variant="regular"></wa-icon>`

Why this is a bug
- It does not reflect how Font Awesome families and styles are actually defined.
- Pro+ packs that contain multiple families fail unless users adopt a non-FA mental model.
- This causes downstream issues for FA-driven UIs that correctly separate family and style.

If a non-breaking fix is required, you could allow both of the following to resolve to the same asset path:
`<wa-icon name="house" family="jelly" variant="fill-regular"></wa-icon>`
`<wa-icon name="house" family="jelly-fill" variant="regular"></wa-icon>`

Both should generate: `/svgs/jelly-fill-regular/house.svg`

This would:
- Preserve backward compatibility
- Correctly support FA’s family naming
- Unblock Pro+ icon packs immediately

claviska:
- Fixes #1951 in a backwards compatible way
- Updates FA to 7.1
- Adds Utility to the Pro+ example
- Note: deprecated family/variant combinations are noted in the comments for removal in 4.0

CleanShot 2026-01-16 at 16 17 28@2x

### DIFF
--- a/packages/webawesome/docs/docs/components/icon.md
+++ b/packages/webawesome/docs/docs/components/icon.md
@@ -397,6 +397,20 @@ If you're a [Font Awesome Pro+ customer](https://fontawesome.com/), you have acc
   </div>
 
   <div class="wa-flank" style="--flank-size: 10ch;">
+    <a href="https://fontawesome.com/icons/packs/utility" target="_blank">Utility</a>
+    <div class="wa-cluster" style="font-size: 1.5em;">
+      <wa-icon family="utility" variant="semibold" name="house"></wa-icon>
+      <wa-icon
+        family="utility-duo"
+        variant="semibold"
+        name="house"
+        style="--secondary-color: skyblue; --secondary-opacity: 0.8;"
+      ></wa-icon>
+      <wa-icon family="utility-fill" variant="semibold" name="house"></wa-icon>
+    </div>
+  </div>
+
+  <div class="wa-flank" style="--flank-size: 10ch;">
     <a href="https://fontawesome.com/icons/packs/whiteboard" target="_blank">Whiteboard</a>
     <div class="wa-cluster" style="font-size: 32px;">
       <wa-icon family="whiteboard" variant="semibold" name="house"></wa-icon>
--- a/packages/webawesome/docs/docs/components/icon.md
+++ b/packages/webawesome/docs/docs/components/icon.md
@@ -412,7 +412,7 @@ If you're a [Font Awesome Pro+ customer](https://fontawesome.com/), you have acc
 
   <div class="wa-flank" style="--flank-size: 10ch;">
     <a href="https://fontawesome.com/icons/packs/whiteboard" target="_blank">Whiteboard</a>
-    <div class="wa-cluster" style="font-size: 32px;">
+    <div class="wa-cluster" style="font-size: 1.5em;">
       <wa-icon family="whiteboard" variant="semibold" name="house"></wa-icon>
     </div>
   </div>
--- a/packages/webawesome/src/components/icon/library.default.ts
+++ b/packages/webawesome/src/components/icon/library.default.ts
@@ -1,7 +1,7 @@
 import { getKitCode } from '../../utilities/base-path.js';
 import type { IconLibrary } from './library.js';
 
-const FA_VERSION = '7.0.1';
+const FA_VERSION = '7.1.0';
 
 function getIconUrl(name: string, family: string, variant: string) {
   const kitCode = getKitCode();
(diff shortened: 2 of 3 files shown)


packages/webawesome/src/components/icon/library.default.ts
```diff
@@ -1,18 +1,24 @@
...
   // Notdog (Pro+)
+  // Correct usage: family="notdog" or family="notdog-duo", variant="solid"
   if (family === 'notdog') {
-    if (variant === 'solid') folder = 'solid';
-    if (variant === 'duo-solid') folder = 'duo-solid';
-    return `https://ka-p.fontawesome.com/releases/v${FA_VERSION}/svgs/notdog-${folder}/${name}.svg?token=${encodeURIComponent(kitCode)}`;
+    // NOTE: variant="duo-solid" is deprecated, use family="notdog-duo" variant="solid" instead
+    if (variant === 'solid') folder = 'notdog-solid';
+    if (variant === 'duo-solid') folder = 'notdog-duo-solid';
+    return `https://ka-p.fontawesome.com/releases/v${FA_VERSION}/svgs/${folder}/${name}.svg?token=${encodeURIComponent(kitCode)}`;
```
lindsaym-fa: Do we need these `return` statements for each Pro+ family? Doesn't the final `return` statement at the end of the `getIconUrl` function take care of this already?

packages/webawesome/src/components/icon/library.default.ts
lindsaym-fa: If we decide to remove these `return` statements, we should just need to specify a folder for any families that don't have one already. e.g. `folder = 'chisel-regular'`

packages/webawesome/src/components/icon/library.default.ts
```diff
@@ -26,18 +32,35 @@ function getIconUrl(name: string, family: string, variant: string) {
...
   if (family === 'jelly') {
-    if (variant === 'regular') folder = 'regular';
-    if (variant === 'duo-regular') folder = 'duo-regular';
-    if (variant === 'fill-regular') folder = 'fill-regular';
-    return `https://ka-p.fontawesome.com/releases/v${FA_VERSION}/svgs/jelly-${folder}/${name}.svg?token=${encodeURIComponent(kitCode)}`;
+    // NOTE: variant="duo-regular" and variant="fill-regular" are deprecated
+    // Use family="jelly-duo" variant="regular" or family="jelly-fill" variant="regular" instead
+    if (variant === 'regular') folder = 'jelly-regular';
+    if (variant === 'duo-regular') folder = 'jelly-duo-regular';
+    if (variant === 'fill-regular') folder = 'jelly-fill-regular';
```
lindsaym-fa: All families _except_ Jelly work perfectly if you only specify `family` and not `variant`. For example: `<wa-icon family="notdog">` renders a Notdog Solid icon as expected. But, `<wa-icon family="jelly">` instead renders a Classic Solid icon — I'm guessing because `solid` is the default `variant`, and Jelly doesn't have a `solid` variant.

Should we make sure that `<wa-icon family="jelly">` renders a Jelly Regular icon? e.g.

```suggestion
    folder = 'jelly-regular';
    if (variant === 'regular') folder = 'jelly-regular';
    if (variant === 'duo-regular') folder = 'jelly-duo-regular';
    if (variant === 'fill-regular') folder = 'jelly-fill-regular';
```

packages/webawesome/docs/docs/components/icon.md
```diff
@@ -396,6 +396,20 @@ If you're a [Font Awesome Pro+ customer](https://fontawesome.com/), you have acc
...
+        name="house"
+        style="--secondary-color: skyblue; --secondary-opacity: 0.8;"
+      ></wa-icon>
+      <wa-icon family="utility-fill" variant="semibold" name="house"></wa-icon>
+    </div>
+  </div>
+
   <div class="wa-flank" style="--flank-size: 10ch;">
     <a href="https://fontawesome.com/icons/packs/whiteboard" target="_blank">Whiteboard</a>
     <div class="wa-cluster" style="font-size: 32px;">
```
lindsaym-fa:
```suggestion
    <div class="wa-cluster" style="font-size: 1.5em;">
```

Past Lindsay seems to have given this icon a different size than the rest. 😅 Mind if we touch this up while we're in here?

packages/webawesome/src/components/icon/library.default.ts
claviska: Because they're pro only I was hitting the pro CDN for these packs, but we can let them fall through. Folks will see a 403/404 in either case.

claviska: [ca345b6](https://github.com/shoelace-style/webawesome/pull/1960/commits/ca345b6b4f8e1e8d411aedeee6871f81a50b1d4c)

packages/webawesome/src/components/icon/library.default.ts
claviska: yep!

packages/webawesome/src/components/icon/library.default.ts
claviska: We want it to "just work" so this works for me. It's one less line to omit 40, but updated in [f62f7e8](https://github.com/shoelace-style/webawesome/pull/1960/commits/f62f7e8526d610b48f278e1694b6102e9e122d0e).

lindsaym-fa (approved):
No complaints! ✨

PR #1967, opened 2026-01-20
fix tooltip safe triangles regression

KonnorRogers:
TLDR: when we changed anchors, we never accounted for "mouseout" occurring on the tooltip. (In fact, it was never possible to hover a tooltip because mouseout was only being listened for on the anchor, not the tooltip as well)


claviska (dismissed):
Looks good. I manually tested all of the tooltip examples as well and they seem fine.

PR #1976, opened 2026-01-23
fix(popover): maintain event handlers when moved in DOM

### LINKED ISSUE #1963: `wa-popover` disconnected from event handlers when parent container is moved
### Describe the bug

If a `wa-popover`'s container element is moved within the document, the `wa-popover` loses its event handlers and will not open when its anchor is triggered or dismiss post-move.

_I'm not sure if this is a duplicate of #1910 or solved by #1913, but wanted to raise it if not_

### Demo

The following Codepen includes steps to reproduce and a working minimal reproduction of the problem.

https://codepen.io/tcannonfodder/pen/JoKWRdm

### Browser / OS
 - OS: macOS 15.7.4 (24G508)
 - Browser: Safari Version 26.2 (20623.1.14.18.4), Google Chrome Beta Version 145.0.7632.5 (Official Build) beta (arm64)

Masty88:
Fixes #1963

## Summary
When a popover element is moved in the DOM using `appendChild` or similar methods, the `AbortController` signal gets aborted in `disconnectedCallback`, causing all event listeners to be removed. This fix ensures event handlers are properly re-established when the element is reconnected.

## Changes
1. A new `AbortController` is created in `connectedCallback` if the previous one was aborted
2. The anchor connection is re-established by forcing `handleForChange` to run
3. Event listeners are properly re-attached with the new signal

This follows the same pattern used in the `wa-tooltip` component.

## Testing
- Manual testing confirms the popover now works correctly after being moved in the DOM
- The fix has been verified against the reproduction case in the original issue

### DIFF
--- a/packages/webawesome/src/components/popover/popover.ts
+++ b/packages/webawesome/src/components/popover/popover.ts
@@ -97,6 +97,17 @@ export default class WaPopover extends WebAwesomeElement {
     if (!this.id) {
       this.id = uniqueId('wa-popover-');
     }
+
+    // Recreate event controller if it was aborted
+    if (this.eventController.signal.aborted) {
+      this.eventController = new AbortController();
+    }
+
+    // Re-establish anchor connection after being moved in the DOM
+    if (this.for && this.anchor) {
+      this.anchor = null; // force reattach
+      this.handleForChange();
+    }
   }
 
   disconnectedCallback() {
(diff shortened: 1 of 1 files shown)


packages/webawesome/src/components/popover/popover.ts
```diff
@@ -295,9 +306,9 @@ export default class WaPopover extends WebAwesomeElement {
             arrow:popup__arrow
           "
           class=${classMap({
-            popover: true,
-            'popover-open': this.open,
-          })}
+      popover: true,
+      'popover-open': this.open,
+    })}
```
claviska:
```suggestion
            popover: true,
            'popover-open': this.open,
          })}
```

claviska (approved):
Thanks for the PR! This looks good. I'll add a changelog entry.

PR #1977, opened 2026-01-23
Docs: Use Combobox for Examples with Many Options

talbs:
This PR updates documentation examples to use `wa-combobox` instead of `wa-select` when there are many options to choose from. This improves the user experience by providing a searchable/filterable interface for long option lists.

- [Popup Docs](https://webawesome-git-talbs-c-c-c-combo-boxer-font-awesome.vercel.app/docs/components/popup)
- [Animation Docs](https://webawesome-git-talbs-c-c-c-combo-boxer-font-awesome.vercel.app/docs/components/animation)


claviska:
@KonnorRogers what's the best way forward to ensure contributors can see the combobox in the docs? Since we're not publishing this to the early CDN, we can't use that trick.

It would be good to solve this for `<wa-page>`, `<wa-combobox>`, and other pro components we're releasing.

claviska (approved):

Issue #1933, opened 2026-01-10
Required <wa-input> with no label still shows label field with *


patrickvuarnoz:
### Describe the bug
When having the following configuration the field shows the asteriks (*) to indicate that the field is required. This is undesired, since it breaks the layout. The expectation is to only show the input field.

```
<wa-input name="email" type="email" placeholder="Email" autocomplete="email" required></wa-input>
```

<img width="409" height="68" alt="Image" src="https://github.com/user-attachments/assets/46c2c11d-5af5-4442-b760-77f47a83f028" />

### Browser / OS
 - OS: MacOs
 - Browser: Chrome
 - Browser version: 143.0.7499.170

### Additional information
WebAwesome 3.1.0

### Quickfix

```
wa-input:not([label]) {
  &::part(label) {
    display: none;
  }
}
```

claviska:
Copied from https://github.com/shoelace-style/webawesome/pull/1981#issuecomment-3837084123:

---

We typically say that labels are required for accessibility. In some cases, a visually hidden label is acceptable when sighted users can convey what the form control does in context (e.g. a search bar with an icon and/or placeholder).

However, even in that cases it's always better to include a visually hidden label for screen readers. This will effectively hide the label (and the asterisk) but still make it available to assistive devices.

I suggest we do this through the use of a utility class, e.g.:

```html
<wa-input class="wa-vh-label">
```

@lindsaym-fa do you already have plans for visually hidden utility classes?

lindsaym-fa:
@claviska, we actually do already support `.wa-visually-hidden-label`!

@patrickvuarnoz, if you add `class="wa-visually-hidden-label"` to your `<wa-input>`, you won't see that pesky asterisk anymore. This also allows you to set a `label=""` for better a11y.

Here's the docs: https://webawesome.com/docs/utilities/visually-hidden/#visually-hidden-input-parts
And an example: https://codepen.io/team/webawesome/pen/NPrMMGr

claviska:
@lindsaym-fa you're adding awesome way too fast. I didn't even realize we had added it 😄

BBovvser:
Apparently the `wa-visually-hidden-label` class trick doesn't work with `<wa-radio-group>`. The label (and asterisk when setting  `required`) still shows up.

lindsaym-fa:
Thanks, @BBovvser — we'll have that fixed up ASAP.

patrickvuarnoz:
[...]

PR #1981, opened 2026-01-27
updated styles for inputs with required attributes

kelseythejackson:
**Issue**: Required `<wa-input>` with no label still shows label field with `*` ([#1933](https://github.com/shoelace-style/webawesome/issues/1933))                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                           
  **Root Cause**: Style rules for the required asterisk (`::after` pseudo-element) were applied when the host had `[required]`, whether or not there was anything in the label.                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                           
  ## Style Fixes (3 files)                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                           
  | File | Change |                                                                                                                                                                                                                                                                                                                                                        
  |------|--------|                                                                                                                                                                                                                                                                                                                                                        
  | `src/styles/component/form-control.styles.ts` | Added `:is(:has(*:not(:empty)), .has-label)` condition |                                                                                                                                                                                                                                                               
  | `src/components/checkbox/checkbox.styles.ts` | Added `:has(*:not(:empty))` condition |                                                                                                                                                                                                                                                                                 
  | `src/components/switch/switch.styles.ts` | Added `:has(*:not(:empty))` condition |                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                           
  ## Tests Added (3 files)                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                           
  | File | Tests Added |                                                                                                                                                                                                                                                                                                                                                   
  |------|-------------|                                                                                                                                                                                                                                                                                                                                                   
  | `src/components/input/input.test.ts` | 2 tests for required asterisk visibility |                                                                                                                                                                                                                                                                                      
  | `src/components/checkbox/checkbox.test.ts` | 2 tests for required asterisk visibility |                                                                                                                                                                                                                                                                                
  | `src/components/switch/switch.test.ts` | 2 tests for required asterisk visibility |                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                           
  ## Components Affected                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                           
  | Component | File |                                                                                                                                                                                                                                                                                                                                                
  |-----------|-----------|                                                                                                                                                                                                                                                                                                                                                
  | `wa-input` | via `form-control.styles.ts` + test |                                                                                                                                                                                                                                                                                                                     
  | `wa-textarea` | via `form-control.styles.ts` |                                                                                                                                                                                                                                                                                                                         
  | `wa-select` | via `form-control.styles.ts` |                                                                                                                                                                                                                                                                                                                           
  | `wa-checkbox` | via `checkbox.styles.ts` + test |                                                                                                                                                                                                                                                                                                                      
  | `wa-switch` | via `switch.styles.ts` + test |                                                                                                                                                                                                                                                                                                                          
  | `wa-radio-group` | via `form-control.styles.ts` |                                                                                                                                                                                                                                                                                                                      
  | `wa-color-picker` | via `form-control.styles.ts` |                                                                                                                                                                                                                                                                                                                     
  | `wa-slider` | via `form-control.styles.ts` |                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                           
  The checkbox and switch needed their own fixes because they have component-specific style rules that override the shared `form-control.styles.ts`. The other components (textarea, select, radio-group, color-picker, slider) all use only the shared styles, so they're automatically fixed.


claviska:
This fix doesn't seem to work. The user's original [test case](https://github.com/shoelace-style/webawesome/issues/1933#issue-3799887176) is failing in all browsers.

CleanShot 2026-02-02 at 13 56 09@2x

I didn't dive in, but I suspect the `:empty` select isn't doing what it's expected to be doing.

I'm closing this PR because it doesn't solve the stated problem and I think we may want to approach this differently...

## Alternative Solution

We typically say that labels are required for accessibility. In some cases, a visually hidden label is acceptable when sighted users can convey what the form control does in context (e.g. a search bar with an icon and/or placeholder).

However, even in that cases it's always better to include a visually hidden label for screen readers. This will effectively hide the label (and the asterisk) but still make it available to assistive devices.

I suggest we do this through the use of a utility class, e.g.:

```html
<wa-input class="wa-vh-label">
```

@lindsaym-fa do you already have plans for visually hidden utility classes?

lindsaym-fa:
We already have `.wa-visually-hidden-label` to account for this 😊 — https://webawesome.com/docs/utilities/visually-hidden/#visually-hidden-input-parts

Example: https://codepen.io/team/webawesome/pen/NPrMMGr

PR #1983, opened 2026-01-27
fix framework documentation

KonnorRogers:
More framework documentation fixes


claviska (approved):
Great work! If @lindsaym-fa or @kelseythejackson or @talbs have time, this page could use some visual love:

CleanShot 2026-01-27 at 16 23 07@2x

Not a blocker. We can merge as-is and improve visuals later.

PR #1987, opened 2026-01-27
fix code example docs

KonnorRogers:
@kkoop64 I didn't get a chance to test this, you mind double checking it works as expected for you? It was breaking the app using the relative path 😅


kkoop64 (dismissed):
Still works on my end, apologies for breaking the app!

lindsaym-fa (approved):

Issue #1366, opened 2025-08-23
wa-cloak being removed before components update


KonnorRogers:
### Describe the bug

it seems `class='wa-cloak'` gets removed too soon and you still see FOUCE (only in Firefox)

### To Reproduce

Use firefox go here: https://codepen.io/paramagicdev/pen/pvjKyOx?editors=1010

hard refresh page.

you'll see component flash unstyled.

Was able to reproduce locally too on an `<html class="wa-cloak">` element.

### Demo

https://github.com/user-attachments/assets/bc92a422-f868-445c-97cd-99a3effffd4d

### Screenshots
If applicable, add screenshots to help explain the bug.

### Browser / OS
 - OS: Mac
 - Browser: Firefox
 - Browser version: 142.0

claviska:
Does this happen in Firefox _outside_ of CodePen?

claviska:
Not quiet apples to apples, but the Web Awesome site is dogfooding `wa-cloak` and it appears to be hard reloading fine for me in Firefox.

https://webawesome.com/docs/components/button

I have a feeling this is a CodePen thing, but let me know if you're seeing it outside of that environment.

KonnorRogers:
@claviska I could reproduce it locally on my personal laptop, but on my work laptop its like 1 out of every 10 loads on my work computer so can't reliably reproduce it despite being the same Firefox version 🤔 even in codepen on this laptop its not 100% either 👀

owenthewizard:
Not sure if it's related, but I still get FOUCE with `wa-cloak`.

claviska:
@KonnorRogers can we co-review #1988 when you get back? I don't see how it will solve the issue, but I also can't replicate it locally.

KonnorRogers:
i can't replicate this anymore either

PR #1988, opened 2026-01-28
updated autoloader

kelseythejackson:
## Summary: Fix for `wa-cloak` FOUCE Issue (#1366)                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                           
 ### Problem                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                           
`wa-cloak` was being removed prematurely, causing FOUCE in Firefox. Components were rendering unstyled right before styles were applied.                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                           
  **Root Cause:** The `discover()` function in `autoloader.ts` only waited one `requestAnimationFrame` cycle after component imports completed before firing the `wa-discovery-complete` event. This wasn't enough time for the components to finish their first render cycle.                                                                  
                                                                                                                                                                                                                                                                                                                                                                           
  ### Solution                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                           
  **File Modified:** `src/utilities/autoloader.ts`                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                           
  Replaced the single `requestAnimationFrame` promise with waiting for all discovered components' `updateComplete` promises to resolve. This ensures each component's shadow DOM is fully rendered before the `wa-cloak` class is removed.                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                           
  **Before**                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                           
      await new Promise(requestAnimationFrame);                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                           
  **After**                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                           
      if (tagsToRegister.length > 0) {                                                                                                                                                                                                                                                                                                                                     
        const elements = root.querySelectorAll<HTMLElement>(tagsToRegister.join(','));                                                                                                                                                                                                                                                                                     
        await Promise.all(                                                                                                                                                                                                                                                                                                                                                 
          [...elements].map(el => (el as HTMLElement & { updateComplete?: Promise<boolean> }).updateComplete),                                                                                                                                                                                                                                                             
        );                                                                                                                                                                                                                                                                                                                                                                 
      }                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                           
  `updateComplete` is a promise [Lit](https://lit.dev/docs/components/lifecycle/#updatecomplete); it resolves when the element has finished updating and rendering its template.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                           
  ### Tests Added                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                           
  **File Created:** `src/utilities/autoloader.test.ts`                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                           
  - **dispatch `wa-discovery-complete` after discovering components** - Basic functionality verification                                                                                                                                                                                                                                                                     
  - **dispatch `wa-discovery-complete` even when no components need registration** - Edge case (empty container)                                                                                                                                                                                                                                                             
  - **wait for component updateComplete before dispatching** - Core fix verification                                                                                                                                                                                                                                                                                       
  - **have components fully rendered when `wa-discovery-complete` fires** - Core fix verification                                                                                                                                                                                                                                                                            
  - **handle nested components** - Complex DOM structures                                                                                                                                                                                                                                                                                                                  
  - **not fail when components fail to import** - Error handling / graceful degradation                                                                                                                                                                                                                                                                                    
  - **handle components that are already defined** - Pre-registered component edge case                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                           
  ### Verification                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                           
  - **Build:** Passes                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  - **New Autoloader Tests:** 7 passed

### DIFF
--- a/packages/webawesome/src/utilities/autoloader.ts
+++ b/packages/webawesome/src/utilities/autoloader.ts
@@ -51,8 +51,14 @@ export async function discover(root: Document | Element | ShadowRoot) {
     }
   }
 
-  // Wait a cycle to allow the first Lit update to run
-  await new Promise(requestAnimationFrame);
+  // Wait for all components to complete their first render to prevent FOUCE.
+  // Using updateComplete ensures the component's shadow DOM is fully rendered.
+  if (tagsToRegister.length > 0) {
+    const elements = root.querySelectorAll<HTMLElement>(tagsToRegister.join(','));
+    await Promise.all(
+      [...elements].map(el => (el as HTMLElement & { updateComplete?: Promise<boolean> }).updateComplete),
+    );
+  }
 
   // Dispatch an event when discovery is complete.
   root.dispatchEvent(
--- a/packages/webawesome/src/utilities/autoloader.test.ts
+++ b/packages/webawesome/src/utilities/autoloader.test.ts
@@ -1,0 +1,179 @@
+import { expect } from '@open-wc/testing';
+import sinon from 'sinon';
+import { discover } from './autoloader.js';
+
+describe('autoloader', () => {
+  describe('discover()', () => {
+    it('should dispatch wa-discovery-complete event after discovering components', async () => {
+      const container = document.createElement('div');
+      document.body.appendChild(container);
+
+      const handler = sinon.spy();
+      container.addEventListener('wa-discovery-complete', handler);
+
+      // Add an undefined component
+      container.innerHTML = '<wa-button>Test</wa-button>';
+
+      await discover(container);
+
+      expect(handler).to.have.been.calledOnce;
+
+      container.remove();
+    });
+
+    it('should dispatch wa-discovery-complete even when no components need registration', async () => {
+      const container = document.createElement('div');
+      document.body.appendChild(container);
+
+      const handler = sinon.spy();
+      container.addEventListener('wa-discovery-complete', handler);
+
+      // No components in the container
+      container.innerHTML = '<div>No components here...</div>';
+
+      await discover(container);
+
+      expect(handler).to.have.been.calledOnce;
+
+      container.remove();
+    });
+
+    it('should wait for component updateComplete before dispatching wa-discovery-complete', async () => {
+      const container = document.createElement('div');
+      document.body.appendChild(container);
+
+      // Add a component that needs to be discovered
+      container.innerHTML = '<wa-spinner></wa-spinner>';
+
+      // Track the order of events
+      const events: string[] = [];
+
+      container.addEventListener('wa-discovery-complete', () => {
+        events.push('discovery-complete');
+      });
+
+      await discover(container);
+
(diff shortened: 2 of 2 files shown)


claviska:
> Root Cause: The discover() function in autoloader.ts only waited one requestAnimationFrame cycle after component imports completed before firing the wa-discovery-complete event. This wasn't enough time for the components to finish their first render cycle.

**How can I reliably see the problem in Firefox?**

This solution doesn't appear necessary because the autoloader already awaits component registration BEFORE awaiting the `requestAnimationFrame()` call, which [Lit suggests](https://lit.dev/docs/components/lifecycle/#updatecomplete) (and [verified here by a core team member](https://github.com/lit/lit-element/issues/1206#issuecomment-889342074)) is preferred for awaiting component updates since it runs after the microtask queue executes.

I'd like to replicate this issue before landing on a solution, but I don't see any FOUCE on webawesome.com in Firefox at the moment.

packages/webawesome/src/utilities/autoloader.ts
```diff
@@ -51,8 +51,14 @@ export async function discover(root: Document | Element | ShadowRoot) {
...
   }
 
-  // Wait a cycle to allow the first Lit update to run
-  await new Promise(requestAnimationFrame);
+  // Wait for all components to complete their first render to prevent FOUCE.
+  // Using updateComplete ensures the component's shadow DOM is fully rendered.
+  if (tagsToRegister.length > 0) {
+    const elements = root.querySelectorAll<HTMLElement>(tagsToRegister.join(','));
+    await Promise.all(
+      [...elements].map(el => (el as HTMLElement & { updateComplete?: Promise<boolean> }).updateComplete),
```
claviska, line 59: We already loop through undefined WA components on line 33, then we await all registrations on line 45. At that point, they haven't updated yet, but they are all guaranteed to be registered (if the tag names are valid).

> await new Promise(requestAnimationFrame);

The removed line should be sufficient per the Lit docs, but I'd like to verify the before and after behavior. I suspect there may be something else causing the FOUCE.

claviska:
> How can I reliably see the problem in Firefox?

gentle nudge

Issue #1688, opened 2025-10-30
Number Input


talbs:
This issue tracks the upcoming work on the Web Awesome Number Input component (thanks to the support of our Kickstarter backers).

claviska:
Added in https://github.com/shoelace-style/webawesome/pull/1990

Issue #1076, opened 2025-06-25
Stretch goals: 14 new components in WA free


claviska:
Per stretch goal promises, we have committed to releasing 14 new components in Web Awesome Free:

- [x] 1 – ~Scroller~ ✅ 
- [x] 2 – ~Popover~ ✅ 
- [x] 3 – ~Zoomable Frame~ ✅ 
- [x] 4 –  ~Intersection Observer~ ✅ 
- [x] 5 – ~Number Input~ ✅ 
- [x] 6 – ~Markdown~ ✅ 
- [x] 7 – Accordion / Accordion Item ✅ 
- [x] 8 – Known Date ✅ 
- [x] 9 – Time Input ✅ 
- [x] 10 – Checkbox Group ✅ 
- [x] 11 – Random Content ✅
- [x] 12 – Passcode / PIN Input / OTP ✅
- [x] 13 – Pagination ✅
- [x] 14 – Tag Input ✅ ([view PR](https://github.com/shoelace-style/webawesome/pull/2796))

claviska:
Closing this as all core stretch goal components have been released except for `<wa-tag-input>`, which is [currently in review](https://github.com/shoelace-style/webawesome/pull/2796).

PR #1990, opened 2026-01-28
Add `<wa-number-input>`

claviska:
Adds Number Input as a free stretch goal component (5 of 14) per #1076 and #1688.

- Requesting @lindsaym-fa for design review
- Requesting @KonnorRogers for form-associated review


packages/webawesome/package.json
```diff
@@ -57,8 +57,7 @@
     "url": "https://github.com/shoelace-style/webawesome/issues"
   },
   "scripts": {
-    "start": "node scripts/build.js --develop",
-    "start:alpha": "echo 'This command has been deprecated. Use: \"npm start\"\n'",
+    "start": "node scripts/build.js --develop -- --no-incremental",
```
claviska: Added this because it's still not showing rebuilding properly without it half the time.

packages/webawesome/src/internal/webawesome-element.ts
```diff
@@ -114,7 +114,7 @@ export default class WebAwesomeElement extends LitElement {
   }
 
   /**
-   * Methods for setting and checking custom states.
+   * @internal Methods for setting and checking custom states.
```
claviska, line 117: Added because these were erroneously showing up in the component docs.

packages/webawesome/src/internal/webawesome-element.ts
```diff
@@ -148,7 +148,7 @@ export default class WebAwesomeElement extends LitElement {
   };
 
   /**
-   * Given a native event, this function cancels it and dispatches it again from the host element using the desired
+   * @internal Given a native event, this function cancels it and dispatches it again from the host element using the desired
```
claviska, line 151: Added because these were erroneously showing up in the component docs.

claviska:
Thanks for the style updates @lindsaym-fa! Let me know if you see anything else that needs to be addressed.

lindsaym-fa:
One final style option that I wanted to get your thoughts on before I commit —

Removing the borders from the stepper buttons feels more cohesive with the rest of the library, imo. (We don't border off the password toggle or clear button in inputs, for example.)

Here's what it looks like if we get rid of the borders (left) and add a clearer hover state (right):

Screenshot 2026-01-29 at 4 39 02 PM
Screenshot 2026-01-29 at 4 38 44 PM


This also gives us the opportunity to round off the buttons in `pill` inputs, which feels a little bit cleaner:

Screenshot 2026-01-29 at 4 36 40 PM

The only scenario that gives me pause is steppers + slotted icons since we lose the visual anchor for the icon...

Screenshot 2026-01-29 at 4 37 20 PM

Curious if you have a preference, @claviska?

claviska:
@lindsaym-fa that looks great! Feel free to tweak as you see fit 🤩

lindsaym-fa (dismissed):
No further changes, boss! Works great. 😄

PR #1986, opened 2026-01-27
updated qr code

kelseythejackson:
**Problem**: `wa-qr-code` extended edge-to-edge without a ["quiet zone"](https://www.the-qrcode-generator.com/blog/qr-code-quiet-zone) (buffer area), making them difficult _or_ impossible to scan on dark-themed websites.                                                                              
                                                                                                                                                                                                                                                      
  **Solution**: Added a `margin` property that creates a quiet zone around the QR code using the background color.                                                                                                                                    
                                                                                                                                                                                                                                                      
  ## Changes Made                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                      
  ### `src/components/qr-code/qr-code.ts`                                                                                                                                                                                                             
  - Added `margin` property (default: 4) for the quiet zone size in pixels                                                                                                                                                                            
  - Wrapped the canvas in a container `<div>` with padding and background color matching the QR code's background                                                                                                                                     
  - Updated `generate()` to set explicit width/height on the host element (including margin)                                                                                                                                                          
  - Changed the canvas check from `hasUpdated` to `!this.canvas` for more reliable timing                                                                                                                                                             
                                                                                                                                                                                                                                                      
  ### `src/components/qr-code/qr-code.styles.ts`                                                                                                                                                                                                      
  - Simplified styles for the new structure                                                                                                                                                                                                           
  - Container div uses `box-sizing: content-box` to ensure padding adds to dimensions                                                                                                                                                                 
                                                                                                                                                                                                                                                      
  ### `src/components/qr-code/qr-code.test.ts`                                                                                                                                                                                                        
  - Added tests for default margin (4px)                                                                                                                                                                                                              
  - Added tests for custom margin values                                                                                                                                                                                                              
  - Added test for margin=0 (no quiet zone)                                                                                                                                                                                                           
                                                                                                                                                                                                                                                      
  ## Usage                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                      
  ```html                                                                                                                                                                                                                                             
  <!-- Default (4px quiet zone) -->                                                                                                                                                                                                                   
  <wa-qr-code value="https://example.com"></wa-qr-code>                                                                                                                                                                                               
                                                                                                                                                                                                                                                      
  <!-- Custom margin -->                                                                                                                                                                                                                              
  <wa-qr-code value="https://example.com" margin="10"></wa-qr-code>                                                                                                                                                                                   
                                                                                                                                                                                                                                                      
[...]

Issue #1866, opened 2025-12-11
QR scanability on dark mode sites


Jack-L94:
### Describe the bug
When using wa-qr-code on a dark mode site to set up two-factor auth, because the QR code fills the entire canvas, apps like 2FAS & Authy have issues reading the QR code.

### To Reproduce
Steps to reproduce the behaviour:

1. Go to https://webawesome.com/account/settings while using the dark mode theme
2. Start the Two-Factor Authentication setup
3. Attempt to scan the QR code with a two-factor app like 2FAS, Authy, Google Authenticator. 
4. App will do nothing when trying to scan the QR code

### Demo

https://codepen.io/Jackl-94/pen/ByKbaMb

Remove the wa-dark in the html settings to make the QR scannable.

### Browser / OS
 - OS: Windows 11 & Android 16
 - Browser: Firefox Developer
 - Browser version: 147.0b2 (64-bit)
 - Two-factor apps: 2FAS, Authy, Proton Authenticator & Bitwarden Browser extension

### Additional information
This is fixed by adding a border around the QR code on dark backgrounds, meaning the apps can read where the code is against the background.

Doing some further research into this, the [QR Code specification on page 14 ](https://cdn.standards.iteh.ai/samples/83389/dee29007cb62437f9767323e6af02f90/ISO-IEC-18004-2024.pdf) talks about a "Quiet Zone" - The symbol shall be surrounded on all four sides by a quiet zone border.

I wonder if adding a quiet zone attribute with a default of 4 by default to the component to ensure the code can be read?
https://www.the-qrcode-generator.com/blog/qr-code-quiet-zone

Nexus360Digital:
Would reversing the colours in dark mode also work -  so white areas become black etc.

Jack-L94:
I don't believe so, I believe the issue is the lack of quiet zone preventing apps from differentiateing where the QR code starts and stops on dark backgrounds

Nexus360Digital:
I agree - but if you reverse the colours with fill="white" background="black" there will be a black quiet zone.

Jack-L94:
Just tested that on the CodePen I posted and that does make the otpauth QR code scannable in 2FAS so that's a good workaround for now, just add that to the dark styles so it changes and should be scannable

claviska:
[...]

PR #1991, opened 2026-01-28
Improve QR code API

claviska:
- Deprecates the `fill` and `background` attributes as discussed (they still work when set)
- Uses `color` and `background` CSS properties to determine the QR code's color and background, allowing it to automatically adapt when the color scheme changes
- Improves sizing so a QR quiet zone can be added via `padding`
- Updated the docs to demonstrate

This improves the component in a backward compatible way.

Related: #1986, #1866

### DIFF
--- a/packages/webawesome/docs/docs/components/qr-code.md
+++ b/packages/webawesome/docs/docs/components/qr-code.md
@@ -41,20 +41,32 @@ QR codes are useful for providing small pieces of information to users who can q
 
 ## Examples
 
-### Colors
-
-Use the `fill` and `background` attributes to modify the QR code's colors. You should always ensure good contrast for optimal compatibility with QR code scanners.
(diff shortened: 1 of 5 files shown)


packages/webawesome/docs/docs/components/qr-code.md
```diff
@@ -41,20 +41,32 @@ QR codes are useful for providing small pieces of information to users who can q
 
 ## Examples
 
-### Colors
+### Size
```
claviska, line 44: I just swapped the positions of Size and Colors in the docs.

kelseythejackson (dismissed):

claviska:
On second pass, the `transitionend` handler was too eager. It now only re-renders the canvas when `color` changes, improving performance.

PR #1994, opened 2026-01-29
Add `flex-wrap` utility classes

lindsaym-fa:
Adds `flex-wrap` utility classes to control wrapping in flex containers. The changes refactor existing layout components to use lower-specificity defaults that can be easily overridden by the new utilities.

- Adds three new utility classes for `flex-wrap`: `wa-flex-wrap`, `wa-flex-nowrap`, and `wa-flex-wrap-reverse`
- Updates existing layout utility classes to give their default `flex-wrap` properties zero specificity
- Adds Flex Wrap documentation with visual previews


kkoop64 (dismissed):
Looks good to me. I was able to replace a number of `flex-wrap` and `justify-content` instances with these utilities in my patterns demo.

packages/webawesome/src/styles/utilities/flex-wrap.css
```diff
@@ -0,0 +1,16 @@
+@layer wa-utilities {
+  /* Apply Flexbox with 0 specificity to ensure a flex wrap util produces a visible change */
+  :where(.wa-flex-wrap, .wa-flex-nowrap, .wa-flex-wrap-reverse) {
```
claviska, line 3: Question: is `:where` still useful even with the `@layer` rule?

claviska (dismissed):
Nice addition. Just one non-blocking question to satisfy my curiosity :)

talbs (approved):
Makes sense and works for me. Thanks for adding these!

packages/webawesome/src/styles/utilities/flex-wrap.css
lindsaym-fa: Yep! Layers control cascade precedence, but not specificity. A less specific rule — even it's unlayered — could still be stomped out by a more specific rule in a cascade layer.

This also ensures that our other classes in `@layer wa-utilities` have the right priority. E.g., if someone uses `class="wa-grid wa-flex-wrap"`, we'd rather the higher order utility `wa-grid` win the battle to set the `display` property and let the lower order utility `wa-flex-wrap` take a back seat.

Issue #1992, opened 2026-01-29
Dropdown closes itself immediately when opened inside form


HerrNaN:
### Describe the bug
After bumping to 3.1.0, dropdown closes itself immediately when opened inside form.

### To Reproduce
1. Open CodePen
2. Click to open second dropdown (which is in a form)

### Demo

https://codepen.io/herrnan/pen/azZErBJ

### Screenshots
N/A

### Browser / OS
 - OS: Linux (Ubuntu)
 - Browser: Chromium
 - Browser version: 144

### Additional information
N/A

claviska:
Thanks for the report. Fixed in #1995.

PR #1995, opened 2026-01-30
Don't inherit slots when generating a submitter

claviska:
See #1992

This regression was introduced in 3d395653fcd7113d0e22b4efc5bf7f101ec8d68f. The temporary submitter button was copying _all_ attributes except `style`, so it copied the dropdown trigger's `slot` attribute, effectively making it a second "trigger" and closing the dropdown.


KonnorRogers:
People could be relying on queries that check the `slot` attribute.

I think the more a better approach is not to construct a light dom button for non-submit / non-reset buttons.

PR here: https://github.com/shoelace-style/webawesome/pull/1996

claviska:
Closing in favor of #1996

PR #1998, opened 2026-01-30
Add Agent Skill for WA

claviska:
This PR adds an [Agent Skill](https://agentskills.io/) to Web Awesome, including documentation on what it is and how to use it.

- Added Agent Skill generation for:
  - Installation
  - General usage docs
  - Support (how to get help from a human)
  - Components and their APIs
  - Framework integrations
  - Design tokens
  - Native styles and other utilities
  - Theming docs
  - Localization docs
- Added AI section to the sidebar
- Added Agent Skills docs page
- Updated llms.txt page with a comparison table of when to use llms.txt vs. Agent Skill
- Updated llms.txt generation to _not_ be a CEM plugin for consistency and maintainability

The goal was to limit static content and duplication while still keeping reference files relatively small. Some files are pulled in as-is and 

Due to heavy use of nunjucks, files are parsed from the Eleventy build instead of directly from the markdown source files. They are then cleaned up and converted back to markdown for the reference files. This required adding only one new dev dependency: [`turndown`](https://www.npmjs.com/package/turndown)

This is an experimental feature and is easy to adjust as feedback comes in.

### How To Review

The generated output lives in `dist/skills/webawesome` and `dist-cdn/skills/webawesome`. Remember that these files are optimized for LLMs, not humans.

- Run `npm run build` to generate the skill files
- Spot check `dist/skills/webawesome/*.md` for obvious issues
- **Review the Agent Skills doc page to understand how to use and install it**
- Bonus points: try it out by installing it and using it with Claude Code or another AI tool

At this point, the skill is NOT being published to [skills.sh](https://skills.sh/) but it can be installed locally. We'll consider publishing it after folks have had a chance to test things out.


packages/webawesome/scripts/agent-skill.js
```diff
@@ -0,0 +1,1077 @@
...
+| Theme | Palette | Brand Color |
+|-------|---------|-------------|
+| **Active** | Rudimentary | Green |
+| **Brutalist** | Default | Blue |
+| **Glossy** | Elegant | Indigo |
+| **Matter** | Mild | Purple |
+| **Mellow** | Natural | Blue |
+| **Playful** | Rudimentary | Purple |
+| **Premium** | Anodized | Cyan |
+| **Tailspin** | Vogue | Indigo |
```
claviska, line 687: I'd love to find a way to generate these without duplicating them here.

packages/webawesome/scripts/agent-skill.js
```diff
@@ -0,0 +1,1077 @@
...
+- **Bright** - Enhanced saturation
+- **Shoelace** - Classic Shoelace colors
+
+### Pro Palettes
+- **Rudimentary**
+- **Elegant**
+- **Mild**
+- **Natural**
+- **Anodized**
+- **Vogue**
```
claviska, line 704: Again, it would be nice to generate these instead of duplicating.

packages/webawesome/scripts/agent-skill.js
```diff
@@ -0,0 +1,1077 @@
...
+\`\`\`html
+<html class="wa-palette-bright">
+\`\`\`
+
+CSS variables follow the pattern \`--wa-color-{hue}-{tint}\`:
+
+\`\`\`css
+.my-element {
+  color: var(--wa-color-blue-600);
+  background: var(--wa-color-gray-100);
```
lindsaym-fa:
```suggestion
  color: var(--wa-color-blue-60);
  background: var(--wa-color-gray-10);
```

Issue #2001, opened 2026-02-02
[Docs] wa-radio-group `size` example is wacky


benopotamus:
### Describe the bug
The interactive example at https://webawesome.com/docs/components/radio-group/#sizing-options has problems (it doesn't work correctly)

### To Reproduce

1. Go to https://webawesome.com/docs/components/radio-group/#sizing-options
2. Click a size
3. Observe that only the label changes size
4. Click a different size
5. Observe that the label changes to the selected size but the radio buttons text changes to the size selected at step 2.

### Screenshots

<img width="388" height="331" alt="Image" src="https://github.com/user-attachments/assets/1715c3dc-7c0a-4d4b-8569-f8c279b11aaf" />

### Browser / OS
 - OS: Linux
 - Browser: Firefox
 - Browser version: 147.0.1

PR #2005, opened 2026-02-02
Fix radio group size bug

claviska:
While looking at a docs fix for #2001 I noticed the radio's behavior wasn't as documented. This PR fixes that bug and updates the docs to demonstrate how to control the size with both `<wa-radio-group size="*">` and with `<wa-radio size="*">` directly.

This allows you to have mixed sizes by omitting the size on `<wa-radio-group>` whereas before it would always default to medium. The example has been fixed and a new one added to demonstrate the correct behavior.

### Misc

- Also adds `.claude/settings.local.json` to gitignore
- Also adds the `--no-incremental` flag to local dev builds because I'm losing way to much time troubleshooting ghosts when it's just not rebuilding/reloading properly


packages/webawesome/src/components/radio-group/radio-group.ts
```diff
@@ -98,8 +98,8 @@ export default class WaRadioGroup extends WebAwesomeFormAssociatedElement {
   /** The default value of the form control. Primarily used for resetting the form control. */
   @property({ attribute: 'value', reflect: true }) defaultValue: string | null = this.getAttribute('value') || null;
 
-  /** The radio group's size. This size will be applied to all child radios and radio buttons, except when explicitly overridden. */
-  @property({ reflect: true }) size: 'small' | 'medium' | 'large' = 'medium';
+  /** The radio group's size. When present, this size will be applied to all `<wa-radio>` item inside. */
```
lindsaym-fa:
```suggestion
  /** The radio group's size. When present, this size will be applied to all `<wa-radio>` items inside. */
```

lindsaym-fa (dismissed):
Looks good to me! I think we just have a small typo, but otherwise approved!

PR #2007, opened 2026-02-02
Fix color tokens in Agent Skill

lindsaym-fa:
Corrects the color tokens in the Agent Skill documentation by changing from three-digit tint values (e.g., -600, -100) to two-digit tint values (e.g., -60, -10).


claviska (approved):

claviska:
Good catch. Approved

PR #2010, opened 2026-02-02
Add `align-self` utility classes

lindsaym-fa:
Adds `align-self` CSS utilities for individually defining an item's alignment on the cross-axis in a flex or grid container.
Previously requested in https://github.com/shoelace-style/webawesome/discussions/1668.

- Adds five new utility classes for `align-self` (start, end, center, stretch, and baseline)
- Updates existing Align Items documentation with details and previews of new `align-self` utilities

---

@talbs, I could use your opinion on my choice to wrap the new `align-self` classes into the existing Align Items doc and `styles/utilities/align-items.css`. Felt right to me, as long as we don't feel this hurts discoverability of the new classes.

### DIFF
--- a/packages/webawesome/docs/docs/utilities/align-items.md
+++ b/packages/webawesome/docs/docs/utilities/align-items.md
@@ -1,6 +1,6 @@
 ---
 title: Align Items
-description: Align items and align self utilities align items within flex and grid containers on the cross axis.
+description: Control cross-axis alignment in flex and grid containers with align-items and align-self utility classes.
 layout: docs
 tags: layoutUtilities
 ---
--- a/packages/webawesome/docs/docs/utilities/align-items.md
+++ b/packages/webawesome/docs/docs/utilities/align-items.md
@@ -1,6 +1,6 @@
 ---
 title: Align Items
-description: Align items utilities align items within flex and grid containers on the cross axis.
+description: Align items and align self utilities align items within flex and grid containers on the cross axis.
 layout: docs
 tags: layoutUtilities
 ---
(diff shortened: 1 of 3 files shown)


packages/webawesome/docs/docs/utilities/align-items.md
```diff
@@ -19,18 +19,35 @@ tags: layoutUtilities
     background-color: var(--layout-example-element-background);
     border-radius: var(--wa-border-radius-s);
     min-block-size: 1em;
+
+    &.de-emphasize {
```
talbs: `de-lightful` idea! If we use this elsewhere, can we abstract it out (present us or future us)?

packages/webawesome/docs/docs/utilities/align-items.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Align Items
-description: Align items utilities align items within flex and grid containers on the cross axis.
+description: Align items and align self utilities align items within flex and grid containers on the cross axis.
```
talbs:
```suggestion
description: Control cross-axis alignment in flex and grid containers with align-items and align-self utility classes.
```

The change here felt a bit awkward. Here's an attempt to rewrite it.

talbs (dismissed):
> @talbs, I could use your opinion on my choice to wrap the new align-self classes into the existing Align Items doc and styles/utilities/align-items.css. Felt right to me, as long as we don't feel this hurts discoverability of the new classes.

☝️ I think appending `align-items` makes sense - as these utilities would only ever be useful if you were already using `align-items`.

If we're worried about discoverability of `align-self`, I'd recommend adding an 11ty alias (soft redirect) so that ` /docs/utilities/self-align` points to this doc specifically.

I checked around the repo/codebase, and we don't do that for anything yet, but I'm happy to add it in a commit here if you'd like.

packages/webawesome/docs/docs/utilities/align-items.md
lindsaym-fa: Promoted in https://github.com/shoelace-style/webawesome/pull/2010/changes/7aecd291da50fc7fb571b05654a86e1e0f484231!

Let me know if you feel there's a better place or scope for this.

lindsaym-fa:
Great suggestions, @talbs — thank you!

A soft redirect sounds great. I'm not sure how to do that, so I'm happy to leave that to you and learn from it. 😄

talbs:
@lindsaym-fa, adding a redirect requires a bit more code setup than expected. And I'd like to make sure Konnor is cool with the approach.

So let's not hold this work up. I'll create a separate PR to get that in place AND create the redirect we chatted about.

talbs (approved):

PR #2014, opened 2026-02-03
Fix popover closing unexpectedly when used inside a shadow root #1969

### LINKED ISSUE #1969: wa-popover closed unexpectedly on click when it is declared in a shadow DOM
### Describe the bug
The wa-popover component is supposed to close when the user clicks outside of the component. When the component is declared in a shadow root of another component. The wa-popover closes even when the user clicks on a component inside the popover.

### To Reproduce

here is the code of a component to reproduce the bug:

```
import { LitElement, css, html } from 'lit';
import { customElement } from 'lit/decorators.js';

@customElement('bug-popover')
export class BugPopover extends LitElement {
  static styles = css`
    :host {
      display: block;
    }
  `;

  render() {
    return html`
      <wa-popover for="popover__autofocus">
        <div style="display: flex; flex-direction: column; gap: 1rem;">
          <wa-textarea autofocus placeholder="What's on your mind?" size="small" resize="none" rows="3"></wa-textarea>
          <wa-button variant="primary" size="small" data-popover="close"> Submit </wa-button>
        </div>
      </wa-popover>

      <wa-button id="popover__autofocus">
        <wa-icon name="comment" slot="start"></wa-icon>
        Feedback
      </wa-button>
    `;
  }
}

declare global {
  interface HTMLElementTagNameMap {
    'bug-popover': BugPopover;
  }
}
```

```
<!doctype html>
<html lang="en">
<head>
    <meta charset="UTF-8" />
    <link rel="icon" type="image/svg+xml" href="./icon.svg" />
    <script type="module" src="/test/main.ts"></script>
</head>
<body>
    <bug-popover></bug-popover>
</body>
</html>
```

- Click on the button feedback
- Click on the input field inside the popover
- The popover closes (expected behavior: the popover should stay opened)

### Browser / OS
 - OS: Windows, Linux
 - Browser: Chrome, Chomium, Firefox
 - Browser version: v143 (Chrome)

### Additional information
As a workaround, we add an event handler for the click event on the wa-popover. This handler stops the propagation of the event for any click except from the submit button

```
  private fixCloseOnClickForPopover(e: Event) {
    const submitButton = this.shadowRoot?.querySelector('wa-button');
    if(submitButton && !e.composedPath().includes(submitButton)) {
      e.stopPropagation();
    }
  }

```

kelseythejackson:
This fixes the popover closing unexpectedly when clicked inside while declared in a shadow DOM.
                                                                                                                                                                                                                                                                                     
### Problem                                                                                                                                                                                                                                                                                                                                                               

When a `wa-popover` is placed inside another custom element's shadow DOM, clicking on elements inside the popover (like a `textarea` or `button`) causes the popover to close immediately.                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                           
  The issue was in `handleDocumentClick()`:                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                           
  ```typescript                                                                                                                                                                                                                                                                                                                                                            
  if (target.closest('wa-popover') !== this) {                                                                                                                                                                                                                                                                                                                             
    this.open = false;                                                                                                                                                                                                                                                                                                                                                     
  }                                                                                                                                                                                                                                                                                                                                                                        
 ```                                                                                                                                                                                                                                                                                                                                                                          
`Element.closest()` cannot traverse shadow DOM boundaries, so when the popover is nested in another shadow root, `closest()` returns null and the popover closes.                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                           
  ### Solution                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                           
  Replace `closest()` with `composedPath().includes()` which properly traverses shadow DOM boundaries:   

  ```typescript                                                                                                                                                                                                                                                                                                                                                            
  if (!event.composedPath().includes(this)) {                                                                                                                                                                                                                                                                                                                              
    this.open = false;                                                                                                                                                                                                                                                                                                                                                     
  }                                                                                                                                                                                                                                                                                                                                                                          
 ```      


This mirrors the existing anchor check pattern already used in the same method.                                                                                                                                                                                                                                                                                          
Fixes #1969


KonnorRogers (approved):

packages/webawesome/src/components/popover/popover.ts
```diff
@@ -161,15 +161,13 @@ export default class WaPopover extends WebAwesomeElement {
...
       return;
     }
 
-    // Detect when clicks occur outside the popover
-    if (target.closest('wa-popover') !== this) {
+    // Detect when clicks occur outside the popover (using composedPath to traverse shadow DOM boundaries)
+    if (!event.composedPath().includes(this)) {
       this.open = false;
     }
   };
```
KonnorRogers, line 173: Minor change, but `composedPath()` can sometimes be expensive, so better to compute it once.

```suggestion
  private handleDocumentClick = (event: PointerEvent) => {
    const composedPath = event.composedPath()
    // Ignore clicks on the anchor so it will be closed by the anchor's click handler
    if (this.anchor && composedPath.includes(this.anchor)) {
      return;
    }

    // Detect when clicks occur outside the popover (using composedPath to traverse shadow DOM boundaries)
    if (!composedPath.includes(this)) {
      this.open = false;
    }
  };
```

KonnorRogers (changes requested):

KonnorRogers (approved):

PR #2015, opened 2026-02-03
Docs: Set Up Redirects

talbs:
Following up on a thread from https://github.com/shoelace-style/webawesome/pull/2010#pullrequestreview-3745765118...

This work allows Docs pages to have alternate URLs that send people to the same page.  

For example, `/docs/utilities/self-align` will redirect to the Align Items Docs (`/docs/utilities/align-items/`).

### How it works
- In a Doc-based view's front matter, you add an aliases list with the alternate path(s).
- The build creates a small redirect page at each alias (meta refresh + “click here” link), so the alias URL works when you open it.
- No server or app changes: everything is static HTML.

### How to add an alias
On the Doc-based view you want an alias to point to, add an `aliases` entry to the front matter like:

```md
---
...
aliases:
  - /docs/utilities/self-align
 ...
---
```

You can list multiple aliases; each gets its own redirect page.

### DIFF
--- a/packages/webawesome/docs/.eleventy.js
+++ b/packages/webawesome/docs/.eleventy.js
@@ -124,6 +124,7 @@ export default async function (eleventyConfig) {
   eleventyConfig.addFilter('stripExtension', string => path.parse(string + '').name);
   eleventyConfig.addFilter('stripPrefix', content => content.replace(/^wa-/, ''));
   eleventyConfig.addFilter('uniqueId', (_value, length = 8) => nanoid(length));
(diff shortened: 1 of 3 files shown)


claviska:
There might be a much simpler way to do this in `webawesome-app` where we use, e.g. a JSON config of old/new paths and handle it with a middleware.

talbs:
> There might be a much simpler way to do this in webawesome-app where we use, e.g. a JSON config of old/new paths and handle it with a middleware.

Okay, I'll hang tight on this until we explore that simpler way.

Issue #1814, opened 2025-11-21
wa-tag remove icon and text color mismatch with accent appearance and brand variant


andrewbabs63:
### Describe the bug
When using  `appearance="accent"` and `variant="brand"` on a tag with a removable button, the icons are very difficult to see. The icon color should match the text color I believe.

### To Reproduce

1. Navigate to https://webawesome.com/docs/components/tag/#removable. Open the associated CodePen. 
2. Add an `appearance="accent"` and `variant="brand"`  to the tag.
3. See that the icon color isn't matching the text color.

### Screenshots

<img width="1346" height="450" alt="Image" src="https://github.com/user-attachments/assets/4771b691-4fbc-4ccd-a5c5-c71255b648c8" />

### Browser / OS
 - OS: Mac
 - Browser: Chrome

kelseythejackson:
@andrewbabs63 Hey there, we have a fix in for this issue. Thanks

PR #2016, opened 2026-02-03
updated tag styles

kelseythejackson:
Fix wa-tag remove icon color mismatch with accent appearance (#1814)                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                           
  The remove button icon now inherits the tag's text color via `currentColor`, making sure there is proper contrast when using appearance="accent" with any variant.


lindsaym-fa (approved):
Added a couple of lines to remove the mismatched background color on hover and active. Good to go!

Issue #1839, opened 2025-12-02
wa-color-picker not flipping horizontally


cellcrowd:
Recently, the wa-color-picker component has been fixed with flip, however this only works vertically. When added to the righthand side of the screen, it will not flip as seen here:
https://codepen.io/m6246/pen/azvrvyB

I found that when removing sync="width" from the internal popup component through devtools, it also flips horizontally.

kelseythejackson:
Hey @cellcrowd, we'll take a look at this. Thanks

kelseythejackson:
Hey @cellcrowd, we have a fix in for this. Thanks for raising the issue.

PR #2024, opened 2026-02-04
removed sync

kelseythejackson:
Fix for #1839 
                                                                                                                                                                                                                                                                                                                                                                           
  Removed `sync="width"` from the wa-popup element in the color picker component.
---
The `sync="width"` attribute forces the popup width to match the anchor element (the small trigger button). This constraint prevents Floating UI from properly flipping the popup horizontally when positioned near the right edge of the viewport.

  Reasoning:
---
- The color picker panel has its own designed dimensions via CSS custom properties (--grid-width, --grid-height)
- The trigger is a small color swatch button, so syncing the panel width to it would make the picker unusable
- The popup was rendering correctly due to CSS overrides, but sync was still affecting Floating UI's flip calculations
- The sync attribute is useful for components like dropdowns where the menu should match the trigger width, but may not be the best used in the color picker


claviska (approved):
The fix checks out.

claviska (changes requested):
The fix works. Just needs a changelog entry.

PR #1925, opened 2026-01-08
upgraded JSX type generator and include SolidJS support

break-stuff:
- Upgraded JSX type generator package
- Added support for SolidJS
- Added strongly typed event targets for JSX users
- Updated types with a payload to use the event type rather than the details type for better typing
- Removed unused `modal` property from the Drawer component

PR #2026, opened 2026-02-04
Bug fix for StencilJS compatibility

### LINKED ISSUE #1919: custom-elements-jsx.d.ts is not Stencil-compatible and fails with skipLibCheck=false
### Describe the bug
When using Web Awesome in a Stencil (TSX) project, the published typings in
`dist/custom-elements-jsx.d.ts` are not compatible with Stencil’s JSX typing system and also fail TypeScript checking when `skipLibCheck` is disabled.

I am not an expert in how all this works, but I think it is supposed to work because this exists.
```
declare module '@stencil/core' {
  namespace JSX {
    interface IntrinsicElements extends CustomElements {}
  }
  export interface CSSProperties extends CustomCssProperties {}
}
```

I noted that:

* JSX attributes such as onClick are not included, which causes valid Stencil TSX like: `<wa-button onClick={this.closeMenu} />` to error with something like `Property 'onClick' does not exist on type 'Partial<WaButtonProps & BaseProps<WaButton>>'
* There might be some broken type references in the published file.  There are a few `onundefined` props.

### To Reproduce

```
npm init stencil
...
npm install @awesome.me/webawesome --save-dev
...

Then try and:
<wa-button variant="brand" onClick={() => {}} />
```

I get a bit further by adding `skipLibCheck` or commenting out the `onundefined` types.

For instance, in the screenshot below - I would expect the "variant=brands" to fail, since it should be "brands".   However, onClick should work.

Image

It might just be me misunderstanding how this is supposed to work, but any advice/best practice would be much appreciated on working with StencilJS.

### Browser / OS
 - OS: Windows
 - Browser: Chrome
 - Browser version: [e.g. 22]

kelseythejackson:
Fixes #1919 

**Custom-elements-jsx.d.ts** generating "onundefined" event handlers 
---                                                                                                                                                                                                                                                                                                    
The Custom Elements Manifest analyzer was detecting events in two ways: from @event JSDoc tags (which include event names) and from analyzing dispatchEvent() calls in code (which only captured the type, not the name). This created duplicate event entries where some had only a type but no name, causing the JSX types generator to produce `onundefined` event handlers instead of proper names like "onwa-error".

Added a `wa-filter-unnamed-events plugin` that removes events without names before the JSX types are generated. This fixes 25 broken event entries across components, including `wa-icon`, `wa-button`, `wa-input`, `wa-select`, `wa-dialog`, `wa-drawer`, `wa-dropdown`, `wa-tooltip`, and others.

Also added a verification step to the build process that checks for unnamed events in the CEM and `onundefined` in the generated JSX types. The build will fail if these issues are detected, preventing regression.


claviska:
I'm not sure this is the right fix, but if @JohnHardy can verify if it actually works I'd be more apt to entertain it.

This also might be fixed by #1925. Unfortunately, we had to revert the original PR but putting this burden on a CEM plugin that generates the correct types would be much better than trying to patch it with a filter plugin.

Also, any CEM validation should be part of `generateManifest()`, not a separate build step.

JohnHardy:
I'll test it as soon as I can, thank you for doing this! 🙏

kelseythejackson:
Updated to remove the file and add validation to `generateManifest()`, if the aforementioned PR doesn't fix it, this is ready to go.

claviska (dismissed):
I'm fine with the update to `custom-elements-manifest.js` but I'm not sure we want to put verifications directly in `build.js`. If we're just filtering unnamed events, the plugin seems to handle that well enough to make the verification redundant.

I'd suggest reverting `build.js` and, if any of that code is actually executing, fix the plugin to ensure it works properly.

packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -13,6 +13,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 <small>TBD</small>
 
 - Added `wa-button` class for styling `<a>` elements as buttons [pr:2040]
+- Fixed a bug that was causing errors when some of our components were used in stencil. [pr:2026]
```
KonnorRogers: I don't think this is Stencil specific, I think this is any rendering library using JSX, and this would also generate a manifest with undefined names for events which may break any manifest parsing.


```suggestion
- Fixed a bug in the custom elements manifest where events may not have a name. [pr:2026]
```

KonnorRogers (dismissed):

KonnorRogers (approved):

KonnorRogers (approved):

PR #2039, opened 2026-02-06
Lit change-in-update warnings from many components #1269

### LINKED ISSUE #1269: Lit change-in-update warnings from many components
### Describe the bug
When running in dev mode, many wa-* components cause Lit to warn about "scheduled an update after an update completed" when first rendered.

The equivalent Shoelace components did not issue this warning.

### To Reproduce

1. Create a custom LitElement that renders some Web Awesome components, e.g., wa-button
2. Run a development server and view your project
3. Examine the browser console:

Results:
> Element wa-button scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.

Expected: did not generally occur with Shoelace. (There were some specific component/property value combos that could cause that warning, but not just rendering components in the initial state.)

### Demo

Here's a codepen that loads Lit in dev mode and renders wa-button and sl-button side by side. Notice the console output warns about wa-button but not sl-button.
https://codepen.io/mvedmunds/pen/azvWxYq?editors=0011

### Screenshots

Image

### More information
At a minimum, these components are affected in 3.0.0.beta-4:
- wa-button
- wa-card
- wa-checkbox
- wa-input
- wa-radio
- wa-radio-group
- wa-select
- wa-slider

(I'm using lit@3.3.1, but saw the same warning with earlier versions.)

kelseythejackson:
Fixes #1269 - Lit "change-in-update" warnings that occur when components set reactive properties after an update cycle completes, causing inefficient re-renders.                                                                                                                                                                                                                
   
  Changes
---                                                                                                                                                                                                                                                                                                                                                         
 - `wa-card`: Move slot detection from `updated()` to `willUpdate()` so property changes are part of the current render cycle
  - `wa-slider`: Move value clamping logic to `willUpdate()` while keeping form value updates in `updated()`
  - `wa-checkbox`: Remove duplicate `@watch('defaultChecked')` handler since `willUpdate()` already handles this case
  - `wa-radio-group`: Add change detection before setting child properties (forceDisabled, checked, tabIndex) to avoid triggering unnecessary updates
  - wa-select: Remove `@state()` decorator from optionValues since it's only used internally for filtering, not for rendering

Why these changes
---
  Lit's development mode warns when a component schedules a new update after one just completed. This typically happens when:
  1. Reactive properties are set in `updated()` instead of `willUpdate()`
  2. Properties are marked as `@state()` when they don't actually affect rendering
  3. Child element properties are set unconditionally during parent updates

Testing
---

  All existing tests pass.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -13,6 +13,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 
 ## Next
 
+- Fixed a bug in `<wa-card>`, `<wa-slider>`, `<wa-checkbox>`, `<wa-radio-group>`, and `<wa-select>` that occurred when reactive properties were set after update cycles completed [pr:1269]
 - Fixed a bug `<wa-color-picker>` that prevented it from flipping horizontally when position to the right of the viewport. [pr:2024]
 
 ## 3.2.1
--- a/packages/webawesome/src/components/card/card.ts
+++ b/packages/webawesome/src/components/card/card.ts
@@ -57,7 +57,7 @@ export default class WaCard extends WebAwesomeElement {
   @property({ reflect: true })
   orientation: 'horizontal' | 'vertical' = 'vertical';
 
-  updated() {
+  protected willUpdate() {
     // Enable the respective slots when detected
     if (!this.withHeader && this.hasSlotController.test('header')) this.withHeader = true;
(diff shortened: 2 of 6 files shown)


claviska:
Looks like there's some overlap between this and https://github.com/shoelace-style/webawesome/pull/1284. We should reconcile and then manually test all affected components, as test coverage for some components is currently questionable.

PR #2043, opened 2026-02-07
Fix import of `wa-popup` in dropdown.ts

farmio:
Import to register custom element `wa-popup` used to render the dropdown.

There is no 
```ts
  static dependencies = { 'wa-popup': WaPopup };
```
in this module, like it is eg. in https://github.com/shoelace-style/webawesome/blob/080f96f496d6ca31149c7684eb70ea35e4ec5fc0/packages/webawesome/src/components/tooltip/tooltip.ts#L41
Would this be an alternative / also required?

### DIFF
--- a/packages/webawesome/src/components/dropdown/dropdown.ts
+++ b/packages/webawesome/src/components/dropdown/dropdown.ts
@@ -17,7 +17,8 @@ import { LocalizeController } from '../../utilities/localize.js';
 import type WaButton from '../button/button.js';
 import '../dropdown-item/dropdown-item.js';
 import type WaDropdownItem from '../dropdown-item/dropdown-item.js';
-import WaPopup from '../popup/popup.js'; // Added import for wa-popup
+import '../popup/popup.js';
+import type WaPopup from '../popup/popup.js';
 import styles from './dropdown.styles.js';
 
 const openDropdowns = new Set<WaDropdown>();


claviska:
> There is no
>
> ```
>  static dependencies = { 'wa-popup': WaPopup };
> ```
> in this module

Nice catch. I don't believe we use that static property anymore, so they should be safe to remove. Pinging @KonnorRogers to verify.

claviska:
This PR appears to just split the existing import into two separate ones, but they are functionally the same. The static dependencies variable is no longer used, so I'm going to close this out.

farmio:
The issue was that the custom element `wa-popup` wasn't in scope when loaded when `WaDropdown` was used. Thus it didn't render the dropdown button, but only the slotted elements that should popup when clicked.
As soon as another element loads `wa-popup` as dependency this starts to render correctly.
See https://github.com/home-assistant/webawesome/pull/30 for an example which uses WaDropdown here https://github.com/home-assistant/frontend/blob/7b0e4651c44f235036690bb68b5d784e0f9a01a6/src/components/ha-dropdown.ts#L27

When imported like `import '../popup/popup.js';` `wa-popup` custom element is registered fine and works as expected.

Would you know why `import WaPopup from '../popup/popup.js';` didn't register the `wa-popup` element?

claviska:
> Would you know why import WaPopup from '../popup/popup.js'; didn't register the wa-popup element?

If you're tree shaking, it's possible the side effect is getting removed at build time. Otherwise, either import will trigger a side effect under normal ESM usage.

KonnorRogers:
> When imported like import '../popup/popup.js'; wa-popup custom element is registered fine and works as expected.
> Would you know why import WaPopup from '../popup/popup.js'; didn't register the wa-popup element?

```js
import WaPopup from '../popup/popup.js';
```

is treated fundamentally differently than 

```js
import '../popup/popup.js';
```

by bundlers.

the first one generally can get tree shaken as its assumed to not have side effects and if there is no calls to it, then the bundler will tree shake it.

However, the 2nd one is treated as a "side effect only" import and most bundlers will not attempt to tree shake it for that reason.

farmio:
Thanks!
I'm by no means a frontend expert, but this sounds to me like we'd want to use `import '../popup/popup.js';` here so the component is bundler-friendly, don't we?

claviska (approved):

claviska:
Approved, but I worry about regressions and other components falling into the same gap. A lint rule would be helpful.

PR #2044, opened 2026-02-07
POC: Refactor button groups using pure CSS

### LINKED ISSUE #1975: Outlined `wa-button`s in `wa-button-group` do not collapse
### Describe the bug

When using outlined buttons within a button group, the facing borders do not collapse.\
This results in a doubled border where the buttons meet.

### To Reproduce

Create a button group with at least two buttons:

```html
<wa-button-group>
  <wa-button size="small" appearance="outlined" variant="neutral">
    Some action
  </wa-button>
  <wa-button size="small" appearance="outlined" variant="neutral">
    Another one
  </wa-button>
</wa-button-group>
```

This results in a doubled border between the two:

Image

A closer look:

Image

davidenke:
This is the plan I made up to change the button styling when grouped.
I tried to group the changes in the commits accordingly, so one can go through step by step.

For comparison and to play around with some edge cases, I added some more examples to the button group docs for the time being. These can easily be removed by dropping / reverting 4fd9bbf.

---

## Prerequisites

These prerequisites are based on the current implementation of button groups. To prevent breaking changes, the following conditions must be met after the refactor:

- Tightly coupled buttons to button groups
- Grouping is applied initially and at runtime
- Single nested buttons are supported (current `findButton` logic), grouping multiple buttons is currently not supported

## Goal

- Fix race condition in button detection (`appearance === 'outlined'`)
  - Will fix the issue of outlined buttons not collapsing #1975
  - Will fix the issue of `filled-outlined` buttons not collapsing
- Simplify the internal API being independent of JavaScript detection logic
  - Prevent initial flesh of grouped buttons on page load
  - No expensive DOM traversal for styling purposes

## Steps

1. Introduce a convention for internal custom properties to avoid conflicts with user facing CSS variables. For example, prefix internal properties with `--internal-wa-*`, use triple dash `---wa-*` or something derived from js conventions like `--_wa-button-*` or `--$wa-button-*`.
   > For now, we just use an underscore `--_wa-button-*` to indicate internal properties, however, this can be changed easily later on.
2. Prepare `wa-button` to be styled internally from `wa-button-group` by introducing internal custom properties for indentation and border radius overrides:

   ```css
   consumed-in-button {
     /* Indentation overrides for grouping generic buttons (replacing the button group `gap`) */
     --_wa-button-horizontal-indent: 0;
     --_wa-button-vertical-indent: 0;

     /* Indentation overrides for grouping outlined buttons */
     --_wa-button-horizontal-indent-outlined: 0;
     --_wa-button-vertical-indent-outlined: 0;

     /* Border radius overrides for grouping */
     --_wa-button-start-start-radius: 0;
     --_wa-button-start-end-radius: 0;
     --_wa-button-end-start-radius: 0;
     --_wa-button-end-end-radius: 0;
   }
   ```

3. Introduce css selectors in `wa-button-group` to set custom properties to be inherited by slotted buttons. Make sure to:
   - replace group `gap` with button specific indentions
     - address outlined button individually
   - border radius overrides
     > Do not remove all border radius and add them back, but remove the _unneeded ones_ directly: Buttons have to know which exact radius their configurations need (pill, size dependent, etc.), the group just knows that some radii have to be _removed_.
4. Check some specific stuff not related to spacing or border radius, such as the vertical button group flex behavior and button width:

   ```css
   :host(.wa-button-group__vertical) {
     flex: 1 1 auto;
   }

   :host(.wa-button-group__vertical) .button {
     width: 100%;
     justify-content: start;
   }
   ```

   > These two pieces lead to a strange behavior I can't explain myself:\
   > The button hosts are forced to stretch across vertical button groups and align its contents (the actual buttons) to the top.\
   > Maybe the idea was to either stretch the whole buttons to cover all of the groups height if set explicitly, or it relates to a specific use case I don't get right now. Or is it some browser specific styling for flex items that I'm not aware of?\
   > @lindsaym-fa do you know what these styles were supposed to achieve?

### DIFF
--- a/packages/webawesome/docs/docs/components/button-group.md
+++ b/packages/webawesome/docs/docs/components/button-group.md
@@ -5,22 +5,69 @@ layout: component
 category: Actions
 ---
 
+## Special cases not tracked by examples (remove after refactor)
+
+Slotted buttons might be nested in other elements, such as divs. The button group should still be able to detect these buttons properly.
+
 ```html {.example}
 <wa-button-group label="Alignment">
-  <wa-button>Left</wa-button>
+  <div>
+    <wa-button>Left nested</wa-button>
+  </div>
   <wa-button>Center</wa-button>
-  <wa-button>Right</wa-button>
+  <div>
+    <wa-button>Right nested</wa-button>
+  </div>
 </wa-button-group>
 ```
 
-## Examples
-
-### Vertical Button Groups
-
-Set the `orientation` attribute to `vertical` to make a vertical button group.
+A negative example where multiple buttons are nested within the same div. That didn't work before and should produce the same result after the refactor (no grouping).
 
 ```html {.example}
-<wa-button-group orientation="vertical" label="Options">
+<wa-button-group label="Alignment">
+  <div>
+    <wa-button>Left nested</wa-button>
+    <wa-button>Inbetween nested</wa-button>
+  </div>
+  <wa-button>Center</wa-button>
+  <div>
+    <wa-button>Right nested</wa-button>
+  </div>
+</wa-button-group>
+```
+
+Outlined or `filled-outlined` buttons should collapse after refactor.
+
+```html {.example}
+<wa-button-group label="Alignment">
+  <div>
+    <wa-button appearance="outlined">Left outlined</wa-button>
+  </div>
+  <wa-button appearance="outlined">Center outlined</wa-button>
+  <div>
+    <wa-button appearance="outlined">Right outlined</wa-button>
+  </div>
+</wa-button-group>
+
+<p></p>
+
+<wa-button-group label="Alignment">
+  <div>
+    <wa-button appearance="filled-outlined">Left outlined</wa-button>
+  </div>
+  <wa-button appearance="filled-outlined">Center outlined</wa-button>
+  <div>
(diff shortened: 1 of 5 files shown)


lindsaym-fa (commented):
Great work, @davidenke!

Only one minor suggestion: We reserve the `wa` prefix in custom properties for those defined at the global scope. For custom properties scoped to components, that prefix isn't necessary. i.e. `--_button-start-start-radius`, `--_button-horizontal-indent`, etc. would be preferred for each of these new, internal custom properties.

As for this:
```css
:host(.wa-button-group__vertical) {
  flex: 1 1 auto;
}

:host(.wa-button-group__vertical) .button {
  width: 100%;
  justify-content: start;
}
```

...I'm stumped, even after a winding journey through our commit history. The `justify-content: start;` is the only style that seems useful since it aligns button labels to the start of the button container, which we see the benefit of in button groups with icons in the `start` slot, à la

Image

That said, we seem to be safe to remove these styles and default to center alignment for the button labels instead. No complaints or changes necessary from what you have now.

lindsaym-fa (approved):
I'll merge this in and follow-up with a PR to remove the test docs examples and rename those custom properties. Thanks again for the solid work, @davidenke!

PR #2049, opened 2026-02-09
Fix wa-size utility classes on native buttons in Shoelace theme

### LINKED ISSUE #1892: `wa-size-*` not working on native buttons styles in shoelace theme
`wa-size-*` not working on native buttons styles in shoelace theme.
The `font-size` for native button styles in shoelace.css overides the `font-size` set in `wa-size-*`
https://github.com/shoelace-style/webawesome/blob/7e5f18ea97953c69bcbc692c76351458256eac82/packages/webawesome/src/styles/themes/shoelace.css#L375-L380

### To Reproduce

1. Go to https://webawesome.com/docs/resources/visual-tests#button-2
2. Set the theme to Shoelace
3. Look for the sizes in the button section

AR:
All the buttons are the same size.

ER:
The buttons should be their respective size.

### Screenshots

Image

kelseythejackson:
Fixes #1892

`wa-size-*` utility classes were not working on native button elements in the Shoelace theme due to a layer conflict. The Shoelace theme applies `font-size: var(--wa-font-size-smaller)` to buttons in `@layer wa-theme-overrides`, which comes after `@layer wa-utilities` in the layer order, so the theme always wins over the size utilities. 

Added `:not(.wa-size-s, .wa-size-m, .wa-size-l)` to the native button and radio-button selectors so the font-size override is skipped when a sizing utility class is present. This only affects the Shoelace theme. The `wa-button::part(label)` selector was left unchanged since it targets the shadow DOM label part, which correctly inherits and scales relative to the host element's font-size.

Result
---
Screenshot 2026-02-09 at 11 54 52 AM

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -15,6 +15,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 
 - Added `wa-button` class for styling `<a>` elements as buttons [pr:2040]
 - Fixed a bug `<wa-color-picker>` that prevented it from flipping horizontally when position to the right of the viewport. [pr:2024]
+- Fixed a bug where `wa-size-*` was not working on native buttons styles in shoelace theme. [pr:2049]
 
 ## 3.2.1
 
--- a/packages/webawesome/src/styles/themes/shoelace.css
+++ b/packages/webawesome/src/styles/themes/shoelace.css
@@ -373,9 +373,9 @@
     }
 
     wa-button::part(label),
-    wa-radio[appearance='button'],
-    button,
-    input:where([type='button'], [type='reset'], [type='submit']) {
+    wa-radio[appearance='button']:not(.wa-size-s, .wa-size-m, .wa-size-l),
+    button:not(.wa-size-s, .wa-size-m, .wa-size-l),
+    input:where([type='button'], [type='reset'], [type='submit']):not(.wa-size-s, .wa-size-m, .wa-size-l) {
       font-size: var(--wa-font-size-smaller);
     }
 
--- a/packages/webawesome/src/styles/themes/shoelace.css
+++ b/packages/webawesome/src/styles/themes/shoelace.css
@@ -372,14 +372,52 @@
       color: var(--wa-color-neutral-on-loud);
     }
 
-    wa-button::part(label),
+    wa-button {
+      &::part(label) {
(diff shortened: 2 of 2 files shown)


lindsaym-fa (commented):
We'll need a different approach here, though I'm not quite sure what that looks like myself.

In Shoelace, button font sizes were one step smaller than the button's identified size (see [More Design Tokens > Buttons](https://shoelace.style/tokens/more#buttons)). The intention behind `font-size: var(--wa-font-size-smaller)` is to do just that — drop the label's font size to be one step smaller than dictated by the `wa-size-*` class. But, as you noted already, that doesn't quite work for native button elements because both the theme and the utility classes are setting the same property on the same scope.

This change means that while unsized buttons (which are assumed to match `wa-size-m`, equivalent to the default `size="medium"` for `<wa-button>`) use a smaller font size as intended, using any of the `wa-size-*` classes explicitly doesn't have the intended effect. Similarly, unsized buttons won't match `wa-size-m` buttons as expected.

I believe the path forward here is to set a specific `font-size` for each size class and attribute, rather than trying to rely on a single rule to cover every case. Something like:
```css
wa-button,
wa-radio[appearance='button'],
button,
input:where([type='button'], [type='reset'], [type='submit']) {
    &[size='small'],
    &.wa-size-s {
        font-size: var(--wa-font-size-xs);
    }
    &, /* targets buttons with no size set */
    &[size='medium'],
    &.wa-size-m {
        font-size: var(--wa-font-size-s);
    }
    &[size='large'],
    &.wa-size-l {
        font-size: var(--wa-font-size-m);
    }
}
```

BUT that unfortunately doesn't quite get us there either. `--wa-form-control-padding-block`, `--wa-form-control-padding-inline`, and `--wa-form-control-height` all use relative sizes to respond proportionally to the applied font size. In all themes, we assume these values will enforce uniform height and padding across inputs and buttons. That assumption doesn't pan out in the Shoelace theme because the font size used for buttons is smaller than the font size used for inputs. 

I'm definitely open to suggestions, but for this fix, we need two conditions to be met for `<wa-button>` and native buttons in the Shoelace theme: 
1. The font size of the button label needs to be one step smaller than the button's identified size
2. The height and, ideally, inline padding of the button needs to exactly match the height and padding of inputs given the same size (e.g. `<wa-input size="small">` needs the same height as `<wa-button size="small">`)

The fact that we can't target the inner text of native buttons like we can `<wa-button>` labels is definitely a challenge... 🤔

lindsaym-fa (approved):
Solid! 💯

PR #2064, opened 2026-02-12
Fix for dialog and drawer using user agent default colors

### LINKED ISSUE #1978: Dialog and drawer use UA default colors
### Describe the bug

wa-dialog and wa-drawer render their content using the UA stylesheet's default `dialog { color: canvastext; }`. This results in black (or white) text, which won't match a custom `--wa-color-text-normal`.

(Those components *do* use Web Awesome design tokens for dialog background color.)

### To Reproduce

1. Set `:root { --wa-color-text-normal: blue; }` (for example)
2. Open a dialog or drawer with some text in it
3. Observe the dialog or drawer content is black (or white in dark mode) rather than expected blue

### Demo

https://codepen.io/mvedmunds/pen/PwzKeZx?editors=1000

### Screenshots

wa-dialog with azure background and canvastext color text

Expected:
wa-dialog with azure background and blue color text

### Browser / OS
Web Awesome 3.1.0
Chrome, Firefox, Safari

### Additional information

Workaround:

```css
wa-dialog::part(dialog),
wa-drawer::part(dialog) {
  color: inherit; /* or var(--wa-color-text-normal); */
}
```

kelseythejackson:
Fixes #1978
---
- Fixes dialog and drawer components using the user agent default text color (`canvastext`) instead of respecting  design tokens
- Adds `color: inherit` to the `wa-dialog` and `wa-drawer` styles so they inherit the text color from the document context rather than the browser default

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -15,6 +15,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 
 - Added `wa-button` class for styling `<a>` elements as buttons [pr:2040]
 - Fixed a bug `<wa-color-picker>` that prevented it from flipping horizontally when position to the right of the viewport. [pr:2024]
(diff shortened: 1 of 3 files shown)


claviska (approved):
This is a valid bug _because_ we're setting the background color and NOT the text color.

PR #2068, opened 2026-02-16
Fix animateWithClass for 0ms duration

wendevlin:
If you set `--show-duration` for example to respect reduced motion the dialog events doesn't fire correctly.
We had this issue at least on Windows PCs.

With this fix I check if the element actually has animation if not resolve the animation. 
I tested it on Windows.

I also removed the TODO because this should now fulfill all suitable cases.


wendevlin:
This doesn't work as expected always, I need to do a fix

claviska (approved):
Thanks. I'll add a changelog and get this in.

PR #2070, opened 2026-02-17
Fix popup arrow overlapping text

TCWORLD:
This is a rehash of #1437 but with additional changes to avoid the visual artefacts that meant that PR could not be merged.

I'm opening this again as the original change was manually applied in the home-assistant fork and keeps getting lost whenever the version is upgraded, so it would be good to get it pushed back upstream.

---

The backstory for why is the same as in the linked PR, because the arrow on the popup element is a diamond not a triangle, it interferes with the content of the popup:

image

The original fix was to add a clipping mask to the diamond to turn it into a triangle. This however caused problems when there was a border on the host element - the cropping was to the outside edge of the host border, leaving the border visible.

To resolve this, the arrow now has some additional styling applied to grow and translate it to overlap and thus hide the border of the host. This calculation is done by specifying `--popup-border-width` for the popup element (given as width of border). The arrow then takes care of scaling itself up to overlap to the inner edge of the border.

To visualise it, here is the arrow showing on the popover element that caused problems last time, filled in red to see how it is cropped:

image 
image

Scaling up the size of the border to demonstrate the accuracy of the maths, we get:

image

And just normally:
image


claviska:
Thanks for the PR.

- Can we add a changelog entry about this and highlight the new public custom property `--arrow-border-size` (it looks like the others that were added are computed/internal)
- Can we fix the popup example? When arrow is enabled, it should show up here but it doesn't.
   CleanShot 2026-02-17 at 12 48 57@2x

We should probably also add an example demonstrating how to use this to the popup docs, maybe with the arrow border showing a different color. I'm happy to add the example and the changelog if you don't mind making sure the examples are working as intended.

TCWORLD:
I'll have a look at why the popup arrow is not working - it was on my dev setup, but I may have changed something.

The `--arrow-border-size` (perhaps it should be `width` for consistency) was added so that if an element tries to add a border to its popup, the size can be calculated to overlay the border, happy to add an explanation of that to the docs also. 

I'll do some further tinkering tomorrow.

TCWORLD:
@claviska Fixed the missing arrows on the popup - the joys of CSS treating 0 and 0px as different things for size calculations!

I've also tried to make a slight tweak to the clipping mask to minimise/remove artefacts caused by sub-pixel rounding, but I think there is more going on - probably related to the translation of the arrow when selecting start/end. I'll investigate.

TCWORLD:
Ok, I think I've got it working properly now. I missed that the popup.ts file modifies the position of the arrow which was meant the translation I made in the CSS file was getting lost hence weird visual artefacts. I've corrected this by making the translation part of the positioning in the popup.ts file.

TCWORLD:
All looks functional now from what I can tell.

I've renamed the new variable to `--popup-border-width` because it's actually not at all to do with the border on the arrow itself, but rather any border applied to the main popup element content. This is because the width of that border needs to be known to position the arrow such that it is covered up and makes the arrow appear to merge with the contents.

I've added the changelog entries as requested. I've also added an explanation of the new variable to the popup.md file.

Also added an example of how to stylise the border of the popup, including setting the `--popup-border-width` parameter and targeting the arrow itself.

packages/webawesome/docs/docs/components/popup.md
```diff
@@ -426,6 +426,90 @@ By default, the arrow will be aligned as close to the center of the _anchor_ as
 </div>
 ```
 
+### Adding a border
+
+Borders can also be added to the popup element by targetting the contents of the `wa-popup` element. This styling can also be extended to the arrow itself by targetting `.arrow` class in the popup.
```
claviska:
```suggestion
Borders can also be added to the popup element by targeting the contents of the `wa-popup` element. This styling can also be extended to the arrow itself by targeting `.arrow` class in the popup.
```

claviska:
This is looking really good, but we're seeing a pixel or so overlap on `<wa-popover>`. I zoomed in to take the screenshot, but the problem exists at 100% as well (it's just harder to see).

CleanShot 2026-02-25 at 09 40 12@2x

On second look, it's visible in the popup example too.

CleanShot 2026-02-25 at 09 42 07@2x

It seems to be present on all placements, so hopefully this is a simple tweak because everything else looks great!

TCWORLD:
@claviska Unfortunately that's ultimately caused by sub-pixel rounding affects. Because the little arrow is rotated 45 degrees, its dimensions end up being a ever so slightly too large or too small. I tried to compensate by growing the amount it's shifted by which fixed lines from the popup border remaining, but seems to have left the little inward bits of border that you are seeing. 

It's not as noticable on firefox, but seems quite visible in edge, so must be to do with the way the browser behaves. If I remove the fudge factor it looks better on edge but terrible on firefox, though occasionally bad on both.

Instead of trying to be clever and resize/shift the arrow, I've changed it to just grow the clipping mask instead. This makes the arrow a diamond shape instead of a triangle:

image

It does remove the little flecks sticking inward, and still rectifies the problem of overlapping tooltip arrows. Though when the border is large you get flecks into the arrow as shown below, but that happens without this PR anyway, so I guess its just a limitation we'll have to live with.

image

TCWORLD:
Hurrah, a bit trigonometry and head scratching later, I think I've fixed it fully - no fudge factors required. It seems to work now in all four directions with no border, a normal thickness border and also for very thick borders:

imageimage image imageimage

image

image

Tested in both Firefox and Edge, both seem to produce similar clean results at various different zoom levels.

lindsaym-fa (approved):
@TCWORLD, thanks for the diligent and thoughtful contribution!

Your fix is working great from our testing. After chatting with @claviska, we're going to follow-up with an additional change to leverage `--popup-border-width` more broadly so that it directly sets the `border-width` of popups and their arrows in addition to its use in the arrow clipping calculations. That way, users won't need to duplicate their border widths in multiple places.

**We'll tackle this in a separate PR**, so we'll get your change merged meanwhile. Any concerns, just let us know.

Appreciate you! Please give our thanks to any other contributors on the Home Assistant team, too.

PR #2080, opened 2026-02-19
Revamp wa-icon to be able to specify a version data-fa-version

KonnorRogers:
- allows specifying a `data-fa-version` anywhere in the DOM higher up from a `<wa-icon>`, as well as pass along the `<wa-icon>` instance to be able to provide lookups based on the DOM.


packages/webawesome/src/components/icon/library.default.ts
```diff
@@ -120,6 +121,15 @@ function getIconUrl(name: string, family: string, variant: string) {
...
   }
 
+  let version = FA_VERSION
+
+  if (iconElement) {
+    const el = iconElement.closest<HTMLElement>("[data-fa-version]")
+    if (el && el.dataset.faVersion) {
+      version = el.dataset.faVersion
+    }
+  }
```
claviska: I worry that this will get expensive. It would be better to scope it to the icon itself (at the expense of DX) or through a singleton that's configurable, e.g. like `setBasePath()`.

KonnorRogers: Would you prefer we do like `getKitCode()`


and do:

```js
export function getKitVersion() {
  if (!kitVersion) {
    const el = document.querySelector('[data-fa-version]');

    if (el) {
      kitVersion = el.getAttribute('data-fa-version') || FA_VERSION
      setKitVersion(kitVersion);
    }
  }

  return kitVersion;
}
```

https://github.com/shoelace-style/webawesome/blob/89d22142a79336a8a7a4529ace38b52b15b9c6b6/packages/webawesome/src/utilities/base-path.ts#L70-L80

claviska: It seems reasonable as a one-time set. And better than DOM traversal. In this case I think perf > DX unfortunately.

claviska: In fact, we should probably consider a single `setWebAwesomeConfig()` function at this point instead of cloning the same behavior thrice 😆

KonnorRogers: Probably fair. something like:

```js
setWebAwesomeConfig({
  defaultIconLibrary: "foo",
  fontawesomeIconVersion: "7.8.9",
  fontawesomeKitCode: "123456789"
})
```

🤔

claviska: This seems like a good approach. We can continue exposing the existing utilities but mark them as deprecated. For hosted projects, should we expose this through `window.WebAwesome.setConfig()` or similar? We need to establish a plan for this as well as for importing utilities.

claviska:
@KonnorRogers is this ready for a review?

KonnorRogers:
@claviska no, need to get back to this one and add a couple pieces

packages/webawesome/src/components/icon/library.default.ts
claviska: What if `data-fa-version` is only discovered on page load the same was `data-wa-preload` works? It's a cheap one-time selector to figure out what version to use. (I don't think supporting multiple versions of FA on the same page makes a lot of sense so this limitation seems reasonable.)

I'm fine with something like `setIconVersion()` (name TBD). I don't think I actually care much about a single config function at the moment (and we can add that anytime in the future if we want it).

Slightly related, but I did some work today to improve self-hosting FA icons with `<wa-icon>`: https://github.com/shoelace-style/webawesome/pull/2240

claviska: At this point the separate function like you originally suggested is probably better than a combined config. With our move to recommending the CDN it makes more sense, as we don't need a central function (which was a thought for importing/setting things from hosted projects).

PR #2082, opened 2026-02-19
Add `start` and `end` slots to `<wa-badge>`

### LINKED ISSUE #2072: Icons within badges are styled
### Describe the bug
The documentation doesn't explain treatment of icons in badges but it is styling them anyway.

https://github.com/shoelace-style/webawesome/blob/39d657a2de0de3740a94975a62b7fa1caf4124ff/packages/webawesome/src/components/badge/badge.styles.ts#L102-L107

### Expectation
The feature should be documented or the icon unstyled.

### To Reproduce

1. Add any icon to a badge.
1. See that the icon is styled.

### Demo

```
<wa-badge variant="brand"><wa-icon name="home" auto-width label="Favorite"></wa-icon>Badge</wa-badge>
```

lindsaym-fa:
Closes #2072 by:
- Removing styles for slotted `<wa-icon>` elements in favor of `start` and `end` slots for opt-in, reliable styling.
- Adding `vertical-align: middle` to `<wa-badge>` for consistent alignment regardless of the badge's content. (The original styles attempted to accommodate this by resizing slotted icons.)

---

### Why Slots?

The initial code I set out to remove was, in part, applying `margin-inline-end: var(--wa-space-2xs, 0.25em)` to slotted icons. This is needlessly opinionated and fragile; it assumes icons are only placed before other badge content.

Elsewhere, we do try CSS-only solutions to apply default icon spacing according to whether the icon is at the beginning or end of the supplied content. Take this example from `<wa-tab>`:

```css
::slotted(wa-icon:first-child) {
  margin-inline-end: 0.5em;
}

::slotted(wa-icon:last-child) {
  margin-inline-start: 0.5em;
}
```

But this doesn't actually work. Because content in the default slots of components like `<wa-tab>` are typically plain text nodes, including `<wa-icon>` means that the icon matches both `:first-child` _and_ `:last-child`. i.e., we get this:
Screenshot 2026-02-19 at 6 37 12 PM

Slots give us much more reliable control of spacing, so it felt like the best approach for `<wa-badge>` and seems worth considering for other components, too.

@claviska, I'm 100% open to pushback if you think this is overkill!


claviska (approved):
🐄 🛸

PR #2086, opened 2026-02-22
Add Angular component wrapper generator script

heddendorp:
This is a draft idea of what angular wrappers could look like based on the way the react wrappers are created, still figuring out and will test locally before moving on


claviska:
Hey thanks for the PR. You're welcome to keep working on this but I don't think we're going to be able to maintain it as a first-party extension. However, third-party extensions can use the Web Awesome CEM to do exactly what this is doing, just as a separate project.

Thanks again for the PR!

Issue #2018, opened 2026-02-04
`wa-split-panel` throws errors when resizing


HerrNaN:
### Describe the bug
When resizing a `wa-split-panel` with `primary` set, and the non-primary panel gets "squished" a `ResizeObserver` error gets thrown on the `Window` object.

> [!NOTE]
> This issue does **not** seem to affect Firefox 147

### To Reproduce

1. Open CodePen
2. Move divider close to the text "Start" (but not all the way)
3. Make browser window (page) narrower such that the divider squishes the text "Start".
4. See error in CodePen console (not the browser console)

https://github.com/user-attachments/assets/1b5f7a18-1789-4383-a09a-5983efe29565

### Demo

CodePen: https://codepen.io/herrnan/pen/JoKZxZd

### Screenshots
N/A

### Browser / OS
 - OS: Linux, Ubuntu 24.04
 - Browser: Brave (Chromium)
 - Browser version:
   - Brave: 1.86.146
   - Chromium: 144.0.7559.110

### Additional information
N/A

kelseythejackson:
Thanks for raising this issue. We'll investigate to see what we can do

kelseythejackson:
@HerrNaN Hey, following up. We've not been able to reproduce this error on our end. Is it still occurring for you? Thanks.

HerrNaN:
> @HerrNaN Hey, following up. We've not been able to reproduce this error on our end. Is it still occurring for you? Thanks.

I will check as soon as I can! 👀

HerrNaN:
> @HerrNaN Hey, following up. We've not been able to reproduce this error on our end. Is it still occurring for you? Thanks.

@kelseythejackson I can still reproduce it the same way as described in the issue and shown in the video.

Note that this never was reproducable on Firefox as far as I'm aware.

claviska:
The problem appears to be with `position` and `positionInPixels`. When the panel resizes, each property's watcher was triggering the other, which caused enough layout churn within a single frame to trip Chrome's ResizeObserver loop detection. The fix adds guards so that neither watcher fires if the value hasn't actually changed.

claviska:
Fixed in #2092

PR #2092, opened 2026-02-25
Split panel resize bug

claviska:
Fixes the bug described in #2018


kelseythejackson (approved):
These changes look like they should avoid looping. Approved.

Issue #2085, opened 2026-02-22
Escape key closes multiple levels of components


jwillesen:
### Describe the bug

If you have nested components that have an “open” state, and you press the escape key while focus is on a nested component, the outer component also closes unexpectedly. I would expect only the nested component’s interface to close. For example, if you have a `<wa-select>` inside a `<wa-drawer>`, and the select has focus and its menu is open, pressing escape should only close the select’s menu and leave the drawer open. Instead, the drawer also closes unexpectedly.

I think this is an accessibility concern. A keyboard user would not expect escaping an open menu to close multiple levels of the UI, and now may have to reproduce lost state.

### To Reproduce

1. Go to [this CodePen](https://codepen.io/jwillesen/pen/KwgPwGz)
2. Click on 'Open Drawer'
3. Click on the ‘Open me’ <wa-select> in the drawer to open the menu and give it focus.
4. Press ‘escape’ on the keyboard. Notice that the drawer closes. Only the open select menu should have closed.

### Demo

Here is the [CodePen](https://codepen.io/jwillesen/pen/KwgPwGz) that reproduces the issue.

### Browser / OS
 - OS: Mac, iOS
 - Browser: Chrome, Mobile Safari
 - Browser version: -

### Additional information

The problem is not just limited to drawers and selects. For example, if you have a `<wa-dialog>` inside a `<wa-drawer>` (whether this is good design is beside the point), and you press “escape” while the `<wa-dialog>` is open, it will close both the dialog and the drawer. I would expect it to just close the dialog.

kelseythejackson:
Hey, thanks for raising this issue. We'll take a look and see if we can get it sorted out.

PR #2096, opened 2026-02-26
Implement Dismissible Stack

kelseythejackson:
This fixes the issue raised in #2085 

This PR adds a shared function that keeps track of when listeners are added to the document to avoid keydown conflicts. This change touches several components.


claviska (approved):
Well done. I did some refactoring, fixed some missing guards/checks, and added tests to affected components.

I also renamed it to `dismissible stack` which feels better than "overlay stack" because not all components that need it require an overlay but all focus on the dismissal behavior.

I'm working on a companion PR for combobox in `webawesome-pro`, but this is ready to go.

PR #2105, opened 2026-02-27
fix checkbox + input issues and some SSR issues

### LINKED ISSUE #1205: `<wa-input>` stays invalid when updating `value` property
### Describe the bug
When rendering a  `required` `<wa-input>` without an initial `value` and then later setting the `value` prop the input stays invalid.

### To Reproduce

https://codepen.io/kasper-i/pen/ogjBKOv

1. Render `<wa-input>` without a `value` but having `required` set
2. Update `value` with a non-empy string
3. The new value is displayed in the input field
4. Try submitting the form and the browser promts you to fill out the field
5. Submit again and the validation passes

### Browser / OS
 - OS: Linux
 - Browser: Firefox<del>, Chrome<del>
 - Browser version: 138.0.1<del>, 128.0.6613.137<del>

EDIT: Looks like the issue is only present in the Firefox browser.

### LINKED ISSUE #1230: Vue 3 template wa-switch with checked attribute not drawing checked
### Describe the bug

In a Vue 3 component, using a wa-switch in our template, setting the checked attribute does not result in a checked switch.  However, other attributes (like disabled) seem to be working fine.

### To Reproduce

https://jsfiddle.net/qj9vk4et/

### Browser / OS
 - OS: Linux
 - Browser: Chrome
 - Browser version: 138.0.7204.92

### LINKED ISSUE #1273: wa-radio-group and wa-slider don't update when value attribute changed dynamically
### Describe the bug
When wa-radio-group's value attribute is updated, it does not render the new value. Other input controls (wa-select, wa-checkbox, wa-input) *do* update on value attribute changes. 

### To Reproduce

1. Go to https://codepen.io/mvedmunds/pen/zxvwVJq?editors=0010
2. Click on the "Set value to 3" button
3. Observe that all controls *except* the first wa-radio-group update to show the value 3

### Demo

https://codepen.io/mvedmunds/pen/zxvwVJq?editors=0010

### Screenshots

Image

### Additional information
3.0.0-beta.4

sl-radio-button didn't seem to have this issue.

A workaround is to set the wa-radio-group's value *property* rather than its value attribute. (Demonstrated in the codepen. But it seems odd this would be inconsistent with other input controls.)

### LINKED ISSUE #1640: wa-input's internal value is not reset after the form resets
### Describe the bug
When using forms with `<wa-input>`s the inputs don't properly reset their values. Visually they are reset yes, but internally they still seem to hold the old value. This leads to wrong form validation, when submitting the form after resetting it, without changing the input's value.

### To Reproduce

1. Inside a form element place a `<wa-input required></wa-input>`, `<button>submit</button>` and a `<button type="reset">reset</button>`.
2. Leave the input empty and press the submit button. The browser correctly notifies me about the required field.
3. Type something in that field.
4. Press the reset button. The input field is visually empty.
5. Press the submit button. The form submits, even tho the input field is empty.

### Demo

https://codepen.io/Duki-the-typescripter/pen/jEWzdWN

### Browser / OS
 - OS: Mac
 - Browser: Chrome

### LINKED ISSUE #1860: top-level `document.addEventListener` call from wa-drawer
The drawer registers a global event on document, which is not possible when rendering on the server

https://github.com/shoelace-style/webawesome/blob/579e86fc4999cd2f7194823696267457a5c21aa3/packages/webawesome/src/components/drawer/drawer.ts#L287

I believe this could be pulled into the `!isServer` condition just below, or moved into `connectedCallback`?

### LINKED ISSUE #1861: SSR: `HasSlotController` uses client-only APIs (`wa-drawer`, `wa-button`)
Using `HasSlotController` class breaks server-side rendering elements which use it in the `render` method.

My naive solution is to initialize and call it conditionally. Alternatively, I the controller could be modified so that it does not try to attach listeners and does not do DOM queries when running on the server.

Unfortunately, it's obvious that this will cause hydration mismatches in some cases, depending on how a given element is used. The second option is even worse because it ignores individual element's rendering needs.

### LINKED ISSUE #1878: Issue with WaCheckbox wrapper.
### Describe the bug
The checked attribute for WaCheckbox (React wrapper) appears to be not processed when the page is first loaded. This works fine with wa-checkbox.

### To Reproduce
Include the following code:

`<WaCheckbox checked>Checked</WaCheckbox>
`

### Demo
https://codesandbox.io/p/sandbox/gifted-moon-wq2xgr

### Screenshots
Image

### Browser / OS
 - OS: Mac
 - Browser: Chrome
 - Browser version: 143.0.7499.110

### Additional information
Provide any additional information about the bug here.

KonnorRogers:
Fixes #1878
Fixes #1860
Fixes #1640
Fixes #1861
Fixes #1273
Fixes #1205
Fixes #1230


packages/webawesome/src/internal/slot.ts
```diff
@@ -36,19 +41,19 @@ export class HasSlotController implements ReactiveController {
...
-    this.host.shadowRoot!.addEventListener('slotchange', this.handleSlotChange);
+    this.host.shadowRoot?.addEventListener?.('slotchange', this.handleSlotChange);
   }
 
   hostDisconnected() {
-    this.host.shadowRoot!.removeEventListener('slotchange', this.handleSlotChange);
+    this.host.shadowRoot?.removeEventListener?.('slotchange', this.handleSlotChange);
   }
 
   private handleSlotChange = (event: Event) => {
```
KonnorRogers, line 59: 🤔 these _may_ cause hydration mismatches, so may have to return false if `!this.hasUpdated` as a temporary workaround, but would need to test it.

PR #2109, opened 2026-03-02
Support showing full-page patterns in the docs

lindsaym-fa:
Enables us to better present full-page patterns in the docs with the following changes:
- **Adds styles for a file explorer UI** in `utils.css` to use when documenting full page patterns (see also: https://github.com/shoelace-style/webawesome-pro/pull/151)
- **Refactors the docs' search index styles** to:
  - Use container queries instead of media queries
  - Add a “large” grid variant
- **Updates `code-examples.js`** to support:
  - Parsing top-level `<wa-include>` elements in the same way we can parse `<wa-zoomable-frame>` using the attribute `data-select-src`
  - Showing code examples without previews using the `no-preview` class (for when we want the opt-in parsing of `<wa-zoomable-frame>` or `<wa-include>` without rendering the results)
  - Optionally expanding nested `<wa-include>` elements using the attribute `data-expand-includes` (this previously happened automatically when specifying a non-boolean `data-select-src` on `<wa-zoomable-frame>`
- **Refactors `code-examples.css`** to:
  - Ensure any combination of `.no-preview` and `.no-buttons` renders without looking broken
  - Use logical properties instead of physical properties where possible

---

These changes come to life in https://github.com/shoelace-style/webawesome-pro/pull/151, so I strongly recommend pulling both down together.

@talbs, mind checking that my file explorer styles are sound and that I've given them the right home?
@claviska, mind lending your expertise to review my changes to `code-examples.js`?


packages/webawesome/docs/assets/styles/utils.css
```diff
@@ -321,6 +321,61 @@
       content: ' (' attr(href) ')';
     }
   }
+
+  /* patterns file explorer */
+  wa-zoomable-frame:has(+ .file-explorer) {
```
talbs, line 326: Holy Level 12 Mage Selector, Batman! 🦇

talbs (dismissed):
Wow! Really nice work, @lindsaym-fa.

Your code and UI look very solid, and your choices make sense. I did have some NIT-level visual feedback. I'll leave each thread as a separate comment.

Feel free to address those as you'd like - but none are blockers to merging this.

talbs:
CleanShot 2026-03-02 at 20 36 27@2x

☝️ Can the border radius of the zoom controls be made more consistent with the border radius of the frame?  Right now, the small, almost square-ish shape really sticks out.

talbs:
CleanShot 2026-03-02 at 20 44 08@2x

I I know it hurts legibility, but breaking file names into multiple lines feels off for this piece of UI.

Can the indent level of files within directories be tightened up here (to better match or be a smaller multiple of the directory -> directory indent)? If we can squeeze more characters in that way, could truncating/clipping text to a single line be acceptable?

talbs:
CleanShot 2026-03-02 at 20 46 36@2x

Due to the general conventions of file browsers (directories first, files second in alphabetical  order), we're seeing a lot of other files than the main layout one in the file explorer by default. 

When landing on a layout detail view, the main file isn't visible in the explorer. Could the directories be collapsed by default to make sure the main/current is shown? It may also help a user approach the example more easily at first.

lindsaym-fa:
@talbs, thanks for the great suggestions! ✨ All addressed, save this one:

> Can the indent level of files within directories be tightened up here (to better match or be a smaller multiple of the directory -> directory indent)? If we can squeeze more characters in that way, could truncating/clipping text to a single line be acceptable?

Good idea! I want to make sure we present the full file name _somewhere_ in the file explorer UI, so I'm going to do some thinking and address this in a follow-up.

PR #2133, opened 2026-03-06
fix(carousel): re-sync slides when carousel becomes visible in hidden container

### LINKED ISSUE #1164: Button in carousel in tab is only clickable after interacting with carousel
I originally found this issue in Shoelace and [reported it there](https://github.com/shoelace-style/shoelace/issues/2477), but Web Awesome has the same problem.

I'm pasting the issue description below for convenience.

---

### Describe the bug
I have a button in a carousel in a tab. When the page loads, the button is not clickable. It only becomes clickable once I interact with the carousel (move it slightly or switch slides).

### To Reproduce

1. Add a tab group to the page.
2. Add a carousel in one of the tabs.
3. Add a button in the first slide of the carousel.
4. Load the page and try to click the button.

### Demo

https://codepen.io/diegovogel/pen/KwdKoKO/73933bc9cbbdf3a9aa2f578009060ffa

### Screenshots
https://www.loom.com/share/89e33dd0f93a4d9d866ca57bf50de30b

### Browser / OS
 - OS: macOS
 - Browser: Chrome
 - Browser version: 138.0.7204.93

terminalchai:
## This is the problem:

When a `<wa-carousel>` is placed inside an inactive tab panel (or any container initially hidden via `display: none`), the carousel renders and `firstUpdated()` runs before the element has any layout dimensions.

`initializeSlides()` calls `synchronizeSlides()`, which creates an `IntersectionObserver` with `root: this.scrollContainer`. Because the parent container has `display: none` at this point, **all slides are reported as non-intersecting**, so every slide gets `inert` set to `true`. The `IntersectionObserver` then disconnects — and nothing triggers another sync when the tab is later shown.

The result: when the user switches to the tab, the carousel looks correct visually, but **all slide contents (buttons, links, inputs) are unclickable** because they are `inert`. Only after the user manually interacts with the carousel (triggering `handleScroll` → `synchronizeSlides()`) does the `inert` attribute get removed.

## Steps to reproduce:

1. Add a `<wa-tab-group>` to the page.
2. Add a `<wa-carousel>` inside one of the tab panels with a `<wa-button>` in the first slide.
3. Load the page and switch to that tab.
4. Try to click the button — it is not clickable.

Demo: https://codepen.io/diegovogel/pen/KwdKoKO/73933bc9cbbdf3a9aa2f578009060ffa

Fixes #1164.

## This is my solution:

Add a one-shot `ResizeObserver` in `firstUpdated()` that watches the carousel host element. As soon as `scrollContainer` gains non-zero dimensions (i.e. the tab becomes active and the carousel is laid out), `synchronizeSlides()` is called again to correctly mark the visible slides as non-`inert`. The observer then immediately disconnects itself.

```typescript
this.resizeObserver = new ResizeObserver(() => {
  if (this.scrollContainer?.clientWidth || this.scrollContainer?.clientHeight) {
    this.synchronizeSlides();
    this.resizeObserver?.disconnect();
    this.resizeObserver = undefined;
  }
});
this.resizeObserver.observe(this);
```

The observer is also disconnected in `disconnectedCallback()` as a safety net for elements removed from the DOM before becoming visible.

**No behaviour change** for carousels that are already visible on initial render — the ResizeObserver fires immediately with non-zero dimensions, calls `synchronizeSlides()` once, and disconnects.


claviska:
The fix seems fine, thanks for the PR! Just needs a quick changelog entry and it's good to go.

talbs:
Hey, @terminalchai! Thanks again for this contribution.

I'm popping in to say **"pardon us"** while we use this PR as an opportunity to try out some improvements to managing Web Awesome's changelog with some of our agent tooling/workflows.

This won't affect your work or changes - and we'll clean up any messes we make along the way.

---

FYI, since it looks like forked branches don't get along with the workflow we're trying, the next thing I'm going to attempt is to have an agent suggest an addition to the changelog for this work. Hang tight!

talbs:
Okay, we need to spend more time getting our agent and workflow in shape for suggestions/PRs. 😮‍💨

Sorry for the noise, @terminalchai.  I've cleaned up all of our previous comments related to those efforts.

PR #2135, opened 2026-03-06
Refactor CSS overrides in themes

### LINKED ISSUE #1766: Displaced Horizontal Radio Group Item with Customized Awesome Theme
### Description

Using WA Pro with a customzied Awesome theme, a selected item in a horizontal radio group is slightly displaced:

Image

I don't see this phenomenon with the basic theme using the same code.
All the customizations were done using the Theme editor on webawesome.com

This is the code behind:

```html
<wa-radio-group
   label="Billing"
    hint="Save up to 21% with our annual plans"
    orientation="horizontal"
    name="a"
    x-model="billing_period"
>
    <wa-radio appearance="button" value="month">Monthly</wa-radio>
    <wa-radio appearance="button" value="year">Annual (-17%)</wa-radio>
    <wa-radio appearance="button" value="2years">Bi-Annual (-21%)</wa-radio>
</wa-radio-group>
```

As far as I see removing one of the CSS transform rules puts the item back to its place:

Image

### To Reproduce

1. Go to https://spotkit.net/pricing/
2. See the error

### Browser / OS

Tested on:

 - OS: [Android and Linux Desktop]
 - Browser: [Chrome, Firefox]
 - Browser version: latest as of writing [CR 142 / FF 144]

### Additional information

> Might be related to: #1527

lindsaym-fa:
Refactors CSS overrides in Free themes.

#### Awesome
- Rewrites many selectors
- Updates default input padding to better match Font Awesome
- Fixes #1766
- Fixes broken `appearance` for `<wa-card>`
- Adds inner shadow to `<wa-combobox>`

#### Shoelace
- Rewrites many selectors
- Fixes broken `appearance` for `<wa-card>`
- Fixes broken `size` for `<wa-callout>`
- Fixes focus styles to more closely match Shoelace's
- Adds missing `input[type='range']` overrides
- Adds input-specific focus style to `<wa-combobox>`

---

#### Awesome

awesome-button-sizing
awesome-card-appearance
awesome-combobox

#### Shoelace

shoelace-card-appearance
shoelace-callout-size
shoelace-combobox


lindsaym-fa:
@claude, please review.

lindsaym-fa:
@talbs, would you mind giving me a sanity check on this refactor?

I'm especially curious if you think the change to the `--wa-form-control-padding-*` tokens is a good move. The change brings the Awesome theme much closer to Font Awesome, but increasing the button size also means that the height of our docs nav increases significantly. It could have similar consequences for folks who are currently using the Awesome theme or some customization of it in their own projects. Is the change worth it? 🤔 (Perhaps I should hold off on this "improvement" until we allow those paddings to be edited in the theme builder?)

talbs (dismissed):
This is a whole lot of good work, @lindsaym-fa!

The CSS changes look good to me and, using the visual test views, all looked well to me.

> I'm especially curious if you think the change to the `--wa-form-control-padding-*` tokens is a good move. The change brings the Awesome theme much closer to Font Awesome, but increasing the button size also means that the height of our docs nav increases significantly. It could have similar consequences for folks who are currently using the Awesome theme or some customization of it in their own projects. Is the change worth it? 🤔 (Perhaps I should hold off on this "improvement" until we allow those paddings to be edited in the theme builder?)

Hmmm, yeah, there's definitely a big shift in size here. After looking at the buttons on fontawesome.com, this change is more accurate and true to the visual direction.

**I think we should ship** - Anyone using the Awesome theme today gets the same “closer to Font Awesome” controls. If they need smaller controls, they can override the tokens (e.g. `--wa-form-control-padding-*`) until the Theme Builder exposes them. That’s a reasonable tradeoff for the correctness of the theme.

And we can specifically do that in our Docs Nav if we think it is aggregious.

☝️ I'm a 6/10 on that - if you feel strongly, holding this specific change doesn't feel like icky to me.

---

One visual design nit around vertical button groups in the Awesome theme...

Image

The chin-like hard shadow on elements in this theme has a vertical nature - and in the case of Cards, that shadow is applied to the most outward shape (the card's body). Should the same be true in this case (as a shadow on the group itself visually)?

Here's a quick hack mockup to explain the direction...

Image

I may be opening up a can of depth, button group philosophy or theming worms, so feel free to disregard if that's true and you'd rather not go down that road.

lindsaym-fa:
@talbs, wondering how this approach sits with you: 


https://github.com/user-attachments/assets/f43170cf-6f81-4f74-8b66-3c135ea23d53

I'm using the default gap for filled buttons instead of a custom dotted border. Each button maintains some kind of "punch in" effect, but only the last button gets a visible shadow. Anything feel off or uncanny here?

talbs:
> I'm using the default gap for filled buttons instead of a custom dotted border. Each button maintains some kind of "punch in" effect, but only the last button gets a visible shadow. Anything feel off or uncanny here?

Thanks for the clip and demo. This looks good and using the default gap is smaht. Nothing feels off - on the contrary, this feels natural and works for me!

talbs (dismissed):

PR #2139, opened 2026-03-09
Add forked qr-library with support for images

KonnorRogers:
Adds support for image logos in the center of a QR Code as well as having corner colors separate from the main color


packages/webawesome/src/components/qr-code/qr-code.ts
```diff
@@ -49,41 +47,38 @@ export default class WaQrCode extends WebAwesomeElement {
...
-    // We lazy load because the QR generator will cause the server to crash, but we want to reduce layout shift.
-    if (!QrCreator) {
-      import('qr-creator').then(mod => {
-        QrCreator = mod.default;
-        this.generate();
-      });
-      return;
-    }
-
+    console.log("Generating...")
```
claviska:
```suggestion
```

claviska (dismissed):

claviska:
@KonnorRogers I cleaned up the docs and added a simpler image example to demonstrate. Also renamed `image-ec-cover` to `image-coverage` which feels more idiomatic. Let me know if you have any concerns around that.

Otherwise, this is ready to 🚢

PR #2151, opened 2026-03-11
Fix Label Styling When Using the `label` Slot

### LINKED ISSUE #2124: Un/slotted labels have different appearances for Slider
### Describe the bug
#1757 doesn't appear to have fixed labels for sliders

### Demo
https://codepen.io/roydukkey/pen/KwgzQNe?editors=1000

talbs:
In Web Awesome, form control labels (slider, input, select, etc.) can be set either with the `label` attribute or by putting content in the `label` slot. 

Those two options weren’t always styled the same: slotted labels could use different typography and, on the slider, had less space above the track. This PR applies the same label styling and spacing for both attribute and slotted labels.

Fixes https://github.com/shoelace-style/webawesome/issues/2124

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -8,10 +8,13 @@ Web Awesome follows [Semantic Versioning](https://semver.org/). Breaking changes
 
 Components with the <wa-badge variant="warning">Experimental</wa-badge> badge should not be used in production. They are made available as release candidates for development and testing purposes. As such, changes to experimental components will not be subject to semantic versioning.
 
-## 3.3.2
+## Unreleased
+
+<small>TBD</small>
 
 - Fixed the off-centered position of indent guides in `<wa-tree>`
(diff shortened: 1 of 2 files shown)


talbs:
@claude, review this PR.

talbs:
@lindsaym-fa, mind reviewing my proactive bug fix? Let me know if there's a better way to solve this.

lindsaym-fa (dismissed):
The change to `slider.styles.ts` is perf!

I'm not exactly sure what the addition to `form-control.styles.ts` gets us. As far as I can tell, `color`, `font-weight`, and `line-height` already apply to slotted content by way of the preceding rule, and I'm not sure if Claude is hallucinating or onto something by suggesting that they don't. 🤔 BUT it's not problematic and doesn't seem to block more opinionated elements, e.g. `<h2>`, from getting their usual styling, so I'm fine with it.

talbs:
Thanks! 

I'll ditch 3956c7b1ca893af6eb12110cab482aa1b0021c49 and make this PR focus on the slider fix in that case.

PR #2155, opened 2026-03-12
Improve button gaps when wrapping

claviska:
This PR updates the button page in the docs so buttons wrap like this:

CleanShot 2026-03-12 at 13 59 35@2x

Instead of this:

CleanShot 2026-03-12 at 13 59 19@2x

It also turns the examples that don't wrap:

CleanShot 2026-03-12 at 14 02 31@2x

Into ones that do:

CleanShot 2026-03-12 at 14 03 12@2x


lindsaym-fa (approved):
Love a good delivery from the cluster truck! 🚚

PR #2158, opened 2026-03-12
Fix scroller bug that caused docs bug that seemed like toast bug 🙂‍↔️

### LINKED ISSUE #2157: Component docs pages are scrolling horizontally on smaller screens in Chrome
I haven't been able to reproduce this in Safari or Firefox — only Chrome.

1. Go to any component page [such as this one](https://webawesome.com/docs/components/badge).
2. Resize the viewport to about 470px or so
3. Observe horizontal scroll

There is no obvious overflow happening, but the problem resolves when hiding `<wa-scroller>`. Again, only in Chrome.

Image

claviska:
This fixes #2157. It wasn't a toast bug. Wasn't a docs bug. It was an edge case Chrome layout bug because the scroller didn't have the proper overflow on its host.


talbs (dismissed):
👀 good. Getting 🎁🎁🎁.

This seems to fix the issue for me when viewing locally in Chrome. And a quick browse of the scroller component examples shows no side effects.

PR #2160, opened 2026-03-12
Tree sizing

claviska:
Per my suggestion in #2147, this makes `<wa-tree>` scale with the current font size. At the moment, the base font is hard coded making it really hard to resize.

CleanShot 2026-03-12 at 15 21 33@2x
To do this, I needed to update the indent guide logic which previously used a `font-size: 0` trick. As part of this rework, I noticed the indent guides were off centered. This PR also fixes that.

Before
CleanShot 2026-03-12 at 15 13 42@2x

After
CleanShot 2026-03-12 at 15 14 01@2x

Then I got carried away and touched up some of the examples in the docs. 😅

### DIFF
--- a/packages/webawesome/docs/docs/components/tree.md
+++ b/packages/webawesome/docs/docs/components/tree.md
@@ -6,45 +6,39 @@ category: Navigation
 ---
 
 ```html {.example}
-<wa-tree selection="multiple">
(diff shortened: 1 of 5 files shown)


lindsaym-fa (approved):
Flawless. ✨

PR #2161, opened 2026-03-12
Fix Lit warnings

### LINKED ISSUE #1269: Lit change-in-update warnings from many components

claviska:
Fixes #1269


claviska:
@KonnorRogers double check my math on slider because it required a bit of refactoring. The others have the warning disabled with reason, which I think is a good pattern for this.

claviska:
@KonnorRogers looks like the pro components are good...no visible warnings.

PR #2154, opened 2026-03-12
Extend Typography's Font Size Scale + Utilities

talbs:
This work extends Web Awesome’s typography scale by adding `3xs`(10px) and `5xl` (66px) to the `--wa-font-size-*` tokens and to all related utility classes (`wa-body-*`, `wa-heading-*`, `wa-caption-*`, `wa-longform-*`, `wa-font-size-*`), and updates the typography and text docs (including an accessibility note for sub-12px sizes) and the changelog. 

Addresses https://github.com/shoelace-style/webawesome/issues/1606.

PR #2162, opened 2026-03-13
Badge: Set `font-size` using  `––wa-font-size-3xs` 

talbs:
Updates `wa-badge` so `font-size` uses `max(var(--wa-font-size-3xs), 0.75em)` instead of `-2xs`, now that the `3xs` token is available from #2154.

Addresses suggestion from @lindsaym-fa in #2154.

Generated with [Claude Code](https://claude.ai/code)


claviska (dismissed):
At first I thought it would be better to hard code 10px as a guard, but that breaks font scaling. That said, this looks good to me.

lindsaym-fa (dismissed):

talbs:
@claude, update the `changelog.md` to note this change.

PR #2164, opened 2026-03-13
Add label support to color picker swatches

claviska:
Per #2067, this adds the [optional] ability to add labels to color swatches to control how screen readers announce them.

### DIFF
--- a/packages/webawesome/docs/docs/components/color-picker.md
+++ b/packages/webawesome/docs/docs/components/color-picker.md
@@ -60,6 +60,26 @@ Use the `swatches` attribute to add convenient presets to the color picker. Any 
 ></wa-color-picker>
 ```
 
+You can also pass an array of objects with `color` and `label` properties using JavaScript. When labels are provided, they will be used as the accessible name for each swatch instead of the raw color value.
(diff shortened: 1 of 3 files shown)


paulGeoghegan:
This is great, could we get an example in the docs to go with it please?

claviska:
Yep! It's here: https://github.com/shoelace-style/webawesome/pull/2164/changes#diff-74b314898b0403ead713aefd799eb3ed31819a8986b90869364ff7c6aa29b034

packages/webawesome/docs/docs/components/color-picker.md
```diff
@@ -60,6 +60,26 @@ Use the `swatches` attribute to add convenient presets to the color picker. Any
 ></wa-color-picker>
 ```
 
+You can also pass an array of objects with `color` and `label` properties using JavaScript. When labels are provided, they will be used as the accessible name for each swatch instead of the raw color value.
```
talbs, line 63: 💥

talbs (approved):
Nice work - this is a great addition to the API and UI.

PR #2165, opened 2026-03-13
Kj/zoomable frame light dark

kelseythejackson:
This is an attempt to make the zoomable frame adhere to working with light/dark mode as well as our themes.

### DIFF
--- a/packages/webawesome/docs/docs/components/zoomable-frame.md
+++ b/packages/webawesome/docs/docs/components/zoomable-frame.md
@@ -61,3 +61,11 @@ Apply the `without-interaction` attribute to make the frame non-interactive. Not
 ```html {.example}
 <wa-zoomable-frame src="/examples/themes/showcase" zoom="0.5" without-interaction> </wa-zoomable-frame>
 ```
+
+### Disabling theme sync
+
+By default, the frame automatically mirrors the host page's light/dark mode and theme selector classes (such as `wa-theme-*`, `wa-brand-*`, and `wa-palette-*`) into the iframe document. Add the `without-theme-sync` attribute to opt out of this behavior, for example when the iframe content manages its own theme independently.
+
+```html {.example}
+<wa-zoomable-frame src="/examples/themes/showcase" zoom="0.5" without-theme-sync> </wa-zoomable-frame>
+```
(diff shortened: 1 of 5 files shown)


lindsaym-fa:
This is slick. 👏 I'm not sure if `syncTheme()` has any performance impact, but I'm wondering if we should make this opt-in instead of opt-out. I'd guess that our users have fewer use cases for same-origin iframe theme syncing than we do.

claviska:
> but I'm wondering if we should make this opt-in instead of opt-out

I second this. The more common use case probably won't use this, so let's make them opt-in to the magic instead.

claviska (commented):
Nicely documented.

It looks like this will only sync the zoomable frame when the main `<html class="wa-dark">` changes, but it's possible to have light/dark mode in arbitrary containers with the `wa-light` and `wa-dark` classes:

https://codepen.io/team/webawesome/pen/MYjvvWE?editors=1000

We probably want to support that do, and I'm not sure that classes are the easiest or most reliable way to do this.

`<wa-chart>` uses a style-observer to watch for custom property changes without using a mutation observer or DOM traversal, but there's also a much lighter way to do it.

We'd need to add a "private" `--wa-color-scheme-detector` token to the each theme that gets set to, e.g. `white` in light mode and `black` in dark mode. With that token, any component can be color scheme-aware using this technique:

1. Assign the `--wa-color-scheme` as the background on a hidden element
2. Add a 1ms transition to the hidden element via CSS
3. Check the initial value with `getComputedStyle()`
4. Listen for `transitionend` on the hidden element and checking `getComputedStyle()` again when the `color` property changes

In fact, this is so useful we should probably abstract it into a [Reactive Controller](https://lit.dev/docs/composition/controllers/) so usage is dead simple:

```js
import { ColorSchemeController } from '../utilities/color-scheme.js';

class MyElement extends WebAwesomeElement {
  // ...
  private colorScheme = new ColorSchemeController(this);

  render() {
    return html`<p>Color scheme: ${this.colorScheme.value}</p>`;
  }
}
```

We can't ship this without supporting the full theming API Web Awesome supports, so we definitely need to solve this.

Let me know if you'd like my help with the Color Scheme Controller. That seems like the most reasonable way forward to me.

claviska:
Nice update with f29ac46. The color scheme controller appears to be working well!

kelseythejackson:
I needed to update some tests, and there was a typo in the changelog.

claviska (approved):
Making theme sync opt-in was the right move, especially because of the DOM traversal. Looks good!

PR #2169, opened 2026-03-16
wa-input: Fix withoutSpinButtons

wendevlin:
withoutSpinButtons in `wa-input` needs to be reflected to support property binding.


claviska (approved):
This is valid. All `with-` and `without-` attributes should reflect. Thanks for the PR!

wendevlin:
> This is valid. All `with-` and `without-` attributes should reflect. Thanks for the PR!

sure thing! You're welcome 😊

PR #2170, opened 2026-03-16
Add the experimental `<wa-markdown>` component

claviska:
Adds `<wa-markdown>` as a utility to render markdown in plain ol' HTML pages. Because of the way browsers handle certain nodes, and because arbitrary HTML can also exist inside markdown, we use `<script type="text/markdown">` to provide the source content before rendering it.

Example usage:

```html
<wa-markdown>
  <script type="text/markdown">
    ## Getting Started

    Here's a quick overview with **bold**, *italic*, and `inline code`.

    - Install the package
    - Import the component
    - Start writing markdown
  </script>
</wa-markdown>
```

### How it works

Content goes inside a `<script type="text/markdown">` element, which prevents the browser from parsing your markdown as HTML. The component reads the markdown, strips common indentation so you can indent freely, and runs the result through [Marked](https://marked.js.org/) to produce HTML. The rendered output is injected into the light DOM, so it picks up your page's existing styles (no shadow DOM boundary to fight with).

### Extending the parser

Every `<wa-markdown>` element exposes a `marked` property that points to a shared [Marked](https://marked.js.org/) instance. You can call `marked.use()` on any element to register custom renderers, extensions, or third-party plugins. The configuration applies globally to all instances.

Marked's [extension API](https://marked.js.org/using_advanced#extensions) supports custom tokenizers, renderers, and walkers. If you need a custom markdown flavor or extension, you can plug it in without forking the component or importing a separate instance.

### But why?

Is it neat? Yes. Is it practical? We're not sure. We're releasing this as an experiment to answer that, so please let us know what you think!

### Why did you use Marked instead of...

The main reason is the tradeoff of size and plugin support. Marked hit the, uh, _mark_ in terms of being small enough to reasonably componentize and easy enough to extend.

### But SEO / SSR...

Acknowledged. This is an experimental utility. It won't get you on the front page of Google because markdown doesn't include the semantics that HTML does (e.g. `## Heading` vs `<h2>Heading</h2>`) so avoid using it in sensitive areas.

We're really curious to learn _how_ folks use this and if it's a good idea or not, so please let us know in [Discord](https://discord.com/invite/mg8f26C), [Discussions](https://github.com/shoelace-style/webawesome/discussions), or [however else](https://webawesome.com/docs/resources/support) you prefer to stay in touch with the team!


KonnorRogers:
Speaking from personal experience, `<template>` is a really bad idea for this component.

Consider the following markdown:

```markdown
<https://foo.com>
```

Or, even less ambiguously,

````markdown
```html
<div>
  Foo Bar Baz
</div>
```
````

^ The HTML parser will intercept these and interpret the `<div>` as an element before the markdown can run.

The least ambiguous way to do this is with `<script>` tags, I would probably recommend:

```html
<script type="text/markdown">
</script>
```

with the caveat being if a user needs to add a `<script>` tag, we can let them do:

```html
<script>&lt;/script>
```

This is what lit-playground-elements does as well.

kelseythejackson (dismissed):

packages/webawesome/docs/assets/styles/docs.css
```diff
@@ -290,6 +290,11 @@ h1.title {
   }
 }
 
+.component-summary {
```
talbs, line 293: ✨ Oh, you fancy, huh?

talbs: But seriously, nice touch.

talbs (commented):
This is an interesting component - the Docs and DX make sense to me.

I assume that if you're reaching for this one and want to support Markdown rendering in your project, you have some technical chops - so the less-than-ideal `<script>` syntax we compromised on feels okay considering that.

I did have a few UXy thoughts...

### Feedback When Something Goes Wrong
From a UX perspective, when rendering fails, or the setup is wrong, the user gets no feedback if there's a missing/empty source or parse errors.

Can we think about a dev-time warning, try/catch around `marked:parse()`, or add a troubleshooting/common mistakes section to these docs?

---

### Guidance for User-Supplied Markdown
Marked allows raw HTML in markdown, but does it sanitize? If not, this isn't a big deal when the source is controlled. But should we provide any guidance for user-generated input?

Would it be sound to add a "Do not use <wa-markdown> with unsanitized user input" heads up in these docs?

---

### Use Case Guidance
If we add that unsanitized user input heads up in these docs from the previous thought, it could pair well in a Best Practices section in these docs with the concerns you've noted on SEO.

---

### Supported Markdown Features
Can we make it clearer what Markdown features/content are supported in this component? I know that's Marked's responsibility, but it took me a bit of hunting through their docs to see what content is supported (see https://marked.js.org/#specifications). I could see our users wanting that information as they adopt and use this component.

So maybe just a link or callout to that list would be helpful.

packages/webawesome/docs/docs/components/markdown.md
```diff
@@ -0,0 +1,221 @@
...
+    Here's a quick overview with **bold**, *italic*, and `inline code`.
+
+    - Install the package
+    - Import the component
+    - Start writing markdown
+  </script>
+</wa-markdown>
+```
+
+Note the use of `<script type="text/markdown">` instead of `<template>` — see [why we use script](#why-script-instead-of-template) below.
```
lindsaym-fa:
```suggestion
Note the use of `<script type="text/markdown">` instead of `<template>` — see [why we use script](#why-use-script-instead-of-template) below.
```

claviska:
Good call @talbs. Updated the docs and added warnings.

talbs (approved):
Thanks for the error-handling and docs polishing. 🧼 🧹

This looks good from my perspective to release as experimental. I'm curious to see how folks will use it.

claviska:
> I'm curious to see how folks will use it.

Same!

cameronelliott:
Update: Maybe I should be using 11ty (BA) for my static markdown, and <is-land> for updating islands. 
I am still learning and thinking this through, thanks for any patience as I work through what I really need.

-- original post --


Hey, this is a cool component, but for my use case (basically a networking analyzer for video),
I need a few things: 
1. I've got to have it usable with lit SSR, so it can be usable for SEO.
2. I might need to be able to nest other components inside, and have updatable regions (like regular JS inline code)

I know nested component and updatable regions sounds ugly, but I'm going to give it a shot. 

Anyway, I hope you guys keep working on this or alternatives, this and the wa-prose really look nice together!

PR #2172, opened 2026-03-16
Docs: Normalize Search Result URLs

talbs:
This work normalizes search result URLs so they show real paths (removing `.njk`/`.html`, and `index` files collapse to the parent path).

This comes most into play with views from `webawesome-app` where those extensions were not being stripped out.


talbs:
@claude, review this PR.

talbs:
@claude, re-review my PR.

claviska (approved):
✨

PR #2177, opened 2026-03-17
Fix `<wa-select multiple>` padding

lindsaym-fa:
Takes the same approach as implemented to `<wa-combobox multiple>` in https://github.com/shoelace-style/webawesome-pro/pull/164 

Adds private `--_padding-with-tags` for consistent block and inline padding when tags are present in `<wa-combobox multiple>`. The previous inconsistency wasn't noticeable in most themes, but it can be obvious in others.

Playful theme before:
Screenshot 2026-03-17 at 2 46 19 PM

Playful theme after:
Screenshot 2026-03-17 at 2 46 15 PM


claviska (approved):
📏

PR #2180, opened 2026-03-17
Dropdown Item: Only set aria-checked when type is checkbox

### LINKED ISSUE #2179: wa-dropdown-item improperly applies a aria-checked when type is not checkbox.
### Describe the bug
Invalid aria is created by adding `aria-checked` even when the type isn't `checkbox`.  This can be flagged by axe's `aria-allowed-attr` rule because `aria-checked` is only valid as an attribute on `role="menuitemcheckbox"` rather than 

### To Reproduce

1. If you go to https://webawesome.com/docs/components/dropdown
2. Open the example dropdown then inspect the first item
3. You will see `aria-checked=false` despite it not being a checkbox type, it has the `role="menuitem"`

### Additional information
This is caused because the `checked` attribute is set to default to `false` which Lit will detect as a change on the first render which causes the `updated()` body to set it.  It should be changed to be conditionally added only when the type is `checked` instead.

claydiffrient:
## Summary

- Guard `aria-checked` behind `type === 'checkbox'` so normal menu items no longer get an invalid `aria-checked` attribute
- Handle type changes by adding/removing `aria-checked` when switching to/from checkbox
- Add tests verifying `aria-checked` presence/absence for normal and checkbox types

Fixes #2179

### DIFF
--- a/packages/webawesome/src/components/dropdown-item/dropdown-item.test.ts
+++ b/packages/webawesome/src/components/dropdown-item/dropdown-item.test.ts
@@ -27,4 +27,36 @@ describe('<wa-dropdown-item>', () => {
     await clickOnElement(el);
     expect(clickHandler).to.have.been.calledOnce;
   });
+
+  it('should not have aria-checked when type is normal', async () => {
+    const el = await fixture<WaDropdownItem>(html` <wa-dropdown-item>Item</wa-dropdown-item> `);
+    await el.updateComplete;
+    expect(el.hasAttribute('aria-checked')).to.be.false;
+  });
+
+  it('should have aria-checked when type is checkbox', async () => {
+    const el = await fixture<WaDropdownItem>(html` <wa-dropdown-item type="checkbox">Item</wa-dropdown-item> `);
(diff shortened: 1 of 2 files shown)


claviska (dismissed):
Thanks for the PR. This looks great! We just need to add a changelog.

claviska:
Looks like it's failing on Prettier. Please run `npm run prettier:fix`. We can handle the changelog entry :)

claydiffrient:
> Looks like it's failing on Prettier. Please run `npm run prettier:fix`. We can handle the changelog entry :)

Sorry I missed that!  Pushed up it up now.  I didn't see guidance around changelogs in the documentation, and I didn't see anything like changesets in use in the repo.  So I appreciate you doing that part for me :)

claviska:
> I didn't see guidance around changelogs in the documentation, and I didn't see anything like changesets in use in the repo. So I appreciate you doing that part for me :)

No worries. We manage it by hand right now. Happy to tackle it.

claviska (approved):

PR #2198, opened 2026-03-24
Card: Body Part Wrapper + Fix Horizontal `::slotted` Rules

talbs:
`wa-card` used a `<slot part="body">` for the main (default) content. Other components (e.g. `wa-dialog`, `wa-drawer`) expose the `body` CSS part on a **wrapper** around the default slot instead. 

This PR aligns with that pattern and refines the horizontal layout CSS.

## Changes

1. **Markup** — Wrap the default slot in `<div part="body" class="body"><slot></slot></div>` for vertical and horizontal orientations. The public API remains unchanged: the main content still uses the default slot; `::part(body)` still targets the main content region (now the wrapper `div`), consistent with the “container” wording in the documentation.

2. **Horizontal styles**
   - **Body:** Replace `::slotted([slot='body'])` with `.body slot::slotted(*)`. There is no named `body` slot; the old rule did not match default-slot content.
   - **Actions:** Replace a root-level `::slotted([slot='actions'])` with `slot[name='actions']::slotted(*)`, which is the valid `slot::slotted(...)` form for the actions slot.

3. **Tests** — Add a test that `[part="body"]` is a `div` containing the default slot and that default-slot content is assigned as expected.


talbs:
@claude, review this PR.

claviska (dismissed):
This is the correct convention. We shouldn't be changing the `display` of slots because it adversely affects the accessibility tree. Wrapping is more bulletproof. Thanks for catching this.

PR #2204, opened 2026-03-24
Replace Lunr with MiniSearch

claviska:
This replaces Lunr with MiniSearch, a similar library but with better results. To test, press <kbd>⌘K</kbd> and search for various queries. (You can compare results to the live website to see the difference in results.)


lindsaym-fa (approved):
Most simple searches aren't showing a ton of differences for me, but I'd say that it's at least better for longer/more specific queries. 👍

PR #2208, opened 2026-03-25
Don't trim descenders in dropdown items

### LINKED ISSUE #2207: Letters such as g y j have their tails trimmed in wa-dropdown-item labels
Letters such as g y j have their tails trimmed in wa-dropdown-item labels, as shown below.

Image

Fonts used in the example are from the default library stack, default theme.
This is probably the same bug as reported in issue #2048, which remains unsolved.

I found the bug is caused by the following custom property:

`--wa-line-height-condensed: 1.2;`

If you set `--wa-line-height-condensed` value to 1.3, the problem disappears. The drawback is that it increases the height of a dropdown, what may be undesired.
The other solution, without increasing the value of `--wa-line-height-condensed`, and the one that I have decided to use, is to allow overflow of the dropdown label.

```
wa-dropdown-item::part(label) {
	overflow: visible;
}

```

The dropdown item has enough default padding, so it's not a problem to allow a tiny overflow.

Browser / OS
 - OS: Windows
 - Browser: Chrome
 - Browser version: 145.0.7632.160

claviska:
Fixes https://github.com/shoelace-style/webawesome/issues/2207

The issue was with `overflow: hidden` which was there for text truncation, but text truncation never occurs because the host is flex and items default to `min-width: auto` which prevents shrinking below content size, e.g.:

CleanShot 2026-03-25 at 09 11 25@2x
<br><br>

@lindsaym-fa I removed the truncation properties. I assume this to be safe because the label didn't wrap in the first place. Users who desire truncation or wrapping can apply custom styles if desired, but I think this cover 99% of uses cases since menu items shouldn't normally be very long anyways.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -24,6 +24,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 - Fixed a bug in `<wa-scroller>` that caused horizontal page overflow in Chrome when containing wide content such as tables
 - Fixed a bug in `<wa-details>` and native `<details>` that caused full-width elements to overflow the details content [issue:2137]
 - Fixed the `autocorrect` property type in `<wa-input>` and `<wa-combobox>` to use `boolean` instead of a string union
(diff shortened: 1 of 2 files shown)


lindsaym-fa (dismissed):
No concerns. 👍

PR #2213, opened 2026-03-25
Add Changelog Notification Sign Up

claviska:
Adds a signup form to the changelog so folks can choose to get notified when we ship new versions. This is a separate list, a no spam, no nonsense list. We will only use this to tell people about new releases and what's in them.

@talbs it's all yours 🧞‍♂️

### DIFF
--- a/packages/webawesome/docs/_includes/changelog-email-signup.njk
+++ b/packages/webawesome/docs/_includes/changelog-email-signup.njk
@@ -13,7 +13,7 @@
           <wa-icon name="envelope" variant="regular" slot="start"></wa-icon>
         </wa-input>
         <wa-button type="submit">
-          Yes, Please
+          Yes, please
           <wa-icon name="arrow-right" variant="regular" slot="end"></wa-icon>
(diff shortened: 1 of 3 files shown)


talbs:
@claviska, pushed some polish in https://github.com/shoelace-style/webawesome/pull/2213/commits/7f87d8772b44c4e9a06707f9c1194de11f449c92.

claviska:
I think blue was the wrong call after all. What if we slimmed it down a bit? This isn't something we're necessarily trying to draw attention to, but more of a service we're willing to provide for those who want it. 🤔 

What if we slimmed it down and toned it down a bit?

CleanShot 2026-03-26 at 11 35 13@2x

View in context:

CleanShot 2026-03-26 at 11 36 19@2x

talbs:
Updated in https://github.com/shoelace-style/webawesome/pull/2213/commits/dc9048b00e7001fe36f6cb95e39b64eca6d1512a.

CleanShot 2026-03-26 at 11 59 39@2x

PR #2214, opened 2026-03-25
Add custom states to <wa-button>

claviska:
Adds `disabled`, `icon-button`, `link`, and `loading` custom states to `<wa-button>`. The inspiration for this came from https://github.com/shoelace-style/webawesome/discussions/2185 because it should be simpler to style various button states. No new examples in the docs, but they're all listed in the Custom States table.


talbs (approved):
Looks good. 

The custom styling section could definitely use some love and explanation (especially noting these states). But we've captured that as its own follow-on work across the docs.

Issue #2205, opened 2026-03-24
Rating doesn't work properly with screen readers on mobile


paulGeoghegan:
### Describe the bug
Currently the rating component isn't useable using a screen reader like Talkback on mobile.

### To Reproduce

1. Go to the rating docs
2. Turn on Talkback
3. Navigate to the rating example
4. Try to interact by swiping up and down to change the rating

### Browser / OS
 - OS: Android
 - Browser: Any
 - Browser version: Any

claviska:
This will be fixed as part of #2215

claviska:
Fixed in https://github.com/shoelace-style/webawesome/pull/2215

paulGeoghegan:
@claviska this is actually not really fixed at all. Can you please re-open

paulGeoghegan:
Just to clarify you should / need tobe able to swipe up or down in order to change the rating. Currently tapping it sets it to 3 and then back to 0. This makes them useless for mobile.

claviska:
I verified that it wasn't reading in Voiceover and verified the fix in #2215. The problem was all of the attributes being rendered in a shadow root. The role and all area attributes now appear directly on the host element, preventing any shadow DOM-related issues:

```html
<wa-rating
  role="slider" 
  aria-valuenow="4" 
  aria-valuemin="0" 
  aria-valuemax="5" 
  aria-disabled="false"
  aria-readonly="false"
  aria-label="Rating"
  tabindex="0"
  size="medium"
></wa-rating>
```

I don't have access to Android or Talkback so I can't really help troubleshoot that combination. But with everything on the host element, I can't see how any screen reader wouldn't understand or announce the rating now. It's pretty simple, whereas before it was not.

That said, PRs are encouraged but this feels like one of those screen reader whack-a-mole games because without some kind of a hack I'm not sure what else we could do to make it work.

lindsaym-fa:
I'm also not able to verify with Android + Talkback, but testing with iOS Safari + VoiceOver, the rating toggles between 0 and the middle value on double-tap. Once activated, on swipe up, the rating increases by one, and on swipe down, the rating decreases by one. That all seems to be expected.

@paulGeoghegan, if that isn't the same behavior you're experiencing, I wonder if this is a Talkback limitation. ☹️ [The ARIA APG warns](https://www.w3.org/WAI/ARIA/apg/patterns/slider/) that AT may not support the right gestures and output for this slider pattern.

paulGeoghegan:
[...]

PR #2215, opened 2026-03-25
Make rating a form control

claviska:
Makes `<wa-rating>` form associated so you can submit it with `<form>`.

Related discussion: https://github.com/shoelace-style/webawesome/discussions/2192


packages/webawesome/src/components/rating/rating.ts
```diff
@@ -32,22 +34,39 @@ import styles from './rating.styles.js';
...
+  @property() name: string | null = null;
+
   /** A label that describes the rating to assistive devices. */
   @property() label = '';
 
   /** The current rating. */
   @property({ type: Number }) value = 0;
 
+  /** The default value of the form control. Used to reset the rating to its initial value. */
+  @property({ type: Number, attribute: 'default-value' }) defaultValue = 0;
```
KonnorRogers, line 130: This kinda skates "where we want to be" but does not align with the rest of the library currently. Are we okay with that?

If i do for example:

```html
<wa-rating default-value="3"></wa-rating>
```

im pretty positive it will default to no stars.

claviska: Yeah but it _will_ reset to the correct value. I'm OK with it if you are. We need to get everything updated soon anyways.

claviska:
Also fixes https://github.com/shoelace-style/webawesome/issues/2205 by moving role and aria attributes to the host element.

PR #2218, opened 2026-03-25
Add support for custom buttons in `<wa-copy-button>`

### LINKED ISSUE #1327: Make copy button capable of handling more button types
Some feedback from a user on Discord:

> Hi, when you need to show a couple of buttons and a copy button, the latter is misaligned and has a different size. Shouldn’t it have wa-button sizing and classes? For example: https://codepen.io/Stefano-Rivoir/pen/VYvMgEg

I think the original idea was that the copy button’s placement is unique and oftentimes smaller than regular buttons. [Looking at the docs](https://webawesome.com/docs/components/copy-button), copy button doesn’t support sizes.

My initial thought was we should add `size="…"` to copy button that would mimic the same sizes/paddings as `<wa-button>` to solve use cases like this and give you the option of having other variations. But then how do we support other appearances and variants?

**That leaves us two options:**

1. Continue to attempt to mirror all button features in `<wa-copy>` (meh, imo) OR
2. Adapt `<wa-copy>` to use a default slotted `<wa-button>` and let the user slot in any button they want.

In the case of (2), you’d also be able to slot in native `<button>` elements or really any interactive element you want. It creates a better separation and makes copy more flexible. (For this, the size would go on the slotted button itself, not on the copy button.)

---

It seems like (2) is the best option and already has support from @lindsaym-fa.

claviska:
This PR adds support for using custom buttons in WA's copy button. I was hoping to simplify this and just use a fallback slot content, but unfortunately to keep the API backwards compatible I needed to bring in the ` HasSlotController` to continue supporting custom icons via `copy-icon`, `success-icon`, and `error-icon`.

The good news is existing copy buttons will continue to work just fine, but you can now slot in arbitrary buttons as well :)

I've added an example to the docs to demonstrate.

Fixes #1327

### DIFF
--- a/packages/webawesome/docs/docs/components/copy-button.md
+++ b/packages/webawesome/docs/docs/components/copy-button.md
@@ -40,6 +40,28 @@ Use the `copy-icon`, `success-icon`, and `error-icon` slots to customize the ico
 </wa-copy-button>
 ```
 
+### Custom Trigger
+
+By default, the copy button renders an icon-only button. You can slot in any element to use as a custom trigger instead. This works with Web Awesome buttons, native buttons, or any clickable element.
+
+```html {.example}
+<wa-copy-button value="You can copy anything with a custom trigger!">
+  <wa-button appearance="filled">Copy to Clipboard</wa-button>
+</wa-copy-button>
+```
+
+You can also use a native button as the trigger.
+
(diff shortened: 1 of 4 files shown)


kelseythejackson (approved):
Looks good

PR #2220, opened 2026-03-25
Wrap badge slots and fix invalid role

### LINKED ISSUE #2163: wa-badge : role attribute on <slot> is not allowed
### Describe the bug

The [role attribute](https://github.com/shoelace-style/webawesome/blob/67960d7ce15201967c816a535721d3275f41fd43/packages/webawesome/src/components/badge/badge.ts#L44) is not allowed on `<slot>`.

### To Reproduce
See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/slot#technical_summary

### Additional information
I think that `role="status"` or, even better, `aria-live="polite"` directly on `<wa-badge>` would be a more sensible solution. Alternatively, you could omit the attribute entirely and have it set manually.

claviska:
Wraps slots per the WA convention and fixes #2163 by moving the `role` attribute off of the slot.


talbs (approved):
👀 good. Getting 🎁🎁🎁.

Issue #2102, opened 2026-02-27
Dropdown label's design cannot be altered by themes


roydukkey:
### Describe the bug

The documentation [suggests](https://webawesome.com/docs/components/dropdown#showing-labels-and-dividers):

> Use any heading, e.g. `h1`–`h6` to add labels and the `wa-divider` element for separators.

However, the styling doesn't allow themes to change the design of the labels because of the component's use of `!important`:

https://github.com/shoelace-style/webawesome/blob/b1c62d53bfe6d04114f5faf97414db52b13ef8a9/packages/webawesome/src/components/dropdown/dropdown.styles.ts#L35-L48

### To Reproduce

1. Go to try styling any of the properties using `!important`
1. See that the rules down take affect

### Demo

https://codepen.io/roydukkey/pen/MYjYLOe?editors=0100

### Additional information
Before finding this issue, I was questioning whether the implementation was even good for accessibility in this discussion https://github.com/shoelace-style/webawesome/discussions/2101.

kelseythejackson:
We appreciate you bringing this to our attention. We'll take a look and see if we can get this sorted out. Thanks!

kelseythejackson:
@roydukkey we dug into this, and the `::slotted()` specificity is so low that we use `!important` to ensure the correct styles render in those specific locations.

However, we can recommend some workarounds. For your theming, you can override the tokens:
``` 
wa-dropdown {
   --wa-color-text-quiet: deeppink;
   --wa-font-family-body: 'Inter', sans-serif;
   --wa-font-weight-semibold: 500;
   --wa-font-size-smaller: 0.7rem;
 }
```
Let us know if this works for you. Thanks!

roydukkey:
Humm... ic. Though, the properties I am actually after are `margin` and `padding`.

kelseythejackson:
Gotcha, that would mean switching out those hard coded values with custom properties, which would allow you to put in what you want. That's more of a long term fix, as it may affect other components. We'll make a note of it and keep you up to date. Thanks.

claviska:
This was caused by the `!important` rule which made it impossible to style from userland. Fixed in #2226

PR #2226, opened 2026-03-26
Make dropdown headings not use !important styles

claviska:
Fixes the issue described in #2102.

The `::slotted()` selector was setting color with `!important` from inside the shadow DOM, making it impossible to override the color in userland. (There's an example in the user's CodePen.)


claviska:
@lindsaym-fa This is a pretty trivial fix and I don't see any visual differences, so I'm going to save you the time and merge this one.

Issue #1817, opened 2025-11-22
wa-dropdown-item fires click event when disabled


bennyschudel:
The `wa-dropdown-item` fires the click event when it is disabled. Is this on purpose or a bug?

If it is a bug, I will create a more detailed bug report if needed.

Thank you so much!

kelseythejackson:
Hey @bennyschudel, thanks for raising this issue. This is a bug, and we'll take a look at it.

claviska:
Fixed in #2232

PR #2232, opened 2026-03-26
Fix dropdown disabled item click

claviska:
This PR fixes disabled drop-down items, which were erroneously dispatching with `wa-select`.

See #1817


kelseythejackson (dismissed):
The logic makes sense. Approved

PR #2235, opened 2026-03-26
Fix `--track-width` not being applied correctly

### LINKED ISSUE #1317: Spinner's track-width css property is not being applied to track and indicator width

claviska:
This PR makes the spinners track width true to the actual value provided.

@lindsaym-fa I went with a 2px default because that seemed pretty close to the existing design.

Fixes #1317 and an alternative to https://github.com/shoelace-style/webawesome/pull/1316.


lindsaym-fa (approved):
Perfection. ✨

PR #2237, opened 2026-03-27
Add SSR guards

### LINKED ISSUE #1858: Usability of elements in SSR (wa-page, wa-button, wa-drawer)
### Describe the bug

I tried to use `wa-page` in SSR'd page but so far I failed. I'm creating this issue to round up my issues I've been trying to patch locally and the results

I'm basing this on `v3.0.0-beta.4`

- `wa-page` 
   - [x] #1862
   - [ ] calls `Element#querySelector` in multiple places from `connectedCallback`
   - [ ] calls `addEventListener` from connectedCallback
- `wa-button` and `wa-drawer`
   - [x] #1861
   - [x] #1860

In case of `wa-page`, I was able to wrap the problematic code with `isServer` and I think it should pretty much "just work".

For the `HasSlotController`, I expect that using it conditionally is not enough, because that will likely cause hydration mismatch. At least in some cases.

Unfortunately, I bump into one more problem which prevents my app from running. As soon as I add the necessary `if (!isServer)` conditionals to avoid client-only APIs, the server rendering fails where I provide a context using `@lit-labs/context`. I need to dig more into this.

### LINKED ISSUE #2174: SSR(wa-scroller): ResizeObserver is not defined
### Describe the bug

When using `wa-scroller` with `@lit-labs/ssr`, it throws an error because `ResizeObserver` is instantiated in its constructor.

### Additional information

[This](https://github.com/TanStack/virtual/issues/419#issuecomment-1309915611) helps but ultimately, the resize observer would better be called in `connectedCallback`

```js
import ResizeObserver from 'resize-observer-polyfill'

global.ResizeObserver = ResizeObserver
```

### LINKED ISSUE #2178: SSR(wa-card): HasSlotController causes hydration mismatch
### Describe the bug

Using WA Pro 3.3.1

I think there a problem in `HasSlotController` which causes a mismatch between client and server rendering of `wa-card` and likely other elements.

https://github.com/shoelace-style/webawesome/blob/fe2fb923f5dcee4cfafc3f8dc540a0c4dcfda491/packages/webawesome/src/internal/slot.ts#L44

This line behaves differently on the server. `this.host.querySelector` doesn't exist so the result will be `undefined`. Thus, comparing to `null` returns a false positive result. In `wa-card` this causes this condition to render differently on the server

https://github.com/shoelace-style/webawesome/blob/fe2fb923f5dcee4cfafc3f8dc540a0c4dcfda491/packages/webawesome/src/components/card/card.ts#L81-L88

### Possible solution

```diff
-    return this.host.querySelector?.(`:scope > [slot="${name}"]`) !== null;
+    return !!this.host.querySelector?.(`:scope > [slot="${name}"]`);
```

### Additional information

This is a regression since `@awesome.me/webawesom@3.0.0-beta.4` where `this.hasSlotController` was not used in `WaCard#render`

claviska:
This PR adds guards to improve the SSR experience. It also adds guidance in `contributing.md` for developing with SSR in mind. I will work on a subsequent PR for the `HasSlotController` since those changes might be more controversial.

This PR also includes a small performance improvement to `<wa-textarea>` by loading the resize observer lazily only when necessary.

Fixes #1858
Fixes #2174
Fixes #2178

Companion PR for pro is at https://github.com/shoelace-style/webawesome-pro/pull/175

@KonnorRogers One thing this PR does not do is re-enable SSR tests.

### DIFF
--- a/packages/webawesome/src/components/carousel/carousel.ts
+++ b/packages/webawesome/src/components/carousel/carousel.ts
@@ -103,8 +103,12 @@ export default class WaCarousel extends WebAwesomeElement {
 
   connectedCallback(): void {
     super.connectedCallback();
-    this.setAttribute('role', 'region');
-    this.setAttribute('aria-label', this.localize.term('carousel'));
+
+    // SSR guard: setAttribute is not available during server-side rendering
+    if (!isServer) {
+      this.setAttribute('role', 'region');
+      this.setAttribute('aria-label', this.localize.term('carousel'));
+    }
   }
 
   disconnectedCallback(): void {
--- a/packages/webawesome/src/components/tree/tree.ts
+++ b/packages/webawesome/src/components/tree/tree.ts
@@ -106,13 +106,16 @@ export default class WaTree extends WebAwesomeElement {
   async connectedCallback() {
     super.connectedCallback();
 
-    this.setAttribute('role', 'tree');
-    this.setAttribute('tabindex', '0');
+    // SSR guard: MutationObserver is not available during server-side rendering
+    if (!isServer) {
+      this.setAttribute('role', 'tree');
+      this.setAttribute('tabindex', '0');
 
-    await this.updateComplete;
+      await this.updateComplete;
 
-    this.mutationObserver = new MutationObserver(this.handleTreeChanged);
-    this.mutationObserver.observe(this, { childList: true, subtree: true });
+      this.mutationObserver = new MutationObserver(this.handleTreeChanged);
+      this.mutationObserver.observe(this, { childList: true, subtree: true });
+    }
   }
 
   disconnectedCallback() {
--- a/packages/webawesome/src/internal/webawesome-element.ts
+++ b/packages/webawesome/src/internal/webawesome-element.ts
@@ -56,12 +56,16 @@ export default class WebAwesomeElement extends LitElement {
 
   connectedCallback() {
     super.connectedCallback();
-    // Helpful comment node inside the shadow root that links to the docs
-    this.shadowRoot?.prepend(
-      document.createComment(
-        ` Web Awesome: https://webawesome.com/docs/components/${this.localName.replace('wa-', '')} `,
-      ),
-    );
+
+    // SSR guard: document is not available during server-side rendering
+    if (!isServer) {
+      // Helpful comment node inside the shadow root that links to the docs
+      this.shadowRoot?.prepend(
+        document.createComment(
+          ` Web Awesome: https://webawesome.com/docs/components/${this.localName.replace('wa-', '')} `,
+        ),
+      );
+    }
   }
 
   attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null) {
--- a/packages/webawesome/docs/docs/experimental/ssr.md
+++ b/packages/webawesome/docs/docs/experimental/ssr.md
@@ -110,6 +110,38 @@ function fixDeclarativeShadowDOM(e) {
 });
 ```
 
+### The `with-*` Attributes
+
+Some components use slot detection to conditionally render parts of their template. For example, `<wa-dialog>` only renders its footer when a `footer` slot is present. During SSR, slot detection doesn't work because the DOM isn't available, so these parts would be missing from the initial server-rendered markup.
+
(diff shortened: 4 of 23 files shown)


packages/webawesome/src/components/carousel/carousel.ts
```diff
@@ -103,8 +103,12 @@ export default class WaCarousel extends WebAwesomeElement {
 
   connectedCallback(): void {
     super.connectedCallback();
-    this.setAttribute('role', 'region');
-    this.setAttribute('aria-label', this.localize.term('carousel'));
+
+    // SSR guard: setAttribute is not available during server-side rendering
```
tpluscode, line 107: In fact, `@lit-labe/ssr` does support attribute operations on elements. At least in version 3.3.1

claviska: Thanks, that's good to know! I don't think that matters for this particular lock of code since it will get added at runtime, but do let me know if I missed anything.

tpluscode: I was in fact worried that adding conditional around `setAttribute` would cause a hydration mismatch but luckily it's not the case.

Besides, `connectedCallback` is not called by default but there is a hidden option to switch which hopefully will become a feature. I personally rely on `connectedCallback` in server-side code so I consider this future-proofing

packages/webawesome/src/components/carousel/carousel.ts
```diff
@@ -103,8 +103,12 @@ export default class WaCarousel extends WebAwesomeElement {
...
+    if (!isServer) {
+      this.setAttribute('role', 'region');
+      this.setAttribute('aria-label', this.localize.term('carousel'));
```
KonnorRogers, line 110: setAttribute doesn't need a server guard:

https://lit.dev/docs/ssr/dom-emulation/

(we can also use `this.internals.role = "region"` and `this.internals.ariaLabel = this.localize.term("carousel")` )

Though, i dont remember if the localize controller is SSR compatible

Image

packages/webawesome/src/components/tree/tree.ts
```diff
@@ -106,13 +106,16 @@ export default class WaTree extends WebAwesomeElement {
   async connectedCallback() {
     super.connectedCallback();
 
-    this.setAttribute('role', 'tree');
-    this.setAttribute('tabindex', '0');
+    // SSR guard: MutationObserver is not available during server-side rendering
+    if (!isServer) {
+      this.setAttribute('role', 'tree');
+      this.setAttribute('tabindex', '0');
```
KonnorRogers, line 112: Doesn't need an `isServer` guard

packages/webawesome/src/internal/webawesome-element.ts
```diff
@@ -56,12 +56,16 @@ export default class WebAwesomeElement extends LitElement {
...
-    );
+
+    // SSR guard: document is not available during server-side rendering
+    if (!isServer) {
+      // Helpful comment node inside the shadow root that links to the docs
+      this.shadowRoot?.prepend(
+        document.createComment(
+          ` Web Awesome: https://webawesome.com/docs/components/${this.localName.replace('wa-', '')} `,
+        ),
+      );
```
KonnorRogers, line 67: is this going to cause hydration mismatches?

KonnorRogers (commented):
looks good, few nits, otherwise, 👍

packages/webawesome/src/internal/webawesome-element.ts
claviska: It might. What do you suggest?

KonnorRogers: ideally, its part of the render function, but it means every component needs to call `super.render()` 🤔 perhaps this could be a Controller?

packages/webawesome/src/components/carousel/carousel.ts
claviska: Removed

packages/webawesome/src/internal/webawesome-element.ts
claviska: As discussed, let's see if this actually causes a problem and address it in a future PR. I wouldn't say this work solves SSR completely as we still have to re-enable and update all the tests.

PR #2244, opened 2026-03-31
Move `<wa-page>` from pro to core

kelseythejackson:
This PR moves the `<wa-page>` component over to Web Awesome core (free).

Pro PR: https://github.com/shoelace-style/webawesome-pro/pull/178


packages/webawesome/docs/_includes/sidebar.njk
```diff
@@ -111,15 +111,7 @@
...
-    <li>
-      <span class="wa-split">
-        <a class="wa-cluster wa-gap-xs" href="/docs/components/page/">
-          Page
-          <wa-icon name="flask" aria-hidden="true" class="icon-shrink de-emphasize"></wa-icon>
-        </a>
-        {{ proBadge() }}
-      </span>
-    </li>
+    <li><a class="wa-cluster wa-gap-xs" href="/docs/components/page/">Page</a></li>
```
claviska: I'm not sure why page is at the very top but it should be in alphabetical order. (This error isn't from your PR but we might as well get it fixed while we're here.)

kelseythejackson: Updated

lindsaym-fa (approved):

PR #2265, opened 2026-04-03
Remove slot hack from wa-page

KonnorRogers:
Remove ugly slot hack from `<wa-page>`

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -63,6 +63,7 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 
 - Removed `font-variant-numeric: tabular-nums;` from default `<table>` styles in Native Styles in lieu of an opt-in `wa-tabular-nums` class [pr:2613]
 - Updated `@shoelace-style/localize` to 3.2.3 to fix a bug that caused a `RangeException` to be thrown when using Google Chrome's "Detect Language" translation feature [issue:2479]
+- Updated `<wa-page>` to conditionally render its navigation slots per view instead of swapping in placeholder slot names so the `navigation`, `navigation-header`, and `navigation-footer` slots only ever exist in the active container [pr:2265]
 
 :::
 
--- a/packages/webawesome/src/components/page/page.test.ts
+++ b/packages/webawesome/src/components/page/page.test.ts
@@ -1,4 +1,4 @@
-import { expect } from '@open-wc/testing';
+import { expect, waitUntil } from '@open-wc/testing';
 import { html } from 'lit';
 import { fixtures } from '../../internal/test/fixture.js';
 import type WaPage from './page.js';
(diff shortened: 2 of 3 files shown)


claviska:
The original version has a [`slotchange` listener on the root](https://github.com/shoelace-style/webawesome/blob/c2fdbb079e263beaaf14cc72081bddcf6597f3d0/packages/webawesome/src/components/page/page.ts#L314) but the new version removes the slot in favor of an empty `<div>`.

The resize observers appear to be setup only once on init, so I believe they'll get destroyed when the slots/divs get swapped out. There may be other size/slot-related side effects from this change.

You could consider rendering slots no matter what and applying `hidden` when they're not supposed to be visible, but you may still need to draw an extra `<div>` if the grid styles require a visible element.

KonnorRogers:
none of the resizeObservers use the navigation slot elements, and as far as i can see, nothing attaches directly to these slots, so they shouldn't be affected. I don't think any of the other resize logic relies on these slots either, most of it sits "higher". I'll give it another run through and test it though.

claviska:
@KonnorRogers I rebased and made some updates to tighten this up. Here's the test case I used, in case you find it helpful:

```html
<!doctype html>
<html lang="en" class="wa-cloak">
  <head>
    <meta charset="UTF-8" />
    <title>TEMP: wa-page navigation slot test</title>
    <link rel="stylesheet" href="/dist/styles/themes/default.css" />
    <link rel="stylesheet" href="/dist/styles/webawesome.css" />
    <script type="module" src="/dist/webawesome.loader.js"></script>
    <style>
      .test-box {
        padding: var(--wa-space-s);
        border-radius: var(--wa-border-radius-m);
        font-weight: var(--wa-font-weight-bold);
      }
      .test-box.nav-header {
        background: var(--wa-color-blue-90);
        border: 2px dashed var(--wa-color-blue-50);
      }
      .test-box.nav-content {
        background: var(--wa-color-green-90);
        border: 2px dashed var(--wa-color-green-50);
      }
      .test-box.nav-footer {
        background: var(--wa-color-red-90);
        border: 2px dashed var(--wa-color-red-50);
      }
    </style>
  </head>
  <body>
    <wa-page>
      <header slot="header" class="wa-split">
        <span class="wa-heading-m">Navigation slot test</span>
        <span>
          view:
          <strong class="wa-desktop-only">desktop</strong>
          <strong class="wa-mobile-only">mobile</strong>
        </span>
      </header>

      <div slot="navigation-header" class="test-box nav-header">NAV HEADER (blue)</div>
      <nav slot="navigation" class="test-box nav-content">
        NAV CONTENT (green)
        <a href="#one">Link one</a>
        <a href="#two">Link two</a>
        <a href="#three">Link three</a>
      </nav>
      <div slot="navigation-footer" class="test-box nav-footer">NAV FOOTER (red)</div>

      <main id="main-content">
        <h1>Main content</h1>
        <p>
          The three colored boxes must appear in the left sidebar on desktop, and only inside the drawer on mobile. They
          should never disappear, and never overlap this text.
        </p>
      </main>

      <footer slot="footer">Page footer</footer>
    </wa-page>
  </body>
</html>
```

KonnorRogers:
Poking around, this looks solid to me. 👍

claviska (approved):

claviska:
@KonnorRogers approved and ready to go if you want to merge this.

Issue #2254, opened 2026-04-02
Inconsistency with wa-switch and wa-checkbox


zb140:
### Describe the bug
This is somewhere between a bug and a feature request, I think. The issue I've run into is that `<wa-switch>` and `<wa-checkbox>` don't seem to handle their `value` property the same way.

The easiest way to see this is probably to take a look at the Codepen link. I've created a `<wa-switch>`, a `<wa-checkbox>`, and a native `<input type="checkbox">`, all without default values. Then I attached onChange handlers for all 3 to log the `value` property to the console whenever they're clicked.  Clicking any of the three shows that the value is set to `"on"`. Continuing to toggle either the `<wa-switch>` or the native checkbox leaves the `value` as `"on"` while the `checked` property changes. Continuing to toggle the `<wa-checkbox>` toggles the value between `"on"` and `null`.

It seems to me like `<wa-switch>` and `<wa-checkbox>` should do the same thing in this situation. It probably makes the most sense for it to align with the native `<input type="checkbox">` although ironically that's the opposite of what would actually be most convenient for me :smile:  I discovered this issue while trying to use a `<wa-switch>` with React Hook Form (I'm using React 19 with no wrappers). What I discovered is that I can't use `<wa-switch>` the same way I would use a checkbox:

```
// doesn't work; form value never changes
<wa-switch {...register('boolField')}>...</wa-switch>

// "works" because the `value` property gets updated on every click
<wa-checkbox {...register('boolField')}>...</wa-checkbox>

// works because RHF has special handling to read the "checked" property when type === "checkbox"
<input type="checkbox" {...register('boolField')} />
```

I can of course get around this by wrapping the switch like any other custom component, but I was super confused for a while about why the `<wa-checkbox>` worked and the `<wa-switch>` didn't, so it seemed worth bringing up.

### To Reproduce

1. Open the codepen
2. Play with the checkboxes
3. Watch the console logs

### Demo

Codepen: https://codepen.io/zb140/pen/ByLxEKm

### Screenshots
If applicable, add screenshots to help explain the bug.

### Browser / OS
 - OS: [e.g. Mac, Windows]
 - Browser: [e.g. Chrome, Firefox, Safari]
 - Browser version: [e.g. 22]

### Additional information
Provide any additional information about the bug here.

claviska:
Thanks for the report. Fixed in #2269

PR #2269, opened 2026-04-07
Fix checkbox value when unchecked

claviska:
Fixes the issue with unchecked checkbox's returning `null` instead of the assigned value described in #2254.


KonnorRogers:
good catch 👍

KonnorRogers (approved):

PR #2270, opened 2026-04-07
[docs] explain how to opt out of native styles

### LINKED ISSUE #1751: Document how to selectively opt-out of native styles
We're missing an official recommendation on how to selectively opt-out of native styles, either on a per-element or per-page basis. Definitely worth investigating a preferred approach + documenting.

See also:

- https://github.com/shoelace-style/webawesome/issues/1151
- https://discord.com/channels/739437527907303535/1437422531819737280
- https://discord.com/channels/739437527907303535/1432928899846897664

terminalchai:
## Summary
- document how to opt out of native styles for a single element with `all: revert`
- add guidance for resetting a scoped section when only part of a page should use browser defaults
- clarify that page-level entry points can simply omit `native.css` while still using themes, components, and utilities

Closes #1751

## Validation
- `npx prettier --check packages/webawesome/docs/docs/utilities/native.md`

Note: I did not include the package `build` output here because a clean checkout currently fails in the existing React wrapper generation/type-check path with absolute import errors unrelated to this docs-only change.

### DIFF
--- a/packages/webawesome/docs/docs/utilities/native.md
+++ b/packages/webawesome/docs/docs/utilities/native.md
@@ -59,6 +59,37 @@ Or, if you only want styles for native elements, include a theme and native styl
 
 You can additionally include any pre-made [theme](/docs/themes/) or [color palette](/docs/color-palettes/) to change the look of native elements.
 
+## Opting out of native styles
+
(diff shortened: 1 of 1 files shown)


lindsaym-fa (approved):
Thanks a ton for this, @terminalchai! Approved. ✨

PR #2281, opened 2026-04-13
fix(dropdown-item): correct cursor for disabled items in submenus

### LINKED ISSUE #2276: Pointer cursor is displayed over a disabled <wa-dropdown-item> in a submenu.
When a `<wa-dropdown-item>` is disabled, the default cursor is displayed in the hover state. But for the same disabled `<wa-dropdown-item>` in a submenu, the pointer cursor is displayed instead of the default cursor.

Example

```
<link rel="stylesheet" href="https://ka-p.webawesome.com/kit/b9bfcf2dca544e85/webawesome@3.5.0/styles/webawesome.css">
<script type="module" src="https://ka-p.webawesome.com/kit/b9bfcf2dca544e85/webawesome@3.5.0/webawesome.loader.js"></script>

<wa-dropdown>
  <wa-button appearance="filled" slot="trigger" with-caret>Payment method</wa-button>

  <wa-dropdown-item value="cash">Cash</wa-dropdown-item>
  <wa-dropdown-item value="check" disabled>Personal check</wa-dropdown-item>
  <wa-dropdown-item value="credit">Credit card</wa-dropdown-item>
  <wa-dropdown-item value="gift-card">Gift card</wa-dropdown-item>
  <wa-dropdown-item>
    Other
    <wa-dropdown-item slot="submenu" disabled>Barter</wa-dropdown-item>
  </wa-dropdown-item>
  </wa-dropdown>
```

Default cursor over a disabled, regular `<wa-dropdown-item>`

Image

Pointer cursor over a disabled `<wa-dropdown-item>` in a submenu

Image

Additional information:
I was able to change the cursor over a normal, not disabled,  `<wa-dropdown-item>` in a submenu on my own webpage, but it was not the case when the item was disabled. No selector, no inline styling, !important, nothing worked. Computed styles in Dev Tools were showing, for example: cursor: not-allowed, but the displayed cursor was a pointer. I tried it on Edge, but got the same results.

 - OS: Windows 11
 - Browser: Chrome
 - Browser version: 145.0.7632.160

terminalchai:
Closes #2276

## Problem

When a disabled wa-dropdown-item appears inside a submenu, the cursor shows as pointer instead of 
ot-allowed. This happens because setting pointer-events: none (both via CSS and the inline style set in updated()) causes the mouse to 'fall through' to the parent popup element, which has cursor: pointer.

## Root Cause

Two places were applying pointer-events: none to disabled items:

1. **CSS** — pointer-events: none in the :host(:state(disabled)), :host([disabled]) rule
2. **JS** — 	his.style.pointerEvents = this.disabled ? 'none' : '' in the updated() lifecycle callback

When pointer-events: none is set, the cursor displayed is determined by the element underneath, not the element itself. In a normal dropdown context that underlying element happens to have cursor: default, but inside a submenu popup it has cursor: pointer, causing the inconsistency.

## Fix

Remove pointer-events: none from both the CSS rule and the updated() inline style assignment. Click-event blocking for disabled items is already handled by the handleHostClick and handleClick event listeners, so no functionally is lost.

The cursor: not-allowed CSS declaration now applies correctly because the host element itself receives pointer events and the cursor style is respected.

### DIFF
--- a/packages/webawesome/src/components/dropdown-item/dropdown-item.styles.ts
+++ b/packages/webawesome/src/components/dropdown-item/dropdown-item.styles.ts
@@ -32,7 +32,6 @@ export default css`
   :host([disabled]) {
     opacity: 0.5;
     cursor: not-allowed;
-    pointer-events: none;
   }
 
   /* Danger variant */
--- a/packages/webawesome/src/components/dropdown-item/dropdown-item.ts
+++ b/packages/webawesome/src/components/dropdown-item/dropdown-item.ts
@@ -123,7 +123,6 @@ export default class WaDropdownItem extends WebAwesomeElement {
     if (changedProperties.has('disabled')) {
       this.setAttribute('aria-disabled', this.disabled ? 'true' : 'false');
       this.customStates.set('disabled', this.disabled);
-      this.style.pointerEvents = this.disabled ? 'none' : '';
     }
 
     if (changedProperties.has('type')) {


claviska (approved):
Seems solid. Just needs a changelog. (We'll take care of this!)

PR #2290, opened 2026-04-15
Docs: Use Free Icons in Examples (When Possible)

talbs:
Based on some good feedback in https://github.com/shoelace-style/webawesome/discussions/2288#discussioncomment-16572341, this work swaps Pro-only Font Awesome icons for Free alternatives in our docs so examples work out of the box for new users without a kit code. 

Pro icons remain in sections that specifically demo Pro features (Duotone, Sharp, Pro+ Packs, etc.).

### DIFF
--- a/packages/webawesome/docs/docs/components/badge.md
+++ b/packages/webawesome/docs/docs/components/badge.md
@@ -133,16 +133,16 @@ Use the `start` and `end` slots to add presentational elements like `<wa-icon>` 
 
 ```html {.example}
 <wa-badge>
-  <wa-icon slot="start" name="acorn"></wa-icon>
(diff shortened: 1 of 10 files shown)


talbs:
@claude, review this PR.

claviska:
It might be a good idea to add a brief section about fre/pro icon usage to: https://webawesome.com/docs/resources/contributing

I think you've narrowed it down pretty clearly:

- Default to FA free only (so copy and paste works)
- OK to use Pro for pro examples and such

(Also I'm just noticing that contributing page doesn't have an outline which is odd.)

claviska (dismissed):

PR #2292, opened 2026-04-15
Add warning for FA Pro users

claviska:
This warning is designed to detect when users try to use Font Awesome Pro icons with `<wa-icon>` without setting a kit code and help them get things working. It relies on some less than perfect heuristics, though.

This isn't using a manifest of pro icons so it currently relies on the request domain (`ka-f.fontawesome.com`) to return a 403 which can mean one of two things:
  - It's a Pro icon (expected)
  - It's an invalid icon name, e.g.: https://ka-f.fontawesome.com/releases/v7.2.0/svgs/regular/blah.svg (403 response)

The warning only fires once per page load, even if many icons are affected, to avoid flooding the console. It's scoped to the default icon library so custom libraries and `src` URLs won't trigger it. Users who provide a kit code hit `ka-p.fontawesome.com` instead, and users who self-host don't hit the FA CDN at all, so both cases are naturally excluded.

The warning is intentionally delayed by half a second so it appears below 403 errors in the console instead of getting buried when multiple icons are requested on page load.

I worry mostly about invalid icon names being a problem, so it's probably necessary to get an icon manifest. BUT that would be a lot of bytes, even if we only included free icon names.

Ideally, the CDN would reply with a 404 instead of a 403 but this check probably happens at a higher layer, e.g. during token validation.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -14,6 +14,7 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 
 <small>TBD</small>
 
+- Added a console warning in `<wa-icon>` when a Font Awesome Pro icon is requested via the free CDN without a kit code
 - Fixed a bug in `<wa-checkbox>` where the `value` property returned `null` instead of `'on'` when unchecked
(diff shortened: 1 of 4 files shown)


claviska:
Closing due to the limitation mentioned.

PR #2295, opened 2026-04-16
fix icon button width when caret is applied

claviska:
Fixes a problem where the width of the icon button didn't expand when the caret attribute was present. The bug was only visually present in Firefox.

Before:

image

After:

CleanShot 2026-04-16 at 08 21 15@2x

Mentioned in: https://discord.com/channels/739437527907303535/1491772882655641671


lindsaym-fa (approved):
Looks good!

PR #2298, opened 2026-04-17
Abstract Background-Based Styling in `utils.css`

talbs:
This work improves some of our background-based utilities in `utils.css`.

  - Adds a wa-mask-fade-below utility class that applies a bottom-edge fade mask, working on both standalone elements and .background-wa-pattern      
  pseudo-elements
  - Adds --background-pattern-size and --background-pattern-position custom properties to .background-wa-pattern for controlling the pattern's size   
  and position                                                                                                                                        
  - Uses relative color syntax (@supports) for cleaner opacity transitions when the browser supports it

### DIFF
--- a/packages/webawesome/docs/assets/styles/utils.css
+++ b/packages/webawesome/docs/assets/styles/utils.css
@@ -795,33 +795,13 @@
     background-size: var(--dot-spacing) var(--dot-spacing);
   }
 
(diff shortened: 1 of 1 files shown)


talbs:
@claude review this PR.

claviska (dismissed):

PR #2249, opened 2026-04-01
Rewrite theming documentation

lindsaym-fa:
A comprehensive rewrite and expansion of the theming documentation in the webawesome package.

**New documentation:**

- `theming-overview.md` — A new top-level theming landing page that introduces the layered theming system (theme → color palette → variants → color scheme) and links out to the individual sections. Includes new guidance on:
  - Themes
  - Color palettes
  - Color variants
  - Light and dark mode
  -  Using themes with dynamically generated instructions
- `customizing.md` — Significantly expanded with detailed guidance on creating custom themes. Includes new guidance on:
  - Light and dark mode
  - Inverting color schemes
  - Automatically detecting preferred color schemes
  - Customizing themes with CSS
  - Using multiple themes

**Refactored documentation:**

- `themes.njk` — Major restructure; extracts reusable interactive components.
- `color-palettes.njk` — Refactored similarly with shared components.
- `tokens/color.md` — Revised and condensed to reuse the new color palette viewer.

**New includes:**

- `color-palette-picker.njk`, `color-palette-viewer.njk` — Reusable interactive palette browser, now shared between the color palettes page and the color tokens docs.
- `theme-picker.njk`, `theme-viewer.njk` — Extracted from `themes.njk` for reuse.
- `instructions.njk` — Shared theming usage instructions.
- `light-dark-example.njk` — Live light/dark mode example component.

**Infrastructure:**

- `docs/_transformers/dynamic-snippets.js` — A new 11ty transformer for dynamic code snippets.
- Theme screenshot images for all built-in themes (light + dark variants) for 12 named themes.

**Minor fixes:** Small updates to `button.md`, `zoomable-frame.md`, `usage.md`, and `utilities/native.md`.

PR #2321, opened 2026-04-20
Rewrite design tokens documentation

lindsaym-fa:
A sweeping reorg and rewrite of `/docs/tokens/*`. This PR:

- Moves "Design Tokens" into the "Theming" category of the sidebar
- Removes individual token category pages in favor of sticky tab navigation on `/docs/tokens` — this declutters the sidebar and makes it easier to explore all tokens with less context switching
- Adds descriptions for each token
- Removes "Default Value" from each token table — these were cumbersome (we use a lot of CSS functions to compute tokens values) and often inaccurate, especially for folks using custom themes
- Adds documentation for missing "core color" tokens

---

@claviska, I'd love to have your input! Specifically, I'm wondering if the tab navigation feels like an improvement and if removing the "Default Value" for each token was the right call.

Additional theming documentation improvements in progress in https://github.com/shoelace-style/webawesome/pull/2249


lindsaym-fa:
@claude, review this PR.

claviska (commented):
I like this approach especially because it's sticky, but selecting a new tab after scrolling doesn't bring you to the top of the selected section which is a little confusing.

Image

Some of the pages are rather long, which is OK because the content is great. I think it's OK to keep the pages separate as long as they're presented well in the sidebar — the sidebar remains a much bigger problem IMO and I may take a stab at cleaning it up today.

One thing that came to mind is that we can no longer deep link to specific sections of the docs because of the tabs. Another thing is that they won't show up as separate pages in search. Refreshing the page resets the tab too. 🤔

It looks like these pages were moved to include files, presumably because the main page was getting unwieldy. Not to go in circles, but splitting them out is probably the right move IMO.

---

I'm not sure we should bury this info. It's important enough to be on top, so maybe just make the advanced example expandable if verbosity is a concern.

Image

---

Can we move the token names above or below the palettes in mobile so they don't wrap?

Image

---

A max width of about 450px on the form would feel better on large screens:

Image

---

We list a bunch of tokens and then show examples under a few of the tables. They're designed to show people where these tokens are applied to, but they seem rather disconnected.

Image

Compare this to some of the other examples, which are more concrete:

Image

Image

I wonder if we should spend some time creating more focused examples for certain tokens.

lindsaym-fa:
Thanks, @claviska! I reverted the tabs, exposed the Using/Customizing guidance, and stacked the color tokens above the swatches instead of to the left. I'll add some better previews for component groups in a follow-up PR.

PR #2330, opened 2026-04-24
Fix switch RTL direction

wendevlin:
Android and iOS are mirroring switches in RTL, so I think webawesome should do the same.


claviska (approved):
Verified. Thanks for the PR. I'll add a changelog entry in a follow up.

PR #2339, opened 2026-04-28
Utilities: Further Abstracting Background Styling

talbs:
This PR continues the background utility abstraction work from https://github.com/shoelace-style/webawesome/pull/2298. It does the following...

* extracts the shared `position: relative` / `z-index` lifting /` ::after` base styles into a single grouped selector
* moves `.background-grid` and `.background-dot-grid` to render on `::after` (matching `.background-wa-pattern`)
* namespaces the CSS custom properties
* fixes the `.wa-mask-fade-below` selector to handle all three background utilities correctly.

### DIFF
--- a/packages/webawesome/docs/503.md
+++ b/packages/webawesome/docs/503.md
@@ -34,8 +34,8 @@ unlisted: true
   }
 
   wa-page.background-grid {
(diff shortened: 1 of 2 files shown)


talbs:
@claude review this PR.

lindsaym-fa (approved):
No objections to these abstractions!

PR #2343, opened 2026-04-29
Update copy button feedback and make stable

claviska:
This PR removes the tooltip dependency from copy button and replaces the tooltip with visually similar feedback that appears on copy/error. This works with default triggers and the new custom triggers feature recently added (the tooltip version didn't and couldn't work).

The tooltip styles here use the tooltip design tokens but aren't shared with the actual tooltip component styles. The animations are unique.

Lastly, this PR updates copy button from experimental to stable.


packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -32,7 +32,11 @@ Components with the <wa-badge variant="warning">Experimental</wa-badge> badge sh
 - Fixed a bug where `<wa-dropdown-item>` was erroneously listed as experimental
 - Fixed `<pre>` border radius in Native Styles so it correctly uses the `--wa-panel-border-radius` token like similar elements
 - Refactored component tests across core and pro packages to follow a consistent structure with improved coverage
-- Upgraded the following components from _experimental_ to _stable_: `<wa-bar-chart>`, `<wa-bubble-chart>`, `<wa-chart>`, `<wa-combobox>`, `<wa-doughnut-chart>`, `<wa-file-input>`, `<wa-line-chart>`, `<wa-number-input>`, `<wa-pie-chart>`, `<wa-polar-area-chart>`, `<wa-radar-chart>`, `<wa-scatter-chart>`, `<wa-sparkline>`, `<wa-toast>`, and `<wa-toast-item>`
+- Replaced the tooltip in the experimental `<wa-copy-button>` with a built-in feedback popup that works with default and custom triggers
+  - Added the `feedback-placement` attribute to replace `tooltip-placement`
+  - Added the `feedback` CSS part
+  - Removed the `tooltip-placement` attribute and the related `tooltip__*` parts
```
KonnorRogers, line 50: Should probably have a big ole "BREAKING CHANGE" somewhere in here 😅

KonnorRogers (commented):
I'm curious why we swapped from tooltip to popover.

I'm _assuming_ tooltip didn't play nicely with accessibility and crossing shadow roots for things like slotted elements, but didn't see a reasoning mentioned anywhere.

claviska:
> I'm curious why we swapped from tooltip to popover.

It's not a tool sip or a popover, just styled like one. Tooltip and popover have opinions over when they get triggered and the copy button needed to have full control. It seemed less heavy to add tooltip-like styles and custom animations than to use a tooltip with manual trigger and force it to act like feedback (show, animate, hide without trigger).

packages/webawesome/docs/docs/resources/changelog.md
claviska: > Components with the [experimental] badge should not be used in production. They are made available as release candidates for development and testing purposes. As such, changes to experimental components will not be subject to semantic versioning.

I was concerned that people would think it was a breaking change we made haphazardly to a stable component.

KonnorRogers: I understand the hesitancy, but I do think people are using it (like 11ty) and would be good to call it out front and center 🤔 

Maybe we need a better way to call out breaking changes on experimental components?

lindsaym-fa (commented):
Pushed a few tweaks! I'd appreciate a second glance at the styling changes I made in https://github.com/shoelace-style/webawesome/pull/2343/commits/9ec0f893ecc9249609f7a07a27c9e95decda4ac0 and https://github.com/shoelace-style/webawesome/pull/2343/commits/8c00c849cfe6ceba6a7f76b70484dc1858e2daad — the squashed appearance and slightly-off-center vertical alignment were really starting to wear on me. 😅

Plus, a few possible things to fix up:

1. Looks like we lost `wa-tooltip`'s built-in hover and focus behavior since we moved to a bespoke popup. We're missing the `copy-label` on hover and focus.
2. Should we keep `tooltip-placement` around as a deprecated attribute? Or is that not worth it since it's not the only part of the API that's changed?

claviska:
> Looks like we lost wa-tooltip's built-in hover and focus behavior since we moved to a bespoke popup. We're missing the copy-label on hover and focus.

My thought was folks could apply tooltips directly to their own custom buttons or even to the copy button if they want one. While the feedback looks like a tooltip, it's really designed to just show copy feedback.

I'd be curious how many folks actually want the integrated tooltip behavior. To me it's kind of annoying, But I'd be on board with it as long as it can be turned off (while still showing copy feedback).

I could take a stab at implementing the feedback as a [manual tooltip trigger](https://webawesome.com/docs/components/tooltip#manual-trigger) if we really want this. It will just require juggling the tooltip's behavior during the feedback phase.

Let me know what you think!

lindsaym-fa:
A popup label is great default usability and helps make the mouse/keyboard experience closer to the screen reader experience (SRs currently hear the label, but mouse/keyboard don't see it). It has my vote!

It does only seem useful on the default trigger, though, in the same way the `copy-label` only seems to be read for the default trigger by SRs. If a user has slotted in their own trigger, they likely already have the label covered, so the popup/`copy-label` could be redundant or even confusing.

Side note: It doesn't sound like SRs get any feedback on success or error. That's also the case with the current copy button, so it's not a regression. Probably something to address, whether in this PR or a separate one.

claviska:
To be clear, is the consensus that we add the tooltip back to the default trigger but not custom triggers?

lindsaym-fa:
That's my vote.

claviska:
Even better. I was able to make the tooltip version work (with a slight live region trick for custom buttons) for built-in _and_ custom buttons.

I also added a `tooltip` attribute to control when it shows. Possible options are:

- `full` (default; shows on hover + copy)
- `copy` (only shows on copy)
- `none` (what tooltip?)

This makes the API more consistent regardless of how you use it and gives users more control over when the tooltip displays.

packages/webawesome/src/components/copy-button/copy-button.styles.ts
```diff
@@ -40,14 +43,6 @@ export default css`
...
   }
 
-  slot[name='success-icon'] {
-    color: var(--wa-color-success-on-quiet);
-  }
-
-  slot[name='error-icon'] {
-    color: var(--wa-color-danger-on-quiet);
-  }
-
```
lindsaym-fa, line 50: Curious: why remove the color change?

lindsaym-fa (dismissed):
Beautiful! ✨ VoiceOver reads the success and error labels aloud now too, so that's lovely!

packages/webawesome/src/components/copy-button/copy-button.styles.ts
claviska: Consistency with custom buttons. If a user slots in a custom button, they'd "lose" the color changes and they'd need to target their own icons with the `success | error` states via CSS in order to match the default behavior. I figured that would be confusing.

CleanShot 2026-05-07 at 14 31 41@2x

PR #2345, opened 2026-04-30
Add support for aria-label to wa-button

timmo001:
Useful for icon buttons which have no label and use an alternative popup (like `wa-popup`), but still want ally support


KonnorRogers:
This is generally _not_ the way to do this. Instead, we should look at supporting cross-root aria.

What usually ends up happening in this case is the aria-label gets read twice to users, which isn't ideal (haven't tested this specifically yet)

claviska:
For icon buttons, the label is read from the icon itself:

```html
<wa-button>
  <wa-icon label="Label goes here"></wa-icon>
</wa-button>
```

We don't map ARIA attributes by convention, and this will cause two labels to display in the AOM.

PR #2348, opened 2026-05-01
Remove Redundant `wa-gap-m` from Layout Utilities

talbs:
This work strips wa-gap-m from layout utility wrappers (wa-stack, wa-cluster, wa-grid) where it's already the default gap size. The demo row in gap.md's class table keeps its wa-gap-m, since the row is labeled wa-gap-m and exists to demonstrate that exact class.

### Companion PRs
- https://github.com/shoelace-style/webawesome-pro/pull/196
- https://github.com/shoelace-style/webawesome-app/pull/355


lindsaym-fa (approved):
We 🧡 deletions. Thanks for doing this!

PR #2358, opened 2026-05-05
Docs: Support Components Listing Revision + Polish 

talbs:
A companion PR to https://github.com/shoelace-style/webawesome-pro/pull/195.


packages/webawesome/docs/docs/components/index.njk
lindsaym-fa, line 1: Doesn't moving this file to Pro mean that folks running the core repo locally will get a 404 for `/docs/components`? Do we want to have some version of the page here as a fallback?

talbs: Hmmm, good point!

Our `webawesome-pro`-based template/view overrides the one currently in this repo. I've confirmed this locally.

So with that in mind, I'll close this PR as the file here can stay.

talbs:
Closing this as removing this file is not needed to override it for our own site (see https://github.com/shoelace-style/webawesome/pull/2358#pullrequestreview-4230791794).

PR #2360, opened 2026-05-05
Improve hover and active feedback

### LINKED ISSUE #2255: Weak hover and active feedback on the neutral (default) button
### Describe the bug
In the Default theme (Light mode), the hover and click feedback on the neutral (default) button is very weak and does not provide clear interaction feedback.

### To Reproduce

1. Go to https://webawesome.com/docs/components/button/
2. Hover over and click the neutral (default) button.

### Browser / OS
 - Browser: Any

Image

lindsaym-fa:
Closes #2255 with improved `color-mix()` values and new `transform` values.
Companion Pro PR: https://github.com/shoelace-style/webawesome-pro/pull/203

- Redefines `--wa-color-mix-hover` and `--wa-color-mix-active` in core themes to ensure better color compatibility
  - By default, `--wa-color-mix-hover` derives the mixed color from `currentColor` via `oklch()`, which avoids the gross desaturation effects we have today
  - By default, `--wa-color-mix-active` uses `var(--wa-color-surface-default)` to give the illusion of blending or pushing into the page/surface
- Adds `--wa-button-transform-hover` and `--wa-button-transform-active` as design tokens
  - Default theme has a subtle `scale()` effect on press/active with no effect on hover
  - Awesome theme's previously bespoke `translate()` effect on press/active is now applied via token
  - Shoelace theme sets these tokens to `none`
- Adds `transform` to buttons to leverage the new tokens

---

@claviska, I'm on the fence whether to use design tokens (`--wa-button-transform-hover`/`--wa-button-transform-active`) or button-scoped properties (`--transform-hover`/`--transform-active`) for the new `transform` effect. Enough themes already define their own transformations for buttons that this feels more like theme-level decision to me, but let me know if you have different thoughts.

@talbs, I'd love your 🪙🪙 on the updated button interactions. 

---

**Before** 🤢

https://github.com/user-attachments/assets/3d75f0ae-1ddb-441b-bd8c-54e9c0f153da

**After** 🤓

https://github.com/user-attachments/assets/7e2bdb3e-6fdb-4493-9916-2092b5045355


packages/webawesome/src/components/button/button.styles.ts
```diff
@@ -52,6 +53,27 @@ export default css`
...
+  .button:not(.disabled):not(.loading):active {
+    transform: var(--wa-button-transform-active);
+  }
+
+  @media (prefers-reduced-motion: reduce) {
+    .button:not(.disabled):not(.loading):hover {
+      transform: none;
+    }
+
+    .button:not(.disabled):not(.loading):active {
```
talbs: NIT: Can this be combined with the rule above to have two selectors?

packages/webawesome/src/styles/native.css
```diff
@@ -640,6 +643,17 @@
...
           );
+          transform: var(--wa-button-transform-active);
+        }
+
+        @media (prefers-reduced-motion: reduce) {
+          &:hover {
+            transform: none;
+          }
+
+          &:active {
```
talbs, line 654: NIT: Can this be combined with the rule above to have two selectors?

claviska:
> I'm on the fence whether to use design tokens (--wa-button-transform-hover/--wa-button-transform-active) or button-scoped properties (--transform-hover/--transform-active) for the new transform effect.

I think it makes sense to use the global token so other button-like components (and things like buttons baked into more complex components) can utilize them.

Nice work!

claviska (dismissed):

talbs (dismissed):
This is a super solid improvement IMHO.

The active states look a lot better thanks to the color choices and the default's "squish" (thanks to `--wa-button-transform-active`). I left a couple of NITs, but neither are blocking.

Nice work! Can't wait to squish these in production!

Issue #2331, opened 2026-04-24
Settings Pattern: text overflow in firefox


Ramblurr:
### To Reproduce

1. Open https://webawesome.com/docs/patterns/layouts/app-settings/ in firefox
2. Look at the grid of settings

### Screenshots

<img width="1143" height="786" alt="Image" src="https://github.com/user-attachments/assets/7fface1f-5b17-4bbe-b52f-b347cda872c6" />

### Browser / OS
 - OS: Linux
 - Browser: Firefox
 - Browser version: 149

### Additional information
Provide any additional information about the bug here.

kelseythejackson:
Hey @Ramblurr, thanks for letting us know about this. It seems we used a CSS property that is not yet supported in Firefox. We have a fix on the way. Thanks again.

kelseythejackson:
@Ramblurr, we have a fix in for this, and it'll be out in our next release. Thanks again!

PR #2361, opened 2026-05-06
Updated text-wrap property

kelseythejackson:
This is a fix for #2331 

In `native.css` we are using `text-wrap: pretty` for all `p` elements. That property is not yet supported in Firefox:
https://caniuse.com/mdn-css_properties_text-wrap_pretty

So I reverted it to `wrap` which according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-wrap#pretty) will give us more or less the result we want

Before
Screenshot 2026-05-06 at 12 07 36 AM

After

Screenshot 2026-05-06 at 12 07 45 AM

### DIFF
--- a/packages/webawesome/src/styles/native.css
+++ b/packages/webawesome/src/styles/native.css
@@ -109,7 +109,7 @@
   }
 
   p {
-    text-wrap: pretty;
+    text-wrap: wrap;
   }
 
   blockquote {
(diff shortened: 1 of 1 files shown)


kelseythejackson:
Update: I added `@supports` for `text-wrap`

lindsaym-fa (approved):
Good catch + fix!

PR #2366, opened 2026-05-06
Remove `line-height` from buttons

lindsaym-fa:
For buttons, we currently set `line-height: calc(var(--wa-form-control-height) - var(--wa-form-control-border-width) * 2);`, which causes trouble for buttons with complex content (see below). Since buttons are `inline-flex` containers and already center content in both directions, `line-height` doesn't seem to be necessary.

This PR removes the property from `wa-button` and native buttons. I haven't seen any ill effects from doing so, but please give a shout if I'm missing something.

---

**Before**
Screenshot 2026-05-06 at 5 00 11 PM

**After**
Screenshot 2026-05-06 at 5 02 03 PM

### DIFF
--- a/packages/webawesome/src/components/button/button.styles.ts
+++ b/packages/webawesome/src/components/button/button.styles.ts
@@ -37,7 +37,6 @@ export default css`
     font-family: inherit;
     font-size: inherit;
     font-weight: var(--wa-font-weight-action);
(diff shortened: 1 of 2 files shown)


claviska (approved):
No concerns. It's easy for folks to add back if they have an affected edge case.

PR #2368, opened 2026-05-07
Docs: Improve Prose Rhythm and Typography

talbs:
Web Awesome's native typography sets a uniform 24px gap below every block-level element — `h1–h6`, `p`, `ul`, `pre`, `figure`, all sharing the same `--wa-content-spacing`. In long-form docs copy, that uniformity made everything feel structurally flat: a major section break read the same as the next sentence. 

This PR adds a "Prose rhythm" block to `docs.css` (scoped to `#content`) that replaces the uniform gap with hierarchical, asymmetric rhythm. All values use existing `--wa-space-*` tokens.

### Heading Rhythm
Headings now hug the content they introduce. `h2` gets 40px above and 12px below; `h3` gets 32px above and 12px below; `h4–h6` keep their default top margin and tighten the bottom to 8px. Consecutive headings (e.g., an `h2` immediately followed by an `h3` with no body between) tighten to 16px so they don't double-space.

### Section Breaks and Major Non-Text Blocks
`<hr>` becomes a true section divider with 48px of breathing room on each side — the strongest visual break in the rhythm. Major content blocks (`<pre>`, `<figure>`, `<table>`, `<blockquote>`, `<wa-callout>`, `.code-example`) get 32px above and below when sitting between paragraphs. When they immediately follow a heading, they get a 16px minimum gap so the heading still hugs the block while the block keeps the air it deserves.

### Polish
Multi-line list items get 4px of breath between siblings. Display headings (`h1`, `h2`) get a hair of negative letter-tracking (`-0.01em`) for a tighter optical fit. Component API tables get `font-variant-numeric: tabular-nums` so numeric columns align cleanly.

### Screenies

| Before | After |
|--------|--------|
| image | image |
| image | image |
### Implementation
The block is scoped to `#content` and uses sibling combinators (`* + h2`) so first-on-page elements don't pick up extraneous top margins. The rules are unlayered, which lets them cleanly override `native.css`'s `wa-native` layer via cascade order.

PR #2370, opened 2026-05-07
Add Prose Text Utility

talbs:
🧪 This work is an experiment that spun off from #2368. 

It adds a new `wa-prose` utility class that applies hierarchical, asymmetric typographic rhythm to long-form content — docs pages, blog posts, marketing copy, changelogs.

It builds on `wa-native`'s base block styles rather than duplicating them, so headings, lists, code blocks, callouts, tables, and `<details>` all pick up consistent rhythm without each page reinventing it.

## Why

#2368 added prose rhythm to `docs.css` scoped to `#content`. That solved the docs problem, but the same rhythm is useful anywhere we render long-form content — blog posts, the changelog, future marketing pages, third-party sites embedding our themes.

Promoting it to an official utility means folks using Web Awesome get the rhythm by adding a class instead of nesting their content under a specific ID.

## Under the Hood

Spacing is **em-based**, not rem-based. That's the key design choice: apply any `wa-font-size-*` utility to the same container and the whole prose body — text, headings, *and* the spacing between them — scales proportionally. There's no need for a `wa-prose-small` variant; sizing is a separate axis.

The rhythm itself is tunable through custom properties on the container. `--wa-prose-line-length` (default `65ch`) caps the reading column. `--wa-prose-rhythm-scale` (default `1`) multiplies every margin in the system, so a sidebar can use `0.6` for a tighter rhythm without touching type sizes. Both compose cleanly with theme tokens — colors flow from the existing `--wa-color-*` palette, so dark mode and themes work without additional class variants.

The rhythm covers every heading level (h1 through h6), with back-to-back tightening so a subhead reads as subordinate to the heading above it instead of a peer with its own air. Major non-text blocks — `<pre>`, `<table>`, `<blockquote>`, `<wa-callout>`, and `<details>` — get more breathing room than running prose so they read as distinct chunks. Selectors are wrapped in `:where()` so every
  rule sits at specificity `0,1,0`, and authors can override with simple element or class rules without specificity tricks. Quieter typographic touches come along too: hanging punctuation on the prose container, oldstyle proportional figures for body text (tables reassert `tabular-nums` for column alignment), a `::marker` color override so list bullets and numbers don't compete with prose, and
`overflow-wrap: anywhere` on `<code>` and `<pre>` so long URLs and identifiers don't blow out the column.

For islands of UI inside a prose container (a button cluster, a custom layout), `wa-not-prose` reverts only the properties `wa-prose` touches via `revert-layer`. Other utilities like `wa-cluster` and `wa-stack` keep working inside the opt-out subtree.

## Additional Work Done

As part of this work, the Docs have been updated to use this utility. Here's how...

`base.njk` now adds `wa-prose` to `#content` whenever a page sets `proseContent: true`. A new `docs/docs.json` data file flips that flag for the entire docs section, so every component, utility, token, and resource page picks up prose rhythm without per-page changes. Marketing pages, index pages, and other layouts that opted out of prose rhythm before stay opted out.

* `docs.css` keeps the existing column-width tokens by anchoring `--wa-prose-line-length` to `--content-width-s` on `#content`.
* The docs remap `wa-prose`'s em-based rhythm tokens to the rem-based `--wa-space-*` scale on `#content` so the established docs cadence stays intact.
* `.code-example` live demos still get the major-block treatment via a small docs-level extension. The changelog's release-header tightening and prerelease-`h2` subordination patterns are preserved.
* Existing layouts (default, modern-card-list, page-wide) render at their established widths.
* The previous `#content`-scoped prose rules from #2368 are removed since the utility now owns them.

### DIFF
--- a/packages/webawesome/src/styles/utilities/prose.css
+++ b/packages/webawesome/src/styles/utilities/prose.css
@@ -106,46 +106,17 @@
       margin-block-end: var(--wa-prose-rhythm-2xs);
     }
 
-    /* `::marker` ignores opacity — bake de-emphasis into the color. */
-    & :where(li)::marker {
-      color: color-mix(in oklch, var(--wa-color-text-quiet) 75%, transparent);
-    }
-
-    & :where(dt) {
-      font-weight: var(--wa-font-weight-bold);
-    }
     & :where(dd):has(+ *) {
       margin-block-end: var(--wa-prose-rhythm-s);
     }
     /* #endregion */
 
-    /* #region Captions & block quotes ~~~~~~~~~~~~ */
+    /* #region Captions ~~~~~~~~~~~~ */
 
+    /* Em-based override keeps caption spacing in step with the prose
+       container; wa-native sets the visual style. */
     & :where(figcaption) {
-      color: var(--wa-color-text-quiet);
-      font-size: var(--wa-font-size-smaller);
-      line-height: var(--wa-line-height-condensed);
       margin-block-start: var(--wa-prose-rhythm-xs);
-    }
-
-    /* Em-based size keeps blockquote scaling with `wa-font-size-*`
-       (wa-native's rem-based --wa-font-size-l doesn't). */
-    & :where(blockquote) {
-      color: var(--wa-color-text-quiet);
-      font-size: 1.25em;
-    }
-    /* #endregion */
-
-    /* #region Tables ~~~~~~~~~~~~ */
-
-    /* Lining tabular figures so numeric columns align. Overrides the
-       container's oldstyle proportional figures. */
-    & :where(table) {
-      font-variant-numeric: tabular-nums;
-    }
-
-    & :where(th) {
-      border-block-end: var(--wa-border-style) var(--wa-border-width-s) var(--wa-color-border-normal);
     }
     /* #endregion */
 
@@ -196,7 +167,5 @@
     letter-spacing: revert-layer;
     font-variant-numeric: revert-layer;
     hanging-punctuation: revert-layer;
-    color: revert-layer;
-    border-block-end: revert-layer;
   }
 }
--- a/packages/webawesome/src/styles/utilities/prose.css
+++ b/packages/webawesome/src/styles/utilities/prose.css
@@ -1,0 +1,177 @@
+@layer wa-utilities {
+  /* `wa-prose` applies hierarchical, asymmetric typographic rhythm to a block
+     of long-form content. It builds on `wa-native`'s base block styles rather
+     than duplicating them, and uses em-based spacing internally so it scales
+     when composed with `wa-font-size-*` utilities. Colors flow from theme
+     tokens — no prose-specific color variants required. */
+  .wa-prose {
+    /* #region Tokens ~~~~~~~~~~~~ */
+
+    /* Em-based spacing scale that mirrors `--wa-space-*` numerically. Each
+       token resolves at the use site, so heading and block margins scale with
+       the container's font-size. `--wa-prose-rhythm-scale` lets advanced users
+       tighten or loosen all rhythm at once. */
+    --wa-prose-rhythm-scale: 1;
+    --wa-prose-rhythm-2xs: calc(0.25em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+    --wa-prose-rhythm-xs: calc(0.5em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+    --wa-prose-rhythm-s: calc(0.75em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+    --wa-prose-rhythm-m: calc(1em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+    --wa-prose-rhythm-l: calc(1.5em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+    --wa-prose-rhythm-xl: calc(2em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+    --wa-prose-rhythm-2xl: calc(2.5em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+    --wa-prose-rhythm-3xl: calc(3em * var(--wa-space-scale) * var(--wa-prose-rhythm-scale));
+
+    /* Comfortable reading column. Override or set max-inline-size directly. */
+    --wa-prose-line-length: 65ch;
+
+    max-inline-size: var(--wa-prose-line-length);
+    /* #endregion */
+
+    /* #region Headings ~~~~~~~~~~~~ */
+
+    /* Em-based heading sizes match WA's modular scale at the default root, so
+       `wa-font-size-*` utilities can rescale the whole prose body. Authors
+       overriding these should also use em values to preserve composition. */
+    & h1 {
+      font-size: calc(2.5625em * var(--wa-font-size-scale));
+    }
+    & h2 {
+      font-size: calc(2em * var(--wa-font-size-scale));
+    }
+    & h3 {
+      font-size: calc(1.5625em * var(--wa-font-size-scale));
+    }
+    & h4 {
+      font-size: calc(1.25em * var(--wa-font-size-scale));
+    }
+    & h5 {
+      font-size: 1em;
+    }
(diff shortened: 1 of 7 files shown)


talbs:
@claude, review this PR - focus on how this would work with Web Awesome Themes and native.css features.

talbs:
@claude, re-review this PR (and changes since your last feedback).

talbs:
>  _h6 size discrepancy with native.css (minor)_

The prose heading scale is intentionally distinct from `native.css`'s heading scale, not a port of it. None of the prose sizes match native — h1 is `2.5625em` against native's `--wa-font-size-3xl`, h5 sits at `1em` (body size) against native's `--wa-font-size-m`, and h6 drops to `0.75em` as a label-style step below body. Aligning only h6 to native's `--wa-font-size-s` would break the scale's internal ratio without bringing the others along. Leaving as is.

---

> _wa-font-size-* on an inner element breaks em-scaling for that element (known trade-off, worth documenting)_

Already covered by prose.md's "Composing with other utilities" section, which states that any utility class applied alongside wins automatically and shows a font-size: var(--wa-font-size-m) example.

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+    }
+    & :where(h4) {
+      font-size: 1.25em;
+    }
+    & :where(h5) {
+      font-size: 1em;
+    }
+    & :where(h6) {
+      font-size: 0.75em;
+    }
```
lindsaym-fa, line 57: In response to Claude's concern here, you say:
> The prose heading scale is intentionally distinct from native.css's heading scale, not a port of it.

However, each of the `em` sizes here do **exactly** match corresponding steps on the theme's type scale — assuming the root size is equal to the base font size of the element — _except_ for `h6`'s `0.75em`. E.g., `--wa-font-size-xl`, the font size specified for `h3` in `native.css`, computes to `1.5625rem`, the same value as `1.5625em` given the base size = root size.

It's hard for me not to read this as a port of the existing font scale based on relative sizes instead of the root font size when all but one size matches the theme's font scale. And it's hard not to question that one deviation.

I don't think setting `h6` to `0.75em` is the wrong choice, but it does seem like we should make the _same_ choice for `native.css` so that, until a user changes the base size of their prose to alter this relative sizing, the font sizes used by `prose.css` and `native.css` are the same.

Is there any reason we shouldn't update `native.css` so that `h6 { font-size: var(--wa-font-size-xs); }` to give both prose and not-prose `h6`s the same starting font size?

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+
+    & :where(ul, ol) > :where(li):has(+ li) {
+      margin-block-end: var(--wa-prose-rhythm-2xs);
+    }
+
+    /* `::marker` ignores opacity — bake de-emphasis into the color. */
+    & :where(li)::marker {
+      color: color-mix(in oklch, var(--wa-color-text-quiet) 75%, transparent);
+    }
```
lindsaym-fa: I think this ought to be laterally promoted to `native.css` (along with a number of other styles I'll note below). Any aesthetic changes beyond the relative type scale and tailored spacing feel like improvements that can/should apply outside of prose contexts, imo.

Two other things I wonder with list markers:

1. Should we use `currentColor` instead of `var(--color-text-quiet)`? That way, if someone uses a list in, say, `<wa-card>` with `appearance="accent"`, or with a different font color entirely, the marker has less risk of getting lost or clashing with the surrounding text. E.g.:
Image

2. Should we have different transparency levels for `ul` markers and `ol` markers? `ul` markers will typically fall in the category of "graphical interface elements," which means they only need a 3:1 contrast ratio with the background. On the other hand, `ol` markers are typically text, which means they'd need a 4.5:1 contrast ratio. Perceptually, the thinner strokes of the numerals will wear the change in transparency more noticeably than the filled discs. Maybe:
```css
ul > li::marker {
  color: color-mix(in oklch, currentColor 40%, transparent);
}

ol > li::marker {
  color: color-mix(in oklch, currentColor 60%, transparent);
}
```

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+
+    & :where(dt) {
+      font-weight: var(--wa-font-weight-bold);
+    }
```
lindsaym-fa: This style already exists in `native.css`. Since `prose.css` "[b]uilds on `wa-native`'s base block styles instead of duplicating them," should this be removed?

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+    /* #endregion */
+
+    /* #region Captions & block quotes ~~~~~~~~~~~~ */
+
+    & :where(figcaption) {
+      color: var(--wa-color-text-quiet);
+      font-size: var(--wa-font-size-smaller);
+      line-height: var(--wa-line-height-condensed);
+      margin-block-start: var(--wa-prose-rhythm-xs);
+    }
```
lindsaym-fa, line 120: Definitely worthy of lateral promotion to `native.css`, perhaps only keeping `margin-block-start` defined here with `margin-block-start: var(--wa-space-xs)` for the native styles. We (ashamedly) don't define any styles for `figcaption`.

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+
+    /* Em-based size keeps blockquote scaling with `wa-font-size-*`
+       (wa-native's rem-based --wa-font-size-l doesn't). */
+    & :where(blockquote) {
+      color: var(--wa-color-text-quiet);
+      font-size: 1.25em;
+    }
```
lindsaym-fa: Promote to `native.css` and redefine `font-size` to `var(--wa-font-size-larger)`.

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+    /* #endregion */
+
+    /* #region Tables ~~~~~~~~~~~~ */
+
+    /* Lining tabular figures so numeric columns align. Overrides the
+       container's oldstyle proportional figures. */
+    & :where(table) {
+      font-variant-numeric: tabular-nums;
+    }
```
lindsaym-fa: Promote to `native.css`.

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+
+    & :where(th) {
+      border-block-end: var(--wa-border-style) var(--wa-border-width-s) var(--wa-color-border-normal);
+    }
```
lindsaym-fa: Promote to `native.css`.

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -0,0 +1,202 @@
...
+      font-size: 1em;
+    }
+    & :where(h6) {
+      font-size: 0.75em;
+    }
+
+    /* Slight optical tracking; only the larger sizes need it. */
+    & :where(h1, h2) {
+      letter-spacing: -0.01em;
+    }
```
lindsaym-fa, line 62: No-op Note: I've been pining after some good ol' `letter-spacing` design tokens. The exact amount of spacing depends largely on the chosen font, so it falls in naturally at the theme level. I'll work on a PR for this.

lindsaym-fa (commented):
Lovely work! ✨ I left some comments on styles that I feel have a natural home in `native.css`, but of course feel free to push back if I'm just stir-crazy.

packages/webawesome/src/styles/utilities/prose.css
talbs: D'oh! Good catch! 

> but it does seem like we should make the same choice for native.css so that, until a user changes the base size of their prose to alter this relative sizing, the font sizes used by prose.css and native.css are the same.
>
> Is there any reason we shouldn't update native.css so that h6 { font-size: var(--wa-font-size-xs); } to give both prose and not-prose h6s the same starting font size?

Nope this makes good sense. I'll update this when I make the other `native.css` ports.

lindsaym-fa (approved):

Issue #2367, opened 2026-05-06
No border radius on <wa-button-group> when there is only one item


lxb-nvt:
### Describe the bug
When there is only one item in <wa-button-group> it looses its border-radius decoration (all squared).
The situation may happen with a dynamic button-group that can have only one item.

### To Reproduce
Create a <wa-button-group> with only one item (wa-button or wa-dropdown).

### Screenshots
<img width="122" height="62" alt="Image" src="https://github.com/user-attachments/assets/356a450e-e844-4bff-af67-8adc08931dbf" />

Record per page for a pager using <wa-button-group>. Situation when there is no need to display page selector.

<img width="275" height="60" alt="Image" src="https://github.com/user-attachments/assets/5163cf5b-d1f2-4634-b56c-1aef5524b2b7" />

### Browser / OS
 - OS: Ubuntu 24.04
 - Browser: Vivaldi
 - Browser version: 7.9.3970.60

ABWEBIT:
The issue is here.

https://github.com/shoelace-style/webawesome/blob/5964b9238d8c12762e20d4b73b0a79512b1e5617/packages/webawesome/src/components/button-group/button-group.styles.ts#L59-L82

Here is the fix.

```CSS
  :host([orientation='horizontal']) {
    ::slotted(:first-child:not(:last-child)) {
      --_button-start-end-radius: 0;
      --_button-end-end-radius: 0;
    }

    ::slotted(:last-child:not(:first-child)) {
      --_button-start-start-radius: 0;
      --_button-end-start-radius: 0;
    }
  }

  :host([orientation='vertical']) {
    ::slotted(:first-child:not(:last-child)) {
      --_button-end-start-radius: 0;
      --_button-end-end-radius: 0;
    }

    ::slotted(:last-child:not(:first-child)) {
      --_button-start-start-radius: 0;
      --_button-start-end-radius: 0;
    }
  }
```

claviska:
Fixed in #2371

PR #2371, opened 2026-05-07
Fix <wa-button-group> border radius when only one item is present

ABWEBIT:
Fix for https://github.com/shoelace-style/webawesome/issues/2367

before
before

after
after


claviska (approved):
Thanks for the PR! The fix looks great.

PR #2375, opened 2026-05-07
Fix popover on narrow screens

### LINKED ISSUE #2333: Current Popover implementation may cause horizontal overflow on small viewports
### Describe the bug
[Popover](https://webawesome.com/docs/components/popover/) may exceed parent/viewport dimensions, causing unintended scrolling behavior.

(Likely fix I found is included at the bottom)

### To Reproduce
1. [Go to the "Open and Closing" section of the Popover docs](https://webawesome.com/docs/components/popover/#opening-and-closing)
2. Use the developer tools to simulate mobile view, set width to exactly 400px (the page appears to not overflow between 390-400px, coincidentally)
3. Verify that no horizontal overflow is currently present
4. Click "Show popover"
5. Observe that the popover overflows and the viewport scrolls (forever?)
6. Close the popover
7. Observe that the viewport no longer scrolls

### Demo
CodePen wasn't playing nice with this, but you can quickly verify using any blank page.
1. Add the wa-popover element from the example
2. Constrain ```html <body>``` to a width that is less than the popover's contents

### Screenshots
This is at 404px, just before things should wrap:
Image

This is at 392px, where the popover now exceeds the viewport, thus content won't wrap
Image

This is at 404px, with the proposed fix as described below:
Image

### Browser / OS
 - OS: Windows/Android
 - Browser: Chrome, Firefox and Brave
 - Browser version: Latest

### Additional information
After unsetting/removing the width and max-width rules from [the popover's body slot](https://github.com/shoelace-style/webawesome/blob/f19d860c2fb1924d49e7c6754f71ce32ebb1084d/packages/webawesome/src/components/popover/popover.styles.ts#L81), the bug seems to be fixed.

CSS:
```css
wa-popover::part(body){
  max-width: unset;
  width: unset;
}
```

Was there a reason for why these properties were set? Maybe just a copy paste error from some similar part of the codebase? Either way, will link a draft PR if there's something I missed.

Thank you for this wonderful project! :)

claviska:
Fixes #2333 by adjusting the max width to match 100% of the viewport minus the popover's margin, while still allowing the `--max-width` when the screen size allows.

Also fixes an example that had a flex row that didn't wrap, causing mobile screens to scroll horizontally.

### DIFF
--- a/packages/webawesome/docs/docs/components/popover.md
+++ b/packages/webawesome/docs/docs/components/popover.md
@@ -68,7 +68,7 @@ Use `data-popover="close"` on any button inside a popover to close it automatica
 Use the `placement` attribute to set where the popover appears relative to its anchor. The popover will automatically reposition if there isn't enough space in the preferred location. The default placement is `top`.
 
 ```html {.example}
-<div style="display: flex; gap: 1rem; align-items: center;">
+<div style="display: flex; gap: 1rem; flex-wrap: wrap; align-items: center;">
   <wa-button appearance="filled" id="popover__top">Top</wa-button>
   <wa-popover for="popover__top" placement="top">I'm on the top</wa-popover>
 
@@ -129,7 +129,7 @@ Use the [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_a
 ```html {.example}
 <wa-popover for="popover__autofocus">
   <div style="display: flex; flex-direction: column; gap: 1rem;">
-    <wa-textarea autofocus placeholder="What's on your mind?" size="s" resize="none" rows="3"></wa-textarea>
+    <wa-textarea autofocus placeholder="What's on your mind?" size="s" resize="none" rows="2"></wa-textarea>
     <wa-button appearance="filled" variant="primary" size="s" data-popover="close"> Submit </wa-button>
   </div>
 </wa-popover>
--- a/packages/webawesome/src/components/popover/popover.styles.ts
+++ b/packages/webawesome/src/components/popover/popover.styles.ts
@@ -79,7 +79,7 @@ export default css`
     display: flex;
     flex-direction: column;
     width: max-content;
-    max-width: min(var(--max-width), 100vw - var(--wa-space-l) * 2);
+    max-width: min(var(--max-width), 100vw);
     padding: var(--wa-space-l);
     background-color: var(--wa-color-surface-default);
     border: var(--wa-panel-border-width) solid var(--wa-color-surface-border);
--- a/packages/webawesome/src/components/popover/popover.styles.ts
+++ b/packages/webawesome/src/components/popover/popover.styles.ts
@@ -79,7 +79,7 @@ export default css`
     display: flex;
     flex-direction: column;
     width: max-content;
-    max-width: var(--max-width);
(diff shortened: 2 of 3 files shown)


packages/webawesome/docs/docs/components/popover.md
```diff
@@ -129,7 +129,7 @@ Use the [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_a
 ```html {.example}
 <wa-popover for="popover__autofocus">
   <div style="display: flex; flex-direction: column; gap: 1rem;">
-    <wa-textarea autofocus placeholder="What's on your mind?" size="s" resize="none" rows="3"></wa-textarea>
+    <wa-textarea autofocus placeholder="What's on your mind?" size="s" resize="none" rows="2"></wa-textarea>
```
claviska, line 132: this just looks better

packages/webawesome/src/components/popover/popover.styles.ts
```diff
@@ -79,7 +79,7 @@ export default css`
     display: flex;
     flex-direction: column;
     width: max-content;
-    max-width: var(--max-width);
+    max-width: min(var(--max-width), 100vw - var(--wa-space-l) * 2);
```
lindsaym-fa: Without a `margin` applied to the popover body, this calculation only reserves space on one side of the popover. The popover nudges right up against the side of the viewport closest to the anchor. E.g.:

https://github.com/user-attachments/assets/7adab558-8a77-498d-9593-99c38819b520

I'm not sure if there's a good way to only add margin on the sides of the popover without an arrow. I know we could set `margin` based on the initial `placement`, but that fails in cases like `placement="left"` and `placement="right"` if the popover has to shift to the opposite side given the available viewport space. 

Our best bet might be to simply let the popover extend to the edges of the viewport (but no further):

```suggestion
    max-width: min(var(--max-width), 100vw);
```

packages/webawesome/src/components/popover/popover.styles.ts
```diff
@@ -79,7 +79,7 @@ export default css`
     display: flex;
     flex-direction: column;
     width: max-content;
```
lindsaym-fa: This seems to prevent `placement="left"` and `placement="right"` popovers from fitting within the viewport:

Image

Could we use `auto` instead, or does that cause different problems?


```suggestion
    width: auto;
```

claviska: yuck, good catch. It looks like `auto` does indeed work better here :D

claviska: CleanShot 2026-05-07 at 15 47 02@2x

lindsaym-fa (approved):

packages/webawesome/src/components/popover/popover.styles.ts
claviska: > but that fails in cases like placement="left" and placement="right" if the popover has to shift to the opposite side given the available viewport space

Good news, everyone. We have internal attributes to target for that! 💪 

CleanShot 2026-05-07 at 15 59 30@2x

CleanShot 2026-05-07 at 15 59 08@2x

PR #2386, opened 2026-05-12
Consolidate Pro Callout Styles + Extend the `:::pro` Directive

talbs:
## What changed

1. The styled "Get X with Web Awesome Pro!" callouts (combobox and other Pro component pages).
2. And the migrating-from-shoelace "A Few Components Now Live..." callout shared duplicated CSS across two locations. 
3. The `:::pro` markdown directive used a separate, smaller visual style. 

☝️ This work unifies them: a single `wa-callout.pro` rule in `utils.css` is now the source of truth, and the `:::pro` directive picks up the same styling automatically.                                                                                                                             

 ## What's new
The `:::pro` directive now accepts an optional title (`:::pro Some Title`) and an optional icon override (`:::pro {star-circle} Title`). The default icon is `hand-wave`.
                                                                                                                                               
 ## Page updates                                                                                                                          
The migration page drops its inline `<style>` block and the marketing-palette stylesheet `<link>` — both no longer needed since the shared  `.pro` rule uses `--wa-brand-orange` directly. The `/docs` and `/docs/frameworks/react` callouts now use the new `:::pro Using Web Awesome  Pro?` syntax for a styled title and body.

### DIFF
--- a/packages/webawesome/docs/docs/frameworks/react.md
+++ b/packages/webawesome/docs/docs/frameworks/react.md
@@ -17,6 +17,7 @@ To add Web Awesome to your React app, install the package from npm.
 ```bash
 npm install @awesome.me/webawesome
 ```
+
 Next, include the Web Awesome theme in your app, import the components you need, and start using them!
 
 ```jsx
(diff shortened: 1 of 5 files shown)


packages/webawesome/docs/docs/frameworks/react.md
```diff
@@ -143,8 +144,8 @@ Object.defineProperty(window, 'matchMedia', {
     removeListener: jest.fn(), // deprecated
     addEventListener: jest.fn(),
     removeEventListener: jest.fn(),
-    dispatchEvent: jest.fn()
-  }))
+    dispatchEvent: jest.fn(),
+  })),
```
claviska, line 148: Image

claviska (approved):

PR #2390, opened 2026-05-12
Docs: Standardizing Hover Utilities for Card Links

talbs:
This work introduces and applies two shared hover-card utilities — `hover-grow` (scale-on-hover) and `hover-emphasize-border` (brand border + outline ring on hover). 

**These utilities are meant to standardize some existing hover treatments we were using in different cases with cards**.                                                                                                   
   
### Promoting `hover-grow` to `utils.css`                                                                 
`hover-grow` and its `hover-grow-from-direct-child` sibling were previously defined in `webawesome-pro's `site.css`. They now live in`utils.css`, available here as well as the site/app CSS.

### Adding `hover-emphasize-border`
Based on our components, design utilities, and pattern grid listings, this is a new utility that pairs with `hover-grow` for card-grid links. On hover/focus-visible, the wrapped `wa-card` swaps its `border-color` to `var(--wa-color-brand-border-loud)` and gains a brand-color `outline` ring whose width scales with the theme via `--wa-border-width-s`. 

Using `outline` rather than `box-shadow` leaves `wa-card`'s resting elevation shadow intact across all appearances.                             
                                                                                                                                               
### Adopting across the docs card grids
Replaces the inline border-hover rules previously in `docs.css` (originally introduced in #1085) with the new utility classes on each anchor. The now-superseded inline rules are stripped; layout and typography rules stay put. This affects: `/docs/components`, `/docs/tokens`, `/docs/utilities`, and `/docs/frameworks`.

### DIFF
--- a/packages/webawesome/docs/assets/styles/docs.css
+++ b/packages/webawesome/docs/assets/styles/docs.css
@@ -267,11 +267,6 @@ wa-page > main:has(> .search-list) {
         min-block-size: calc(6rem + var(--spacing));
         padding: var(--spacing);
       }
-
-      &:hover {
-        border-color: var(--wa-color-brand-border-loud);
-        box-shadow: 0 0 0 0.0625rem var(--wa-color-brand-border-loud);
-      }
     }
   }
 }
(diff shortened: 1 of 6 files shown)


talbs:
@claude, review this PR.

lindsaym-fa (approved):

PR #2391, opened 2026-05-12
Docs: Sync Component-Based Badges Across Views

talbs:
This work introduces a shared `componentStatusBadges(component)` macro at `docs/_includes/macros/component-badges.njk` that renders the Stable/Experimental status badge and the "Since X.X" badge for a component. Both the individual component view (`docs/_layouts/component.njk`) and the free components index (`docs/docs/components/index.njk`) now call the macro instead of inlining their own badge markup.                                                                                                                         

### Stable Badge
Standardized on the individual-view treatment: `variant="brand"` with a check icon. The macro adds `class="preview"` to the Stable badge, so it opts out of `wa-theme-site` brand coloring and renders in the default brand (blue) wherever it appears — including on pages themed orange by the site palette.

### "Since X.X" badge
Standardized on the listing's subtler treatment. A new `wa-badge.since` rule in `docs/assets/styles/utils.css` (alongside `.pro`, `.free`, `.planned`) gives the badge a `--wa-color-surface-lowered` background so it recedes from the more prominent status badge. The macro applies the `since` class.

### DIFF
--- a/packages/webawesome/docs/_includes/macros/component-badges.njk
+++ b/packages/webawesome/docs/_includes/macros/component-badges.njk
@@ -1,0 +1,20 @@
+{# Renders the status (Stable/Experimental) and "Since X.X" badges for a component.
+   Used by the individual component view, the components index, and component cards. #}
+{% macro componentStatusBadges(component) %}
+{% if component.status == 'stable' %}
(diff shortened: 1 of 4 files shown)


claviska (approved):
ooh, nice one!

PR #2395, opened 2026-05-13
Docs: Polish Changelog View

talbs:
This work touches up our [Changelog](https://webawesome-git-talbs-ch-ch-changes-for-the-76d256-font-awesome.vercel.app/docs/resources/changelog) View.  Here's a rundown on what was updated...

#### Intro Copy
Stable and Experimental sections now lead the page with inline status badges and one-paragraph API contracts addressed to readers (not contributors). The Keep-a-Changelog reference is woven in so the new category groups feel intentional.

#### Visual Consistency
Stable / Experimental badges in changelog prose and `migrating-from-shoelace.md` now go through a shared `statusBadge` macro alongside the existing `componentStatusBadges`. 

Release dates moved from inline text to `<small><time datetime="...">…</time></small>`, with a new
prose-rhythm rule that pairs each `h2` + date as one tight header unit.

#### Category Groups w/ Iconography
Six new `markdown-it-container` plugins (`:::added`, `:::fixed`, `:::changed`, `:::deprecated`, `:::removed`, `:::breaking`) wrap each release's entries by type. 

A new `_transformers/changelog-list-icons.js` transformer injects a category-specific `<wa-icon>` bullet (plus, wrench, broom-wide, clock, xmark, triangle-exclamation), and sibling groups get an automatic divider. 

Applied across all 10 main releases (3.7.0 → 3.0.0) and all 14 prereleases.

#### GitHub References
The `replaceTextPlugin` output upgrades inline `[issue:N]` / `[pr:N]` / `[discuss:N]` mentions into outlined `<wa-badge>` chips with type-specific icons. 

#### Links to Component Docs
A new `_transformers/linkify-components.js` transformer wraps `<code><wa-*></code>` mentions inside `.changelog-group` in anchors to the component's docs page, validated against the live manifest. 

#### Anchor Permalinks
`anchor-headings.js` now emits the WA `wa-link-plain` + hashtag-icon pattern from the pro home, plus a `<wa-tooltip>` saying "Link to This Section". The anchor reveals on hover and brightens on direct hover.

#### Pre-release Section
"Pre-release Versions" is now a real `h2` (in the outline once), wrapping the existing `<details>` in `data-no-outline` so the 14 prerelease `h2`s don't clutter the outline. 

Each prerelease's entries got the same `:::category` treatment as main releases. Prerelease `h2`s are styled subordinate to main releases since they're historical.

#### Email Sign Up
The email signup moved from interrupting the intro to displaying right after the latest release. 

#### Feedback Footer
[...]

PR #2397, opened 2026-05-13
Make `linkify-components` Transformer Site-Wide

talbs:
This work expands the `linkify-components` doc transformer (started in https://github.com/shoelace-style/webawesome/pull/2395) from a changelog-only scope to the whole `#content` region, so any inline `<code>&lt;wa-*&gt;</code>` mention across the docs becomes a link to that component's page.

### Opt-outs

Adds a `data-no-linkify` attribute that works at three scopes: page (set on `<html>` via layout/front matter), region (any ancestor element), or single element. The transformer also skips code inside `<pre>`, `<a>`, `<button>`, and self-references where the target matches the current page, so component pages don't auto-link to themselves.

### Docs

New `docs/_transformers/README.md` documents the full transformer pipeline (order matters, each sees the previous one's output), the shared `data-no-*` opt-out convention (`data-no-anchor`, `data-no-outline`, `data-no-linkify`), and a short guide for adding new transformers.

### DIFF
--- a/packages/webawesome/docs/.eleventyignore
+++ b/packages/webawesome/docs/.eleventyignore
@@ -1,0 +1,1 @@
+_transformers
--- a/packages/webawesome/docs/_includes/base.njk
+++ b/packages/webawesome/docs/_includes/base.njk
@@ -4,8 +4,10 @@
 {% if hasThemeSelector == undefined %}{% set hasThemeSelector = true %}{% endif %}
 {% if hasSiteDialog == undefined %}{% set hasSiteDialog = true %}{% endif %}
 {% if hasGeneratedTitle == undefined %}{% set hasGeneratedTitle = true %}{% endif %}
-<html lang="en" data-fa-kit-code="38c11e3f20" data-version="{{ package.version }}" class="wa-cloak" {% if
-  hasAnchors == false %} data-no-anchor{% endif %}>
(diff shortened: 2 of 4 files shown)


talbs:
@claude, review this PR.

packages/webawesome/docs/.eleventyignore
```diff
@@ -0,0 +1 @@
+_transformers
```
KonnorRogers, line 1: Is there a reason this file needed to be added? Did something bad happen?

talbs: Nothing bad - I added [a new internal `README` doc](https://github.com/shoelace-style/webawesome/blob/b9e779e1a8d4b532b54dae60b60e8ca534df58cc/packages/webawesome/docs/_transformers/README.md) to that directory to help folks understand all of our [Transformers](https://www.youtube.com/watch?v=bVzf6_wUtlg).

When I pushed this, the Vercel build hit an error because Eleventy was picking up that file and not knowing what to do with it. So this makes sure it ignores the directory (instead of just the current guard against certain file types).

KonnorRogers: ah. the readme was tripping up 11ty, makes sense

KonnorRogers (approved):

Issue #2387, opened 2026-05-12
In wa-breadcrumb, a non-interactive wa-breadcrumb-item is rendered as interactive.


ABWEBIT:
### Describe the bug
Rendering a non-interactive element as interactive may confuse users and create incorrect semantic expectations.

### To Reproduce
1. Go to https://webawesome.com/docs/tokens/borders/
2. Click Borders in the breadcrumb navigation
3. Notice that the breadcrumb item appears interactive, but clicking it does nothing

https://github.com/user-attachments/assets/43e3c074-cdb1-4870-afb0-acabdc5adcbd

kelseythejackson:
Hey @ABWEBIT, thanks for bringing this to our attention. Per https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/ we're following the spec regarding the `cursor` over the current item. There is some discussion on this type of thing, and if the spec changes, we'll make sure to update it. I'm gonna go ahead and close this out. Thanks again for reaching out. If you see something else that feels iffy, don't hesitate to let us know.

ABWEBIT:
> Hey [@ABWEBIT](https://github.com/ABWEBIT), thanks for bringing this to our attention. Per https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/ we're following the spec regarding the `cursor` over the current item. There is some discussion on this type of thing, and if the spec changes, we'll make sure to update it. I'm gonna go ahead and close this out. Thanks again for reaching out. If you see something else that feels iffy, don't hesitate to let us know.

In the WAI-ARIA example, the last element is a fully functional link, so the pointer cursor is necessary to indicate that it is interactive.

PR #2398, opened 2026-05-13
Fix: Set `wa-breadcrumb-item` Render `href=""` as a Link

talbs:
### What Changed
`<wa-breadcrumb-item>` now renders as a link when `href` is set to an empty string, not just when it's a non-empty URL. 

This makes `href=""` a working shorthand for the current-page item — browsers resolve an empty `href` to the current URL, and `<wa-breadcrumb>` already adds `aria-current="page"` to the last item.

### Why
Follow-up to #2387. The [WAI-ARIA breadcrumb pattern](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/) calls for every item — including the current page — to be a link, with the current one marked `aria-current="page"`. The component's previous truthy check (`if (this.href)`) treated `""` the same as no attribute, so the current-page item couldn't be a link without specifying a real URL.

### The Fix
One-line change in `breadcrumb-item.ts`: `if (this.href)` → `if (this.href != null)`. Adds a test for the empty-string case, documents the convention in the Links example, and adds a changelog entry under the new Unreleased section.

### DIFF
--- a/packages/webawesome/src/components/breadcrumb-item/breadcrumb-item.ts
+++ b/packages/webawesome/src/components/breadcrumb-item/breadcrumb-item.ts
@@ -48,7 +48,7 @@ export default class WaBreadcrumbItem extends WebAwesomeElement {
       this.defaultSlot.assignedElements({ flatten: true }).filter(i => i.tagName.toLowerCase() === 'wa-dropdown')
         .length > 0;
 
-    if (this.href) {
+    if (this.href != null) {
       this.renderType = 'link';
       return;
     }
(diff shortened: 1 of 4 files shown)


talbs:
@claude, review this PR.

packages/webawesome/src/components/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -48,7 +48,7 @@ export default class WaBreadcrumbItem extends WebAwesomeElement {
       this.defaultSlot.assignedElements({ flatten: true }).filter(i => i.tagName.toLowerCase() === 'wa-dropdown')
         .length > 0;
 
-    if (this.href) {
+    if (this.href != null) {
```
claviska: nit: we use strict equality so maybe:

```suggestion
    if (this.href !== null && this.href !== undefined) {
```

claviska (dismissed):
Just one nit but not a blocker

PR #2399, opened 2026-05-13
Docs: Use New Size-Based Attribute Values

talbs:
In `3.6.0` we deprecated `small`, `medium`, and `large` size values in favor of `s`, `m`, and `l`. 

Our own docs examples should model the new convention, so this work pdates the single `<wa-button size="small">` example in `docs/docs/usage.md` (the Attributes & Properties section) to `size="s"`.


claviska (approved):
🧹

edge500 (approved):

PR #2400, opened 2026-05-14
Fix: Render `wa-breadcrumb-item` as a Button When `href` is Absent

talbs:
#2398 changed the render-type check in `setRenderType()` from `if (this.href)` to `if (this.href !== null)`. The intent was to render an `<a>` for `href=""` (the current-page item in a breadcrumb), which the old falsy check missed. But because `@property() href?: string` is `undefined` when the attribute is absent — not `null` — the new check evaluated `undefined !== null` as `true` and rendered every breadcrumb item as a link, even ones without an `href`.

This swaps the check to `typeof this.href === 'string'`, which correctly distinguishes the three cases: `undefined` renders a `<button>`, `""` renders an `<a href="">`, and any other string renders an `<a href="…">`. 

Existing tests for both behaviors now pass.


talbs:
@claude, review this fix!

claviska (approved):
Good call

PR #2403, opened 2026-05-14
Add `text-align`-Based Text Utilities + Adopt Flat `wa-text-` Naming

talbs:
This work introduces four new text alignment utility classes and refactors the existing wrapping utilities to adopt a flatter, more concise class naming convention.

### What's New

Four alignment utilities, all using logical values (`start`/`end` instead of `left`/`right`), so they adapt to the document's writing direction:

- `wa-text-start` → `text-align: start`
- `wa-text-center` → `text-align: center`
- `wa-text-end` → `text-align: end`
- `wa-text-justify` → `text-align: justify`

Physical values, `match-parent`, and `justify-all` are intentionally omitted — custom CSS covers the rare cases where they're needed.

### Naming Convention

PR feedback on #2404 raised that `wa-text-{property}-{value}` reads as overly verbose when the property name doesn't add disambiguation (`wa-text-transform-uppercase` is heavy for what it does). 

I surveyed how Bootstrap, Tailwind, Bulma, Primer, PatternFly, and other major design systems handle this and settled on a flat `wa-text-*` namespace where the class name encodes the value, not the property. 

Bootstrap has shipped this convention for 10+ years and has only had to qualify it for `text-decoration` (where values like `line-through` and `none` don't read well flat) — a known constraint that we can plan around if/when we support that property. 

But this would still be a departure from other CSS utility class names, so we'd have to be okay with that.

### Migration: `wa-text-wrap-*` → `wa-text-*`

The three wrap utilities from `3.7.0` get shorter names, with the original names kept as backwards-compatible aliases:

- `wa-text-wrap-nowrap` → `wa-text-nowrap`
- `wa-text-wrap-balance` → `wa-text-balance`
- `wa-text-wrap-pretty` → `wa-text-pretty`

Old names are flagged as deprecated in the docs and changelog. They'll be removed in a future major version.

### Docs

A new **Alignment** section between **Wrapping** and **Truncation** in `docs/docs/utilities/text.md`. The **Wrapping** section is updated to use the new names with a deprecation callout for the
originals.

### DIFF
--- a/packages/webawesome/src/styles/utilities/text.css
+++ b/packages/webawesome/src/styles/utilities/text.css
@@ -172,19 +172,21 @@
 
   .wa-text-capitalize {
     text-transform: capitalize;
-  .wa-text-align-start {
+  }
+
+  .wa-text-start {
     text-align: start;
   }
 
-  .wa-text-align-center {
+  .wa-text-center {
     text-align: center;
   }
 
-  .wa-text-align-end {
+  .wa-text-end {
     text-align: end;
   }
 
-  .wa-text-align-justify {
+  .wa-text-justify {
     text-align: justify;
   }
   /* #endregion */
--- a/packages/webawesome/src/styles/utilities/text.css
+++ b/packages/webawesome/src/styles/utilities/text.css
@@ -172,6 +172,20 @@
 
   .wa-text-capitalize {
     text-transform: capitalize;
+  .wa-text-align-start {
+    text-align: start;
+  }
+
+  .wa-text-align-center {
+    text-align: center;
+  }
+
+  .wa-text-align-end {
+    text-align: end;
+  }
+
+  .wa-text-align-justify {
+    text-align: justify;
   }
   /* #endregion */
 
--- a/packages/webawesome/src/styles/utilities/text.css
+++ b/packages/webawesome/src/styles/utilities/text.css
@@ -148,15 +148,18 @@
     white-space: nowrap;
   }
 
+  .wa-text-nowrap,
   .wa-text-wrap-nowrap {
     text-wrap: nowrap;
   }
 
+  .wa-text-balance,
   .wa-text-wrap-balance {
     text-wrap: balance;
   }
 
   @supports (text-wrap: pretty) {
+    .wa-text-pretty,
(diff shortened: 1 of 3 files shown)


claviska:
> But this would still be a departure from other CSS utility class names, so we'd have to be okay with that.

I think I'm okay with that. I would like to get things consistent for Web Awesome 4 though. 

In `contributing.md` we've stashed away a lot of patterns and conventions we use. It might be worth defining this in there.

There's a [Web Awesome 4 milestone](https://github.com/shoelace-style/webawesome/milestone/10) we can attach issues to to make sure we get around to any breaking changes like this :)

claviska (dismissed):

packages/webawesome/src/styles/utilities/text.css
```diff
@@ -148,19 +148,38 @@
     white-space: nowrap;
   }
 
+  .wa-text-nowrap,
   .wa-text-wrap-nowrap {
```
lindsaym-fa, line 152: Do we want to add a comment here for reinforcement?

```suggestion
  .wa-text-wrap-nowrap /* deprecated */ {
```

packages/webawesome/src/styles/utilities/text.css
```diff
@@ -148,19 +148,38 @@
...
     text-wrap: nowrap;
   }
 
+  .wa-text-balance,
   .wa-text-wrap-balance {
```
lindsaym-fa:
```suggestion
  .wa-text-wrap-balance /* deprecated */ {
```

packages/webawesome/src/styles/utilities/text.css
```diff
@@ -148,19 +148,38 @@
...
     text-wrap: balance;
   }
 
   @supports (text-wrap: pretty) {
+    .wa-text-pretty,
     .wa-text-wrap-pretty {
```
lindsaym-fa:
```suggestion
    .wa-text-wrap-pretty /* deprecated */ {
```

lindsaym-fa (dismissed):
Minor optional suggestions and a changelog conflict, but otherwise, approved!

PR #2404, opened 2026-05-14
Add `text-transform`-Based Text Utilities

talbs:
This work adds three new text transform utility classes for changing the case of text.

### What's New

- `wa-text-uppercase` → `text-transform: uppercase`
- `wa-text-lowercase` → `text-transform: lowercase`
- `wa-text-capitalize` → `text-transform: capitalize`

`full-width` (East Asian typography), `full-size-kana` (ruby annotation), and `math-auto` (MathML) are intentionally omitted — they're niche enough that custom CSS covers the rare projects that need them.

### What Changed Since the Initial Draft

The initial draft used `wa-text-transform-*` and also shipped a `none` value. Both have been revised based on review feedback:

- **Flat naming.** Class names are now `wa-text-uppercase` etc., not `wa-text-transform-uppercase`. The full convention rationale can be found in #2403, which establishes the pattern for all text utilities.
- **Dropping `none`.** `wa-text-none` is too semantically vague to ship — it doesn't read as anything specific, Bootstrap doesn't ship an equivalent, and resetting an inherited transform is rarely
needed as a utility. Easy to add back later if real cases surface.

### Docs

A new **Transform** section in `docs/docs/utilities/text.md`, near **Wrapping** and **Truncation**, with the standard accessibility callout about large blocks of uppercase text.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -19,10 +19,9 @@ These are still finding their shape. APIs can change between minor versions, so 
 :::added
 
 - Added the `capture` attribute to `<wa-file-input>` for capturing media directly from a device's camera or microphone [discuss:2380]
-- Added the `wa-text-transform-uppercase` text utility class for transforming text to uppercase
-- Added the `wa-text-transform-lowercase` text utility class for transforming text to lowercase
-- Added the `wa-text-transform-capitalize` text utility class for capitalizing the first letter of each word
-- Added the `wa-text-transform-none` text utility class for resetting inherited text transformations
+- Added the `wa-text-uppercase` text utility class for transforming text to uppercase
+- Added the `wa-text-lowercase` text utility class for transforming text to lowercase
+- Added the `wa-text-capitalize` text utility class for capitalizing the first letter of each word
 
 :::
 
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -19,6 +19,10 @@ These are still finding their shape. APIs can change between minor versions, so 
 :::added
 
 - Added the `capture` attribute to `<wa-file-input>` for capturing media directly from a device's camera or microphone [discuss:2380]
+- Added the `wa-text-transform-uppercase` text utility class for transforming text to uppercase
+- Added the `wa-text-transform-lowercase` text utility class for transforming text to lowercase
+- Added the `wa-text-transform-capitalize` text utility class for capitalizing the first letter of each word
+- Added the `wa-text-transform-none` text utility class for resetting inherited text transformations
 
 :::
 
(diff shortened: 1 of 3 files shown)


claviska:
I know we're following conventions here, but this is a bit hard to swallow considering the length compared to a raw style attribute. 😕 

```html
<div class="wa-text-transform-uppercase">

<div style="text-transform: uppercase;">
```

talbs:
> I know we're following conventions here, but this is a bit hard to swallow considering the length compared to a raw style attribute. 😕
> 
> ```
> <div class="wa-text-transform-uppercase">
> 
> <div style="text-transform: uppercase;">
> ```

Hmm, what do you think of `wa-text-uppercase` (and `wa-text-lowercase`, `wa-text-capitalize`, and `wa-text-none`following that convention)? To me, this feels lighter, but `wa-text-none` is vague/no bueno.

claviska:
Well `wa-text-*` could also refer to `text-align`, `text-indent`, `text-decoration`, `text-shadow`.

Just dumping another variant for discussion:

```
wa-tt-uppercase
wa-tt-lowercase
wa-tt-none
```

I'm not a red tape kind of guy, but we may want to standardize on a naming convention for CSS properties we want to turn into utilities so we don't paint ourselves into a corner.

For comparison, this is what Tailwind uses: https://tailwindcss.com/docs/text-transform

I don't think it's wise to provide utility classes without a prefix, but we could follow that lead for familiarity and brevity:

```
wa-lowercase
wa-uppercase
wa-normal-case
```

I'm not sold on any particular one so this is very much up for discussion.

talbs:
Yeah, I hear your thoughts/concerns and appreciate them.

A standardized naming convention for CSS properties turned Utilities sounds solid. Let's bring this and https://github.com/shoelace-style/webawesome/pull/2403 into our next Team Tactical to chat out directions.

talbs:
Okay, @claviska.

Thanks for pushing on verbosity — it definitely needed addressing.

After looking at how Bootstrap, Tailwind, Bulma, Primer, and PatternFly handle their text utilities, **I'm landing on flat `wa-text-uppercase`** — between the original `wa-text-transform-uppercase` and a bare `wa-uppercase`.

Bare names break down as soon as we hit alignment: `wa-center` collides with `wa-align-items-center`, and `wa-start`/`wa-end` collide with the equivalents there. We'd end up dropping the prefix for some properties but keeping it for others, which gets messy fast. Tailwind makes bare names work because they have no prefix at all; for us, `wa-` plus an internal category is what keeps the namespace scannable in class lists and autocomplete.

Flat `wa-text-*` cuts 9 characters per class (`wa-text-transform-uppercase` → `wa-text-uppercase`) while staying consistent with `wa-font-size-*`, `wa-font-weight-*`, `wa-align-items-*`, etc. Bootstrap has shipped this exact pattern for 10+ years and only had to qualify it for `text-decoration` — values like `line-through` and `none` don't read flat, which gives us a known constraint to plan around if/when we support that property.

lindsaym-fa (approved):

PR #2379, opened 2026-05-07
Popover mobile fix take 2

claviska:
@lindsaym-fa I somehow merged this prematurely while commenting...

https://github.com/shoelace-style/webawesome/pull/2375

This follows up with the remaining fixes discussed in those comments.

Issue #2396, opened 2026-05-13
Popover misaligned when using *-start and *-end placements


DanielKanyo:
Popover positioning is misaligned when using *-start or *-end placements (e.g. bottom-start, top-end, etc.).

This appears to be caused by the following CSS rule added to the popover:

```css
margin-inline: var(--wa-space-m);
```

The extra inline margin shifts the positioned element and causes incorrect alignment relative to the anchor/trigger element.

### To Reproduce

1. Create a popover with a placement such as `bottom-start` or `bottom-end`
2. Open the popover
3. Observe that the popover is horizontally offset from the trigger element

### Demo

A minimal reproduction can be created using any popover example with `*-start` or `*-end` placement values.

Example placements affected:

* `bottom-start`
* `bottom-end`
* `top-start`
* `top-end`
* `left-start`
* `right-end`

Removing the following CSS rule fixes the alignment issue:

```css
margin-inline: var(--wa-space-m);
```

### Screenshots

<img width="295" height="323" alt="Image" src="https://github.com/user-attachments/assets/5d0d670b-345b-4362-8d67-3f15ca58a825" />

### Browser / OS

* OS: Windows 11
* Browser: Chrome
* Browser version: 148

### Additional information

The issue does not appear with centered placements such as `bottom` or `top`, only with directional alignments using `-start` and `-end`.

The bug was introduced here: https://github.com/shoelace-style/webawesome/pull/2379/changes

PR #2411, opened 2026-05-18
Fix padding bug in `<wa-popover>`

claviska:
Fixes the regression described here: https://github.com/shoelace-style/webawesome/pull/2379#issuecomment-4438753150

It now uses `shift-padding` instead of custom margin wizardry.


DanielKanyo (approved):
Looks good from my point of view, good job! I created an [issue](https://github.com/shoelace-style/webawesome/issues/2396) for this, you can link there the PR and close it. Thank you!

talbs (dismissed):
[👀 good. Gettin' 🎁🎁🎁.](https://www.marriedtothesea.com/index.php?date=031908)

PR #2424, opened 2026-05-20
Improve placement of content in textarea when exceeding rows

roydukkey:
Demo: https://codepen.io/roydukkey/pen/ByQRdMB?editors=1100

https://github.com/user-attachments/assets/778ddea4-9fc2-4502-8532-93bd85599320

This improves the vertical placement of content in the textarea when the input exceeds the amount of visible rows. Instead of having the new line smushed right against the bottom border, it will be offset from the border.


claviska:
CleanShot 2026-05-22 at 10 55 34@2x

I'm not sure why, but the focus ring is getting clipped in the preview. It's clearly not in your video, so there may be something else causing it.

roydukkey:
> CleanShot 2026-05-22 at 10 55 34@2x
> I'm not sure why, but the focus ring is getting clipped in the preview. It's clearly not in your video, so there may be something else causing it.

Yeah. That was happening for me too, even when on the `next` branch.

claviska:
This seems to work really well, but I'd like to ask @lindsaym-fa for a review to triple check things :)

lindsaym-fa (dismissed):
Great improvement! Thanks, @roydukkey!

lindsaym-fa (approved):

PR #2430, opened 2026-05-22
Fix dropdown keyboard scroll

wendevlin:
Fix the dropdown keyboard nav with a scrollable body to scroll the selected option into view when needed.


claviska (approved):
Tested, works great. Thanks for the PR!

PR #2431, opened 2026-05-22
Docs: Unifying Component Category Labels

talbs:
This work collapses the canonical front-matter `category:` labels to match what the components index filter on `/docs/components` displays, so the same names work in both places without per-page translation. 

- Form Controls becomes Forms
- Feedback & Status becomes Feedback
- Organization becomes Layout
- Imagery becomes Media. 
- Actions, Navigation, and Utilities keep their existing labels.

### Why

The filter UI used short, conversational labels (Forms, Layout) while the front matter still used the longer canonical ones (Form Controls, Organization). 

That forced the components index code to maintain a translation table. Bringing the two into alignment kills that translation step and makes adding a new category a single-line change.

### DIFF
--- a/packages/webawesome/docs/docs/components/animated-image.md
+++ b/packages/webawesome/docs/docs/components/animated-image.md
@@ -1,7 +1,7 @@
 ---
 title: Animated Image
 layout: component
-category: Imagery
+category: Media
 synonyms:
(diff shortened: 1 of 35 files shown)


claviska (approved):
✨

PR #2434, opened 2026-05-26
Add <wa-accordion> and <wa-accordion-item> components

kelseythejackson:
Adds two new components — `<wa-accordion>` and `<wa-accordion-item>` — providing a vertically stacked, keyboard-accessible set of expandable sections. 

### Components

**`<wa-accordion>`** 

- `single` attribute — restricts the accordion to one open panel at a time; opening a new item automatically collapses the previously open one
- `icon-placement` attribute (`start` | `end`, default `end`) — propagated to all child items automatically on initial render and whenever the property changes, via a `@watch` handler + slot change listener
- `expandAll()` / `collapseAll()` public methods for programmatic control over all panels; `expandAll()` is a no-op when `single` is set
- Four events with `{ item: WaAccordionItem }` in `event.detail`: `wa-expand` and `wa-collapse` are cancelable (`event.preventDefault()` blocks the action); `wa-after-expand` and `wa-after-collapse` fire after animations complete - `--border` and `--radius` CSS custom properties so the outer frame and item dividers can be restyled in one place; dividers use `::slotted(wa-accordion-item + wa-accordion-item)` so the same `--border` value applies without extra markup

**`<wa-accordion-item>`**

- `label` attribute for plain-text headers; `label` slot for HTML content (icons, badges, etc.)
- `expanded` and `disabled` attributes both reflect to the host element
- `icon` slot to swap the default chevron for any icon; the chevron rotates 90° on expand and -90° in RTL via a CSS `rotate` transition
- `icon-placement` is set by the parent accordion but can also be applied directly on the item; `start` placement uses `flex-direction: row-reverse` on the button to push the icon left without changing DOM order
- `expand()`, `collapse()`, and `toggle()` return promises that resolve when the animation completes, using the existing `animate()` + `waitForEvent()` utilities
- An `animationGeneration` counter guards against stale animation callbacks when `expand()`/`collapse()` is called faster than the current animation can finish
- CSS custom properties: `--padding` (applied to both the header button and the content area), `--duration`, `--easing`
- `:state(animating)` custom state is active during height transitions so consumers can hook into the animating phase in CSS
- `focus()` delegates to the internal trigger button


kelseythejackson:
@MarcMagonMNS It's not, they're two distinct elements.

DanielKanyo:
What is the difference between this and [wa-details](https://webawesome.com/docs/components/details/)?

claviska:
@DanielKanyo they're similar, but details is a one-off disclosure element. An accordion is a coordinated group of disclosures with shared rules (single-vs-multi-open, controlled state, keyboard support, animations, etc.).

There's more info about this pattern here: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/

claviska:
Nice! Here are some initial thoughts:

## Headings

Hardcoding an `<h3>` forces an opinion on document structure that users can't change or opt out of. It may be tempting to consider an attribute to control the level (some libraries do this), e.g. `heading-level="2"`, but I think it would be better to let users slot headings in when they want the expand/collapse button to use such a structure.

So instead of of the heading living in the shadow root, the user will be able to choose which heading (or an alternative if they don't want to use one) by slotting it:

```html
<wa-accordion>
  <wa-accordion-item>
    <h2 slot="summary">Personal Information</h2>
    <p>Name, email, phone, etc.</p>
  </wa-accordion-item>
  ...
</wa-accordion>
```

## CSS Property Documented

For some reason, the `css` property is showing up in the properties table. It may need to be made private or have its JSDoc comment removed so the CEM ignores it.

CleanShot 2026-05-27 at 09 13 56@2x

## RTL

The chevrons should be reversed (along with the expand/collapse animation) in RTL mode. (You can test this by adding `dir="rtl"` to the `<html>` tag in the JS console.)

CleanShot 2026-05-27 at 09 15 55@2x

## Aesthetics

Could we add borders so each button and region is more obvious? In this screenshot, I also removed the internal `<h3>` and made the label semibold. (If users slot in headings, it's fine for them to render as headings. We can show an example of this further down if desired.)

CleanShot 2026-05-27 at 09 20 03@2x

## Focus Ring Clipped

The focus ring appears to be clipped:

CleanShot 2026-05-27 at 09 27 07@2x

## Expanded vs. Open

While I think `expanded` is probably a better name, `<wa-details>` uses `open` and the only other component we have that uses `expanded` is `<wa-tree>`. I worry this may confuse users because of the similarities between details and accordions.

_However_, one could also argue that `<details>` uses `open` over expanded, so we mirrored that to be consistent.

This is probably a no-op, but I wanted to call it out as an observation.

## Keyboard Support

The accordion adheres to the APG pattern. I would suggest using a roving tab index to also support arrow keys when focused on an accordion header. This is an optional pattern, but it lets users navigate between accordions more easily.

## Single Mode

Is there a more appropriate name for this? It forces only one accordion to be expanded at a time, but the name doesn't feel like it matches the behavior.

## Sizing

How does the accordion get resized, if at all? We should document it and show some examples.

While you can scale the font size yourself, the chevron's and padding won't change which will look strange. Perhaps it should be em-based so the entire UI scales with the users font size. (Open to alternate suggestions here.)

claviska:
Thanks for the updates! This is almost ready. Feedback below:

---

The API looks good, but I'm still not sure about a single boolean prop to control single/multiple mode. That paints us into a corner is we want to support other behaviors, e.g.

- `single` — exactly one panel open at all times. Opening another switches to it; you can't close the open one by -clicking it again. There's always one expanded.
- `single-collapsible` — at most one panel open. Same as single except clicking the open panel closes it, so zero open items is a valid state.
- `multiple` — any number open at once, each toggles independently.

So perhaps `mode="single | multiple"` is better than `exclusive` so we can add to it later on.

---

While I agree with the APG on the heading semantics that I previously overlooked, the default styles for them aren't great. The font size and padding make the accordion feel like a toy:

CleanShot 2026-06-01 at 11 09 17@2x

CleanShot 2026-06-01 at 11 23 57@2x

Semantically, the label is a heading but aesthetically it's already emphasized because of the surrounding box and padding. I think we should inherit the font instead. This is already demonstrated when `heading-level` is set to `none`.

CleanShot 2026-06-01 at 11 15 18@2x

Since the entire accordion is em-based, setting the font size effectively scales the whole thing up. The heading level shouldn't cause the visuals to change, as users probably won't expect a specific font size to render seemingly arbitrary heading sizes in part of the component. (This can also create unexpected results if headings use `rem` vs `em` and other relative units.)

It's hard to tell each section apart outside of the chevron, which can probably be balanced by a heavier font weight (bold shown, but perhaps semibold is more appropriate):

CleanShot 2026-06-01 at 11 19 00@2x

Removing the border and tightening up the padding like we do in `<wa-details>` makes the relationship more obvious, but it feels like something may be missing. 🤔 

CleanShot 2026-06-01 at 11 22 35@2x

Let's look at what some other popular libraries are doing in comparison and draw inspiration from there.

---

Outlines are still somewhat trimmed here. The code preview clipping issue seems to be resolved in next, so this may be stale. If not, I'd consider it a bug:

CleanShot 2026-06-01 at 11 27 30@2x

---

Nit — it'd be cool if this rotated 225º and landed as an X when expanded:

CleanShot 2026-06-01 at 11 28 37@2x

And it'd be nice to use the second icon in that example to show users how they can prevent the icon animation and maybe even swap + to - when its expanded.

---

Nit — I'd suggest moving the buttons below the example and using `appearance="filled"` for consistency with the rest of the docs (accent is a CTA and this isn't a CTA button):

CleanShot 2026-06-01 at 11 30 47@2x

---

It's not obvious because we're locked in to using this convention in many components until Web Awesome 4, but we're trying to get rid of the base part:

CleanShot 2026-06-01 at 11 42 22@2x

At a glance it looks like it might be redundant so we can just remove it. If it _is_ required, can we find a more appropriate name?

---

Is it possible for either or both of these to be inherited from the host? This would be more ergonomic, but I'm not sure at a glance if it's possible to do given the structure of the template.

CleanShot 2026-06-01 at 11 44 30@2x

lindsaym-fa:
Suggestions from a passerby —

### Relationship between button + content
Absent any tweaks to the heading styling as Cory suggested, here are a couple of ideas (if we need 'em) for better presenting the relationship between each button and its content.

Taking a cue from `<wa-tab-group>` and using the brand color with a bottom border to serve as an "indicator":
Screenshot 2026-06-01 at 4 57 51 PM

Or, treating the content as a "well":
Screenshot 2026-06-01 at 5 01 28 PM

### CSS custom properties

#### `<wa-accordion>`
Given `--border` and `--radius` both map so cleanly to design tokens, I'm not sure they'd warrant their own component-level custom properties. Personally, I'd lean towards removing these; if a style can be predictably themed with design tokens and/or overridden with CSS parts, we don't usually use a custom property.

That said, I understand that these values are used multiple times within both `<wa-accordion>` and `<wa-accordion-item>`. I still believe the argument above stands, or maybe some refactoring plus inheritance would mitigate this, but I'd happily defer to your judgement here.

#### `<wa-accordion-item>`
We don't seem to be using design tokens from the default values for any of these properties. That seems fine for `--duration` and `--easing`, but it seems like `--padding` should default to `var(--wa-space-m)` to match similar components.

Also, we tend to use `--spacing` instead of `--padding`, given it's usually agnostic to _how_ the spacing is applied. [I know that's caused us some trouble](https://github.com/shoelace-style/webawesome/discussions/2278#discussioncomment-16773953), but it seems like we should stick with consistency until we get a chance to redress in WA 4.

Finally, for `<wa-details>`, we have separate properties for `--show-duration` and `--hide-duration`. Would splitting `--duration` into two matching properties be relevant here?

claviska (approved):
:shipit:

PR #2443, opened 2026-05-29
Docs: Link to a Component's Category

talbs:
This work updates component detail pages (e.g. `/docs/components/avatar/`) so that they carry a clickable category badge in the meta row next to the tag, status, and "Since" badges. This badge links to `/docs/components/?category={slug}` with the index filtered to that category — closing the loop between the Components Listing and the Component Detail view.

### New Shared Data File

`docs/_data/componentCategories.json` holds the label, slug, and icon for each category. It's the single source of truth for category metadata — the pro repo's components index filter loops over the same file, so the icon/slug set can't drift between the two consumers.

### `componentStatusBadges` → `componentMetaBadges`

The badges macro was renamed to reflect what it now renders: status, optional category link, and "Since" pill, in that order. Card callers (the index grid) omit the category args and skip the link; the detail-page layout passes the category in.

### Subtler-Fill Badge Styling Generalized

The "Since X.X" badge's `surface-lowered` background was previously keyed to a `.since` class. That override now lives on `wa-badge[variant='neutral'][appearance='filled']`, so the category badge picks up the same treatment automatically. `.pro`, `.free`, and `.planned` all carry their own class-based backgrounds and are unaffected.

### Heads Up

The pro repo's components index card needs to use the renamed macro at the same time this lands — see shoelace-style/webawesome-pro#224.

### DIFF
--- a/packages/webawesome/docs/_data/componentCategories.json
+++ b/packages/webawesome/docs/_data/componentCategories.json
@@ -1,0 +1,10 @@
+[
+  { "label": "Actions", "slug": "actions", "icon": "hand-pointer" },
+  { "label": "Feedback", "slug": "feedback", "icon": "comments" },
+  { "label": "Forms", "slug": "forms", "icon": "pen-field" },
+  { "label": "Media", "slug": "media", "icon": "image" },
+  { "label": "Navigation", "slug": "navigation", "icon": "compass" },
+  { "label": "Layout", "slug": "layout", "icon": "layer" },
+  { "label": "Utilities", "slug": "utilities", "icon": "wrench" },
+  { "label": "Data Viz", "slug": "data-viz", "icon": "chart-line" }
+]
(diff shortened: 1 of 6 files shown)


claviska (dismissed):
Seems fine. The clickable badge is visually the same as the "since" badge which is not clickable (not sure that even matters though).

talbs:
> The clickable badge is visually the same as the "since" badge which is not clickable 

Yeah, this is a good point. I originally had a different styling to differentiate the badges, but that made the metdata under a component's name too noisy. I settled for keeping the category and since vx.x badges the same for some unity with the understanding that this is for the curious and not a super critical wayfinding element.

That cool by you, @claviska ?

---

One thought I had was that we could set the since vx.x badge link to the changelog entry for that version (if it is listed). Worth doing?

claviska:
> One thought I had was that we could set the since vx.x badge link to the changelog entry for that version (if it is listed). Worth doing?

Oh that's neat! I think we can keep it simple and assume that the fragment exists. If it "fails" it still lands on the changelog.

talbs:
Okay, With https://github.com/shoelace-style/webawesome/pull/2443/changes/d48fa6303e77dac0d3169b272d0892aed9692cca the "Since" version-based badge on Component views now links to the changelog (with version-specific anchors).

claviska (approved):

PR #2446, opened 2026-06-01
Add tests + changelog

claviska:
- Adds tests for the fix in #2437
- Adds a changelog entry

### DIFF
--- a/packages/webawesome/src/components/dialog/dialog.test.ts
+++ b/packages/webawesome/src/components/dialog/dialog.test.ts
@@ -184,6 +184,42 @@ describe('<wa-dialog>', () => {
         });
       });
 
+      describe('light dismiss', () => {
+        it('should not close when clicking the backdrop by default', async () => {
+          const el = await fixture<WaDialog>(html`<wa-dialog open>Content</wa-dialog>`);
+          const dialog = el.shadowRoot!.querySelector<HTMLDialogElement>('.dialog')!;
+
+          // Simulate a backdrop click by dispatching pointerdown with the dialog as the target
+          dialog.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true, composed: true }));
+          await aTimeout(250);
+
+          expect(el.open).to.be.true;
+        });
+
+        it('should close when clicking the backdrop when light-dismiss is enabled', async () => {
+          const el = await fixture<WaDialog>(html`<wa-dialog open light-dismiss>Content</wa-dialog>`);
+          const dialog = el.shadowRoot!.querySelector<HTMLDialogElement>('.dialog')!;
+
+          await expectEvent(el, 'wa-after-hide', () => {
+            // Simulate a backdrop click by dispatching pointerdown with the dialog as the target
+            dialog.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true, composed: true }));
+          });
+
+          expect(el.open).to.be.false;
+        });
+
+        it('should not close when clicking inside the dialog even when light-dismiss is enabled', async () => {
+          const el = await fixture<WaDialog>(html`<wa-dialog open light-dismiss>Content</wa-dialog>`);
+          const body = el.shadowRoot!.querySelector<HTMLElement>('[part="body"]')!;
+
+          // pointerdown originates from the body, not the backdrop, so it should not dismiss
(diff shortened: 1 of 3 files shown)


packages/webawesome/src/components/dialog/dialog.test.ts
```diff
@@ -184,6 +184,42 @@ describe('<wa-dialog>', () => {
...
       });
 
+      describe('light dismiss', () => {
+        it('should not close when clicking the backdrop by default', async () => {
+          const el = await fixture<WaDialog>(html`<wa-dialog open>Content</wa-dialog>`);
+          const dialog = el.shadowRoot!.querySelector<HTMLDialogElement>('.dialog')!;
+
+          // Simulate a backdrop click by dispatching pointerdown with the dialog as the target
+          dialog.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true, composed: true }));
+          await aTimeout(250);
```
KonnorRogers, line 194: rather than awaiting a timeout, why not:


```
await waitUntil(() => el.open === true)
```

packages/webawesome/src/components/dialog/dialog.test.ts
```diff
@@ -184,6 +184,42 @@ describe('<wa-dialog>', () => {
...
+          expect(el.open).to.be.false;
+        });
+
+        it('should not close when clicking inside the dialog even when light-dismiss is enabled', async () => {
+          const el = await fixture<WaDialog>(html`<wa-dialog open light-dismiss>Content</wa-dialog>`);
+          const body = el.shadowRoot!.querySelector<HTMLElement>('[part="body"]')!;
+
+          // pointerdown originates from the body, not the backdrop, so it should not dismiss
+          body.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true, composed: true }));
+          await aTimeout(250);
```
KonnorRogers, line 217: Same here:


```js
await waitUntil(() => el.open === true)
```

packages/webawesome/src/components/drawer/drawer.test.ts
```diff
@@ -209,6 +209,42 @@ describe('<wa-drawer>', () => {
...
       });
 
+      describe('light dismiss', () => {
+        it('should not close when clicking the backdrop by default', async () => {
+          const el = await fixture<WaDrawer>(html`<wa-drawer open>Content</wa-drawer>`);
+          const dialog = el.shadowRoot!.querySelector<HTMLDialogElement>('.drawer')!;
+
+          // Simulate a backdrop click by dispatching pointerdown with the dialog as the target
+          dialog.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true, composed: true }));
+          await aTimeout(250);
```
KonnorRogers, line 219: same here:

```js
await waitUntil(() => el.open === true)
```

packages/webawesome/src/components/drawer/drawer.test.ts
```diff
@@ -209,6 +209,42 @@ describe('<wa-drawer>', () => {
...
+          expect(el.open).to.be.false;
+        });
+
+        it('should not close when clicking inside the drawer even when light-dismiss is enabled', async () => {
+          const el = await fixture<WaDrawer>(html`<wa-drawer open light-dismiss>Content</wa-drawer>`);
+          const body = el.shadowRoot!.querySelector<HTMLElement>('[part="body"]')!;
+
+          // pointerdown originates from the body, not the backdrop, so it should not dismiss
+          body.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true, composed: true }));
+          await aTimeout(250);
```
KonnorRogers, line 242: same here:

```js
await waitUntil(() => el.open === true)
```

KonnorRogers (changes requested):

packages/webawesome/src/components/dialog/dialog.test.ts
claviska: The backdrop click occurs when the dialog is already open, so `el.open` is already true at that point. We're trying to prove that it _doesn't_ close, so an arbitrary timeout seems reasonable unless you have a more reliable suggestion. 🤷🏻‍♂️

KonnorRogers: ah missed that. Shit. Wish there was a way to not have an arbitrary timeout 🤔 

leave it 🤷 just dont change animation times 😓

PR #2476, opened 2026-06-05
Generate Agent Skill component API tables from the CEM

### LINKED ISSUE #2475: Agent Skill generator garbles component API tables (drops words, mashes type/default)
### Describe the bug

`packages/webawesome/scripts/agent-skill.js` builds each `references/components/*.md` by scraping the rendered docs HTML. Those API tables are themselves generated from the Custom Elements Manifest, so scraping them back to markdown is a lossy round-trip. In `processHtmlToMarkdown` (~L219) each cell is flattened with `td.textContent`, only the first `<code>` is kept, and `td.textContent.replace(firstCodeText, '')` runs — so per-row type/default/description collapse into one cell (a literal `Type Default`), `<code>` after the first loses its backticks, and the blind `replace` drops words.

### To Reproduce

1. `npm run build`
2. Open `dist/skills/webawesome/references/components/button.md` → "Attributes & Properties"
3. The `appearance` row collapses type + description + default into one cell ending `… Type Default 'accent'`; the `download` row reads *"Only used when ​ is present"* — the inline `href` was deleted.

### Demo

N/A — reproducible from the repo build above.

### Browser / OS

N/A (build tooling, Node).

### Additional information

The clean data is already loaded: `generateAgentSkill` reads the CEM (`custom-elements.json`) for the SKILL.md component list, and `scripts/llms.js` already generates its API reference from the CEM. Generating the component tables from the CEM instead of scraping HTML yields clean `Property | Attribute | Description | Type | Default` rows. I have a fix ready along those lines and can open a PR against `next` — or adjust to a smaller fix of the cell extractor if you'd prefer.

paulscan:
Fixes #2475.

The Agent Skill generator scrapes the rendered docs HTML for component API tables, but those tables already come from the Custom Elements Manifest, so scraping them back loses the type and default columns, drops backticks, and deletes words (for example `<wa-button>`'s `download` reads *"Only used when is present"* because the inline `href` was stripped). This reads the tables straight from the CEM instead, the same way `scripts/llms.js` already does. Only the API tables change; examples, Importing, and dependencies are left alone.

Before, the `appearance` row collapsed everything into one cell ending in `Type Default 'accent'`. After, it renders as a proper table:

| Property | Attribute | Description | Type | Default |
| --- | --- | --- | --- | --- |
| `appearance` | `appearance` | The button's visual appearance. | `'accent' \| 'filled' \| 'outlined' \| 'filled-outlined' \| 'plain'` | `'accent'` |

### DIFF
--- a/packages/webawesome/scripts/agent-skill.js
+++ b/packages/webawesome/scripts/agent-skill.js
@@ -150,9 +150,113 @@ function createTurndownService(baseUrl) {
 }
 
 /**
- * Processes rendered HTML from Eleventy output and converts it to clean Markdown.
+ * Renders a component's API table (Slots, Attributes & Properties, Methods, Events, CSS custom
+ * properties, Custom States, CSS parts) directly from the Custom Elements Manifest.
+ *
+ * The docs site builds these tables from the CEM, so scraping the rendered HTML is a lossy
+ * round-trip: per-row type, default, and description collapse into a single cell and any inline
+ * `<code>` after the first is dropped. Reading the CEM (as scripts/llms.js already does) keeps
+ * them accurate. Returns a markdown table string, or null when the section has no CEM data.
+ *
+ * The result is restored after the Turndown pass via a placeholder (see processHtmlToMarkdown),
+ * so it can contain real backticks, brackets, and `<…>` text without being escaped or parsed.
  */
-function processHtmlToMarkdown(htmlContent, baseUrl) {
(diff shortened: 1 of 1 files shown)


claviska:
Thanks for the PR! I merged it as part of https://github.com/shoelace-style/webawesome/pull/2531. We had previously made some changes that conflicted, but I liked your approach so I needed to reconcile, add a changelog, etc.

PR #2473, opened 2026-06-05
fix: scroll select typeahead matches into view

skyswordw:
## Summary
- scroll the current `<wa-select>` option into the listbox viewport when type-to-select changes the active option
- add a regression test for typing `ghana` to a below-the-fold option

Fixes #2472.

## Checks
- `npm ci`
- `npm run build`
- `CSR_ONLY=true npx web-test-runner --group select`
- `SSR_ONLY=true npx web-test-runner --group select`
- `npx prettier --check --log-level=warn --ignore-path="../../.prettierignore" src/components/select/select.ts src/components/select/select.test.ts`
- `git diff --check`

PR #2477, opened 2026-06-06
fix: scroll select typeahead matches into view

### LINKED ISSUE #2472: `<wa-select>`: scroll option into view when typing
### Discussed in https://github.com/shoelace-style/webawesome/discussions/2470

<div type='discussions-op-text'>

<sup>Originally posted by **patrickvuarnoz** June  5, 2026</sup>
When having a long list of options for a select and starting to type one of the options that is off-screen, it feels as if nothing is happening. When you then hit enter, it selects the option. It would be a better experience if the matched option could actually scroll into the visible area, i.e. if the dropdown menu would scroll automatically to show the option. 

You can try this in the following pen when typing "ghana":
https://codepen.io/patrickvuarnoz/full/ByQPWRG</div>

Dafenxz0:
Fixes #2472.

When typeahead finds an option while the select popup is open, the matched option is now scrolled into the visible listbox area. This makes long option lists feel responsive because typing a hidden match no longer leaves the visible menu unchanged.

Also adds a regression test with a long option list to verify the matched option becomes current and the listbox scrolls.

Validation:
- SKIP_SLOW_STEPS=true npm run build
- CSR_ONLY=true npx web-test-runner --group select

### DIFF
--- a/packages/webawesome/src/components/select/select.test.ts
+++ b/packages/webawesome/src/components/select/select.test.ts
@@ -622,6 +622,32 @@ describe('<wa-select>', () => {
           await el.updateComplete;
           expect(displayInput.getAttribute('aria-expanded')).to.equal('false');
         });
+
+        it('should scroll the matched option into view when typing', async () => {
+          const el = await fixture<WaSelect>(html`
+            <wa-select>
+              ${Array.from(
+                { length: 30 },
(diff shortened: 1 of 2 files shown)


claviska:
Thanks for the PR. Looks like this is a duplicate fix: https://github.com/shoelace-style/webawesome/pull/2473

PR #2491, opened 2026-06-10
Remove `--wa-accordion-divider-color`

lindsaym-fa:
Removes improperly scoped `--wa-accordion-divider-color` from `<wa-accordion-item>`.
Style customizations can be applied directly to the host element with `wa-accordion-item:not(:first-child) { ... }`


claviska (approved):
Good catch!

PR #2492, opened 2026-06-10
Scope `<th>` border to column headers only

lindsaym-fa:
Scopes the `border-block-end` styles previously applied to `<th>` elements to only those `<th>` elements that don't serve as row headers.

Additionally:
- Removes redundant `padding-block` from `th` selector (already applied in a prior rule)
- Replaces physical properties with logical properties
- Specifies `var(--wa-border-style)` for row borders instead of `solid`


talbs (approved):
👀 good. Getting 🎁🎁🎁! 

Confirmed locally with a separate row-based test (screenie below). Thanks for the fix!

Image

PR #2494, opened 2026-06-10
Use Logical Properties in Native Styles

talbs:
As a follow-up to https://github.com/shoelace-style/webawesome/pull/2492, this work audits `src/styles/native.css` and converts physical-axis CSS to logical properties so the wa-native layer adapts to RTL and vertical writing modes. 

I broke things down into 3 commits: 
1. clear fixes (`hr`, slider thumb, `progress`), 
2. `<pre>` `overflow-inline` plus a `:dir(rtl)` flip for the `<select>` caret
3. `width`/`height` → `inline-size`/`block-size` sweep across form controls, media, tables, and `<dialog>`.

☝️ That last item may be the one worth a confirmation.

A handful stayed physical on purpose: 

- `body { min-height: 100vh }`
- `border-width` shorthands
- `select` caret `background-position`(no logical keyword — `:dir(rtl)` handles it)
- `resize: vertical` on `<textarea>`.


talbs:
After talking with @lindsaym-fa, I'm closing this for now as its not an urgent need AND we should consider individual rules/cases in the future.

PR #2498, opened 2026-06-12
Add Missing `wa-prose` and `wa-flex-wrap` to Agent Design Skill

talbs:
This work wires `wa-prose` and `wa-flex-wrap` into the auto-generated `webawesome` agent skill.

Both shipped utilities had their own docs pages  but weren't being pulled into the skill's references, so agents using the skill couldn't discover them.


claviska (approved):
👍

PR #2499, opened 2026-06-12
fix(carousel): Pagination block height fix

### LINKED ISSUE #2495: Carousel pagination block height doesn't contain its zoomed dot
### Describe the bug
I am combining the Carousel with a Dialog for a cheap Lightbox setup and it works pretty well with a few overrides. One thing I noticed is that the pagination indicates the current dot by scaling it but its container then doesn't contain it properly. This was causing my dialog to indicate a scroll was needed and the bottom part of the dot was cut off. This can be fixed by adding some padding to the pagination block.

### Screenshots
Image

### Browser / OS
 - OS: Mac
 - Browser: Safari (As far as I tested)
 - Browser version: 26.5

### Additional information
Can be fixed with something like this:

```
wa-carousel &::part(pagination) {
  padding-block: var(--wa-space-2xs);
}
```

DanielKanyo:
## Summary

This PR fixes the height of the pagination container in the carousel component to ensure that the active pagination item is always fully visible.

## Changes
Added a small amount of vertical padding to the pagination container:
```
padding-block: var(--wa-space-3xs);
```

## Impact
Prevents the active pagination item from being cropped.

Fixes #2495

### DIFF
--- a/packages/webawesome/src/components/carousel/carousel.styles.ts
+++ b/packages/webawesome/src/components/carousel/carousel.styles.ts
@@ -29,6 +29,7 @@ export default css`
     flex-wrap: wrap;
     justify-content: center;
     gap: var(--wa-space-s);
+    padding-block: var(--wa-space-3xs);
   }
 
   .slides {


claviska:
Thanks for the PR! Merged in #2532.

PR #2501, opened 2026-06-12
Konnorrogers/fix ssr and hydration mismatches

KonnorRogers:
- Fixes SSR when not using the app wrapper
- Adds a playwright script for detecting hydration errors in docs pages
- Accompanying Pro PR which does the same.


packages/webawesome/docs/_includes/head.njk
```diff
@@ -55,6 +55,8 @@
     <script src="/assets/scripts/hydration-errors.js"></script>
     <link rel="stylesheet" href="/assets/styles/hydration-errors.css">
   {% endif %}
+
+  <meta content="{{ isDev }}">
```
claviska: Isn't `isDev` a boolean? Should this be `<meta content="{{ isDev }}" name="isDev">` or something?

packages/webawesome/docs/_includes/head.njk
```diff
@@ -90,18 +92,44 @@
...
     const kitCode = "f4ea8aaaeb9d4f12"
 
+    const define = window.customElements.define
+    window.customElements.define = function (tagName, ctor, options) {
+      if (window.customElements.get(tagName)) {
+        return
+      }
+
+      return define.call(this, tagName, ctor, options)
+    }
```
claviska, line 101: Are we doing this to suppress duplicate registration warnings? Is it necessary for SSR? It's probably worth a quick comment so this doesn't get stripped out in the future.

packages/webawesome/src/styles/component/host.styles.ts
```diff
@@ -14,4 +14,8 @@ export default css`
   [hidden] {
     display: none !important;
   }
+
+  :host([hidden]) {
+    display: none !important;
+  }
```
claviska: To make sure this applies to the host element and any child within the host element, we should combine these:

```suggestion
:host([hidden]),
[hidden] {
   display: none !important;
 }
```

packages/webawesome/docs/_includes/head.njk
KonnorRogers: it should be removed 😓 was left over from debugging.

packages/webawesome/docs/_includes/head.njk
KonnorRogers: comment added. but yes, its to prevent duplicate registrations of dependencies of pro components.

PR #2505, opened 2026-06-15
Unify Page Shell, Footer, and Sidebar Styling

talbs:
Alongside some companion PRs, this work revises some page and nav-level aspects of webawesome.com...

### Consistent Page Width
Tightens up the docs page shell so the body cluster, chrome rows, and footer all share one `--page-max-inline-size` (90rem). 

### Page Sizing Model
Pulls `wa-prose` off `<main>` and onto an inner `<div class="wa-prose">` so layout sizing and content typography stop fighting. 

`.content-container`'s default cap moves to `var(--page-max-inline-size)` so main, footer, and chrome bands line up at the page envelope. 

Locks `wa-page::part(body)` with `inline-size: 100%` so the grid has a definite size, and main fills its track even when `wa-prose` constrains intrinsic content. 

Centers the prose wrapper inside `main`.

### Layout Flags
`proseContent` → `hasProseContent` (matches the `is`/`has` convention). Full-bleed pages (`hasFramedMain: false`) now auto-skip the `wa-prose`
wrapper even if `hasProseContent` is inherited from the directory tree. Retired `hasWideMain` — class was inert; body-class injection + CSS rule
are gone, frontmatter entries scrubbed. New `has-prose-content` and `has-outline` body classes for CSS to key off.

### Sidebar Column Chrome
Moved `overflow: auto`, `max-block-size: 100%`, `padding-block: 2xl`, and the trailing border out of the old `min-width: 73.75rem` media query and
onto `wa-page:not([view='mobile'])`. Sidebar chrome is consistent at every desktop width — not just past the original breakpoint.

### Docs Main + Footer Padding Tiers
Main padding now scales with which chrome is visible: drawer (`xl xl`) → sidebar in column (`xl 4xl`) → sidebar plus outline aside (`3xl 4xl`). Docs footer drops its desktop bump and stays at `var(--wa-space-xl)` everywhere. On prose pages without an outline, we reserve `--aside-width` so `main` keeps the same width as outline-bearing pages.

### Full-Width Docs Index Pages
`hasProseContent: false` (and `hasOutline: false` where needed) on the tokens index, frameworks, themes, color-palettes, the utilities index, and
visual-tests so their grid content can breathe.

### Outline UI Overhaul
Lifted the expandable outline out of the `wa-prose` wrapper so prose typography stops bleeding into nav chrome, and modernized `outline.css` with logical properties, `--wa-space-*` tokens, and a mobile-first `min-width: 75rem` media query to match the docs convention. 

Swapped the bare `<details>` for `<wa-details>` with its default outlined appearance. 

Added `aria-label` to both outline navs, promoted the sidebar `<div>` to a `<nav>` with its own label, and paired `.current` with `aria-current="location"` so the active link reaches assistive tech. 

The expandable outline `<ul>` gets `wa-grid wa-gap-xs wa-list-plain` via a per-nav `linkIcon` / `listClass` option on `outlineTransformer`, with `--min-column-size: 25ch` so long headings stay on one line. Each link leads with a `variant="light"` hashtag `<wa-icon>` clustered next to the text via `wa-cluster wa-gap-xs wa-flex-nowrap` on the `<li>`. The `<wa-details>` summary drops to `--wa-font-size-s` so the chrome reads the same size as the links.


talbs:
@claude, review this PR.

claviska:
I love the consistency!

On pages with an outline the max length of content is about 655px, which feels narrow. Is there redundant padding here, or is this intentional?

CleanShot 2026-06-17 at 13 33 23@2x

---

I thought we resolved to keep the same content length as pages that have an outline. This seems quite wide and the palettes seem really tall as a result. Just making sure this is intentional.

If so, what is the purpose of the full page without outline? Like when would we choose one over the other and are there any rules to follow?

CleanShot 2026-06-17 at 13 35 28@2x

---

Possibly unrelated, but should we tighten up padding on mobile? Definitely in the header, but what about in the body as well?

CleanShot 2026-06-17 at 13 39 38@2x

talbs:
> I thought we resolved to keep the same content length as pages that have an outline. This seems quite wide and the palettes seem really tall as a result. Just making sure this is intentional.
> 
> If so, what is the purpose of the full page without outline? Like when would we choose one over the other and are there any rules to follow?

Yeah, we did resolve to in general. There were a few cases when applying this standardized width where things felt like they needed the full-width. Here are a couple of comparisons of the full vs. the standardized width...

| Full | Standardized |
|--------|--------|
| image | image | 
| image | image | 
| image | image | 

I'm happy to roll either way on this - seeing these compared, do you have a preference?

talbs:
> Possibly unrelated, but should we tighten up padding on mobile? Definitely in the header, but what about in the body as well?

Yep, I'll see what I can do here.

claviska:
@talbs Seeing them side by side, the narrower content looks better to me. If the redundant padding I mentioned earlier gets removed, it won't look so cramped. Text that flows that long (like in the wide variation) is a design smell, imo 👃

packages/webawesome/docs/assets/styles/outline.css
```diff
@@ -1,50 +1,78 @@
+:root {
+  --outline-aside-width: 15rem;
+}
```
lindsaym-fa: Could you explain the benefit of using a custom property for this? Look like we only use this value once, and it doesn't seem like something we'd want to encourage setting on the fly. I could be missing something, though!

talbs: @lindsaym-fa, super fair point — there's no real benefit. It's used once, and I don't have a case in mind for overriding it; I introduced it during the `outline.css` cleanup with vague future-theming intent. 

Dropping the `:root` declaration and inlining `15rem` on `#outline-standard`.

talbs:
Okay, @claviska and @lindsaym-fa, this (and the other PRs) has all of the recent feedback addressed.

Let me know how things are looking now.

claviska (dismissed):
:shipit:

lindsaym-fa (dismissed):
The conflict resolution goblins are knocking, but approved otherwise!

PR #2512, opened 2026-06-17
fix(tooltip): use relatedTarget instead of :hover matching in handleMouseOut

### LINKED ISSUE #2507: Wa-Tooltip closes in Chrome browsers when children are hovered
### Describe the bug
When creating a Wa-Tooltip and slotting a child node inside the tooltip body, the tooltip closes when hovering the child. Slotting simple text does not close the tooltip, it has to be another HtmlElement.

### To Reproduce
1. Open a Chrome Browser (In Firefox and Edge, the Tooltip stays open in both cases => The children can be hovered/selected etc.)
2. Go onto the official documentation page of the Wa-Tooltip: https://webawesome.com/docs/components/tooltip/
3. Try the "HTML in Tooltips" example.

Image

4. Try to hover over the tooltip content.
5. See errorous behaviour

### Browser / OS
 - OS: Windows 11 Version 25H2 (Build 26200.8655)
 - Browser: Chrome
 - Browser version: 149.0.7827.103 (Official Build) (64-bit) (cohort: Stable)

justinlaf:
Fixes #2507

### Root cause
`handleMouseOut` determined whether to keep the tooltip open by checking `this.anchor?.matches(':hover')` and `this.matches(':hover')` after the mouseout event fired. In Chrome, when the pointer moves from the tooltip host onto a slotted child element, the host's `:hover` pseudo-class can briefly evaluate to false during that transition, causing the tooltip to incorrectly close.

### Fix
`handleMouseOut` now reads `event.relatedTarget` — the element the pointer is moving to, which browsers provide reliably on `mouseout` — and checks whether that target is contained within the anchor or the tooltip using `Node.contains()`. This avoids the timing-dependent `:hover` matching issue entirely.

### Testing
Added a new test case in `tooltip.test.ts` that opens the tooltip, dispatches a `mouseout` event with `relatedTarget` set to a slotted child link, and asserts the tooltip remains open, directly reproducing the reported bug.

### Steps to test
1. Open the tooltip docs page: https://webawesome.com/docs/components/tooltip/
2. Try the "HTML in Tooltips" example in Chrome
3. Hover over the tooltip's child content — it should no longer close

### DIFF
--- a/packages/webawesome/src/components/tooltip/tooltip.test.ts
+++ b/packages/webawesome/src/components/tooltip/tooltip.test.ts
@@ -441,6 +441,30 @@ describe('<wa-tooltip>', () => {
       await waitUntil(() => tooltip.open);
       expect(tooltip.open).to.be.true;
     });
+
+    it('should remain open when the pointer moves onto a slotted child element of the tooltip', async () => {
+      const el = await fixtures[0]<HTMLDivElement>(html`
+        <div>
+          <wa-button id="hover-child-btn">Hover me</wa-button>
+          <wa-tooltip for="hover-child-btn" trigger="hover" show-delay="0" hide-delay="0">
+            <a href="#" id="tooltip-link">A link inside the tooltip</a>
+          </wa-tooltip>
+        </div>
+      `);
+      const tooltip = el.querySelector<WaTooltip>('wa-tooltip')!;
+      const childLink = el.querySelector<HTMLElement>('#tooltip-link')!;
+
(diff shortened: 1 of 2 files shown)


claviska:
Thanks for the PR! This looks good to me. I'm going to merge this as is because the fix is solid, then I'll add a change log entry and a pointer-driven test in a separate PR. https://github.com/shoelace-style/webawesome/pull/2516

claviska (approved):

PR #2513, opened 2026-06-17
Wa-random-content Component

kelseythejackson:
One of our [stretch goals](https://github.com/shoelace-style/webawesome/issues/1076) is a random content component. It's a utility component that randomly selects a child element and displays it. One of its main uses is to maintain engagement through variety, keep static pages somewhat fresh, or promote different offerings without cluttering the screen.

##  Attributes

| Attribute | Type | Default | Reflects | Description |
|----------|------|---------|----------|-------------|
| `items` | `number` | `1` | No | Number of children to show simultaneously. Clamped to `[1, childCount]`. |
| `mode` | `string` | `'random'` | Yes | Selection strategy (see Modes below). |
| `animation`|`string`|`none`|Yes|Entrance animation for newly shown children.|

## Methods

| Method | Returns | Description |
|--------|---------|-------------|
| `randomize()` | `void` | Imperatively trigger a new selection using the current `mode`. |

## Modes

**`random`** *(default)*
Picks children at complete random on every selection. No memory of prior picks.

**`unique`**
Same as `random`, but excludes the previously shown children from the group. Falls back to including them only when the remaining group is smaller than `items`. Useful for timed auto-play loops that avoid immediately repeating content.

**`sequence`**
Advances through children in their original DOM order, wrapping at the end. Each `randomize()` call advances the cursor by `items` positions.


packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -33,14 +33,13 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 
 :::added
 
+- Added the `<wa-random-content>` experimental pro component
```
claviska:
```suggestion
- Added the experimental `<wa-random-content>` component
```

packages/webawesome/src/components/random-content/random-content.ts
```diff
@@ -0,0 +1,125 @@
...
+      htmlEl.style.display = isSelected ? '' : 'none';
+      // Strip bottom margin from the last shown element so hidden siblings don't
+      // leave a phantom gap (e.g. <p> elements that aren't the last DOM child).
+      htmlEl.style.marginBlockEnd = isSelected && selected[selected.length - 1] === el ? '0' : '';
+    });
+
+    if (this.animation !== 'none') {
+      const styles = getComputedStyle(this);
+      const raw = styles.getPropertyValue('--animation-duration').trim();
+      const duration = raw.endsWith('s') && !raw.endsWith('ms') ? parseFloat(raw) * 1000 : parseFloat(raw) || 300;
```
claviska: We have a `parseDuration()` function in `src/internal/animate.ts` that should work for this.

packages/webawesome/src/components/random-content/random-content.ts
```diff
@@ -0,0 +1,125 @@
...
+      const htmlEl = el as HTMLElement;
+      const isSelected = selected.includes(el);
+      htmlEl.style.display = isSelected ? '' : 'none';
+      // Strip bottom margin from the last shown element so hidden siblings don't
+      // leave a phantom gap (e.g. <p> elements that aren't the last DOM child).
+      htmlEl.style.marginBlockEnd = isSelected && selected[selected.length - 1] === el ? '0' : '';
+    });
+
+    if (this.animation !== 'none') {
+      const styles = getComputedStyle(this);
```
claviska: This isn't a major problem since it only happens on randomize, but we could probably replace this logic with CSS classes and avoid layout thrashing altogether. (The fact that we're computing the values just to pass them to `.animate()` is a good hint we should let CSS do the work.)

packages/webawesome/src/components/random-content/random-content.ts
```diff
@@ -0,0 +1,125 @@
...
+import { customElement, property } from 'lit/decorators.js';
+import { watch } from '../../internal/watch.js';
+import WebAwesomeElement from '../../internal/webawesome-element.js';
+import styles from './random-content.styles.js';
+
+/**
+ * @summary Random content selects one or more child elements at random and displays them.
+ * @documentation https://webawesome.com/docs/components/random-content
+ * @status experimental
+ * @since 3.8
```
claviska:
```suggestion
 * @since 3.9
```

For the next release it should be 3.9.

claviska (commented):
Image

This is the first example at the top, which needs to drive the point of the component home. Let's show something a little more impressive. Maybe rotate through user cards or something like that. I'd definitely include the randomize button in this example since it's the first one the user will see.

---

> mode="unique" excludes the previously shown children from the candidate pool, so you never see the same item twice in a row. When the pool runs out, history resets and all children become eligible again.

Randomizing it enough times results in non-unique values, presumably when the pool runs out and it happens to guess the same number. This means it doesn't happen that often, but I'd still consider it a bug. As long as there's more than one child element, the next chosen one should always be different in this mode.

---

> Use setInterval to call randomize() on a timer for rotating content. Add animation="fade" for a smooth entrance transition. The duration defaults to 300ms and can be overridden with --animation-duration.

This is a cool feature, but it requires the user to break out of HTML and write the JS. It's out of the original spec's scope, but including a declarative API for this will make the component more powerful.

---

CleanShot 2026-06-17 at 12 56 33@2x

Can we add a button below this (like the animation examples) to demonstrate it better?

---

CleanShot 2026-06-17 at 13 00 48@2x

Because of its placement next to the badges, the button shifts around a bit when you click on it. I'd suggest moving it underneath like the other examples.

Badges are an interesting choice here. Is there anything better than 1, 2, 3 to use inside of them for this example?

---

Can we make the randomize buttons in all the examples normal size (medium) and `appearance="filled"` instead of the default?

CleanShot 2026-06-17 at 13 00 01@2x

---

CleanShot 2026-06-17 at 13 03 36@2x

This was a great idea! Are there other obvious animations worth adding? At the very minimum, we should show users an example of customizing it.

The examples take up a ton of vertical real estate to show basically the same thing. Can we consolidate these examples into one and add a `<select>` or radio group to choose the animation?

See my in-code note about using CSS classes instead of computed properties.

claviska:
Thanks for the updates. Some observations:

- Can we simplify the very first example? The profiles look great but maybe just show one at a time so the user clearly understands what the component does
- Unique mode is still showing dupes on randomize when the pool runs out (if you click the button 20 to 30 times you will almost definitely see it happen at least once)
- Did you have any luck with using CSS instead of compute + animate?


Please review my [previous comment](https://github.com/shoelace-style/webawesome/pull/2513/#pullrequestreview-4517919919). It's not clear from your commits/comments what's actually been addressed since then.

kelseythejackson:
@claviska, something happened with the merge in GitHub; my changes from last night are missing

claviska (commented):
### Main example

CleanShot 2026-06-24 at 09 35 49@2x

This first example got massive on tablet/desktop. It should be smaller and neatly organized inside of a card or similar.

### Reduced motion

I just noticed we're not honoring `pefers-reduced-motion` for the animation. There is prior art for this in other components that handle animations.

### Wrong number of items

In `mode="random"`, the `children.length > count` guard doesn't account for the current selection being excluded from the pool, meaning you'll sometimes see the wrong number of items.

For example, this example frequently shows just one item instead of two.

```html
<wa-random-content id="repro" mode="random" items="2" class="wa-cluster">
  <wa-badge variant="brand">A</wa-badge>
  <wa-badge variant="success">B</wa-badge>
  <wa-badge variant="warning">C</wa-badge>
</wa-random-content>

<div class="wa-cluster" style="margin-top: var(--wa-space-m)">
  <wa-button 
    size="s" 
    appearance="filled" 
    onclick="repro.randomize()"
  >Randomize</wa-button>
</div>
```

Image

### Leaky listener

The `randomize()` function calls `a.cancel()` to stop the previous animation, but `cancel()` fires a `cancel` event, not `animationend`. This leaves the `{ once: true }` attached until a future animation completes.

Not a show stopper, but these shouldn't leak.

claviska (approved):
Great work! :shipit:

PR #2514, opened 2026-06-17
Checkbox Group

claviska:
Adds the `<wa-checkbox-group>` component as [previously discussed](https://github.com/shoelace-style/webawesome/discussions/947).

CleanShot 2026-06-17 at 12 47 58@2x

Unrelated fixes:

- Fixed a bug in the code examples logic that allowed the copy button to receive focus when collapsed
- Fixes button styles in the dialog example for consistency

Related: 

- #1076

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13732,11 +13732,14 @@
     },
     "packages/webawesome-pro": {
       "name": "@awesome.me/webawesome-pro",
-      "version": "3.7.0",
+      "version": "3.8.0",
       "dependencies": {
         "@ctrl/tinycolor": "4.1.0",
         "@floating-ui/dom": "^1.6.13",
         "@konnorr/qr-creator": "^1.0.1",
+        "@lit-labs/ssr": "^4.1.0",
+        "@lit-labs/ssr-client": "^1.1.8",
+        "@lit/context": "^1.1.6",
         "@lit/react": "^1.0.8",
         "@shoelace-style/animations": "^1.2.0",
         "@shoelace-style/localize": "^3.2.2",
@@ -13751,6 +13754,7 @@
         "@wc-toolkit/cem-validator": "^1.0.3",
         "@wc-toolkit/jsx-types": "^1.3.0",
         "@web/dev-server-rollup": "^0.6.4",
+        "cookie-parser": "^1.4.7",
         "eleventy-plugin-git-commit-date": "^0.1.3",
         "esbuild": "^0.25.11",
         "gray-matter": "^4.0.3",
@@ -14165,6 +14169,17 @@
       ],
       "engines": {
         "node": ">=18"
+      }
+    },
+    "packages/webawesome-pro/node_modules/@lit-labs/ssr-client": {
+      "version": "1.1.8",
+      "resolved": "https://registry.npmjs.org/@lit-labs/ssr-client/-/ssr-client-1.1.8.tgz",
+      "integrity": "sha512-PjGh81oKsoI64m3IDjTqqjhC7dr2uC/o0jrllUb5gRAyp/RlAHxapgJrjq9kWz97faCHLQ8jUlTi6tGm+8fgyA==",
+      "license": "BSD-3-Clause",
+      "dependencies": {
+        "@lit/reactive-element": "^2.0.4",
+        "lit": "^3.1.2",
+        "lit-html": "^3.1.2"
       }
     },
     "packages/webawesome-pro/node_modules/@web/dev-server-core": {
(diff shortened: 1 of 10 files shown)


package-lock.json
lindsaym-fa, line 1: Was this change intentional?

packages/webawesome/docs/docs/components/checkbox-group.md
```diff
@@ -0,0 +1,117 @@
...
+
+```html {.example}
+<wa-checkbox-group label="Sizes" orientation="horizontal">
+  <wa-checkbox name="small">Small</wa-checkbox>
+  <wa-checkbox name="medium">Medium</wa-checkbox>
+  <wa-checkbox name="large">Large</wa-checkbox>
+</wa-checkbox-group>
+```
+
+### Sizes
```
lindsaym-fa, line 67: Would it be a real pain to support `size=""` on `<wa-checkbox-group>` directly, then propagate that to each child checkbox or switch? Otherwise, the label + hint aren't able to scale along with the checkboxes.

Here's an XL checkbox group alongside an XL radio group for reference:

Image

package-lock.json
claviska: nope, stale; reverted

packages/webawesome/docs/docs/components/checkbox-group.md
claviska: Done. Size works the same way as radio group now so the label and hint resize accordingly.

lindsaym-fa (approved):
Amazing!

packages/webawesome/docs/docs/components/checkbox-group.md
```diff
@@ -0,0 +1,131 @@
...
+<wa-checkbox-group label="Notifications" hint="Pick at least one channel.">
+  <wa-switch name="email">Email</wa-switch>
+  <wa-switch name="sms">SMS</wa-switch>
+  <wa-switch name="push">Push</wa-switch>
+</wa-checkbox-group>
+```
+
+### Required
+
+The `required` attribute adds a visual indicator to the group's label. Because each checkbox is an independent control, the checkbox group doesn't enforce the requirement. Set the `required` property on the checkbox or call its `setCustomValidity()` method to control validation.
```
brendon, line 121: Being able to indicate that at least one box in the group needs to be ticked would be advantageous (without needing to write that code externally). This is similar to how the `wa-radio-group` works.

In my stimulus controller emulation of this I set the `setCustomValidity` of the first checkbox in the list upon load (if the group is required). I then check if any checkboxes are ticked when any checkbox in the group changes state and either set or remove the `setCustomValidity` string:

```
    const valid = this.checkboxTargets.some(checkbox => checkbox.checked)

    this.checkboxTarget.setCustomValidity(valid ? '' : 'Select at least one tickbox')
```

brendon: Would you like me to open a new issue for this one?

claviska: How would this work? Are you suggesting the checkbox group look at its child elements to determine this automatically? I'm not sure I'm following.

brendon: I guess from a high level we're saying it'd be good to declare that at least one box in a checkbox group is ticked for the form to be valid, so the checkbox-group is 'required'.

I assume similar logic is used for the radio-group? Would the error message be attached to the first radio in the group?

Checking if at least one child checkbox is checked in the group would be required to make this work. Is that a problem in this context?

claviska: Thanks for clarifying. Radio group is unique because it's the controller and each radio item merely acts as an option. In this case, radio group handles all of the form validation logic, not radio item.

Checkbox group is inverted because each checkbox is actually its own separate form-associated element that handles its own validation. It wouldn't be possible for the checkbox group to handle validation without a substantial overhaul.

For example, what if you had a checkbox in a checkbox group and called `checkbox.setCustomValidity('Error')`? The checkbox group would have no way of knowing about that change. (It's possible but, at a glance, it would require substantial changes.)

The "verify at least one checkbox" logic would conflict with custom validation too. What if TWO checkboxes are required? What if some of the available options are mutually exclusive?

It seems better to let users handle custom validation with checkboxes.

brendon: Thanks for the clarification there @claviska, that does make sense, and I was cheekily ignoring other types of validation cases :D Having the checkbox-group wrapper will be a big gain in my case anyway as I can do away with the other wrapping I was already doing and hang our controller off the group element. Cheers :D

Issue #2510, opened 2026-06-17
`<wa-button-group>` does not work with native `<button>` children


Ramblurr:
### Describe the bug

`<button>`s inside a `<wa-button-group>` are not rendering like they should.

### To Reproduce

```html
<wa-button-group label="Alignment">
  <button class="wa-filled">Left</button>
  <button class="wa-filled">Center</button>
  <button class="wa-filled">Right</button>
</wa-button-group>
```

### Demo

https://codepen.io/editor/Ramblurr/pen/019ed466-bbe6-747e-9170-505d5c3325bc

### Screenshots

Actual:

<img width="249" height="77" alt="Image" src="https://github.com/user-attachments/assets/29a56d84-5061-4dcc-b163-c7e25d395eb8" />

Expected:

<img width="243" height="82" alt="Image" src="https://github.com/user-attachments/assets/e06b9af1-df1a-42d4-980f-f0d9ed7820d7" />

### Browser / OS
 - OS:  linux
 - Browser: chrome and firefox
 - Browser version: latest stable

claviska:
We intended button group to work with `<wa-button>` elements. What makes this tricky is that we ship Native Styles as an optional stylesheet, so I'm not sure we can safely support setting opinionated border radii and border colors for native buttons that may not include native styles.

@lindsaym-fa maybe we can support this by baking the styles into `native.css`...feels a bit awkward I'll admit but I'm curious what you think about this idea.

lindsaym-fa:
Because all of the button group border radii and margins are handled with internal custom properties, it's straightforward enough to leverage those same custom properties in our styles for native buttons. Folks would get the expected styling if they're using `native.css`, and they wouldn't if not. That seems fair.

There isn't any `<wa-button-group>`-specific styling that we'd have to add to `native.css` to handle this (unless we want to add a `.wa-button-group` class, which is a different discussion that I'm less convinced of). Button groups are only responsible for defining the custom properties that the button styles consume.

claviska:
PR #2515

PR #2515, opened 2026-06-17
Support native buttons in `<wa-button-group>`

claviska:
As discussed in #2510:

- Added support for Native Style users to use native buttons inside `<wa-button-group>`.
- Added a visual test for button groups comparing WA + native buttons
- Added a section in the button group docs
- Updated tests
- Updated changelog

CleanShot 2026-06-17 at 14 30 49@2x


lindsaym-fa (approved):
Perfection. ✨

PR #2517, opened 2026-06-17
fix issue with style polyfill not checking if the property exists.

KonnorRogers:
Fixes SSR'ed progress-bar in an astro app im trialing out


claviska (dismissed):
Some comments would help future us understand why we're doing this :)

PR #2524, opened 2026-06-19
feat: add useDefault: true to string reflect properties to prevent attribute sprouting

### LINKED ISSUE #2488: Enable `useDefault` to prevent attribute sprouting for default values
For string attributes that reflect, we should be setting [`useDefault: true`](https://lit.dev/docs/components/properties/#use-default) per [Lit's best practices](https://lit.dev/docs/components/properties/#reflected-attributes).

This will alleviate problems with DOM morphing libraries, but comes at a tiny cost of default attributes no longer reflecting. For example, a normal button will currently sprout the `variant="m"` attribute:

```html
<wa-button>  =>  <wa-button variant="m">
```

As maintainers, this means we need to use a more verbose selector for targeting default values:

```css
/* Not a default value, stays the same */
wa-button[variant="s"] {
   ...
}

/* Default value, check both variant="m" and the absence of the attribute
wa-button:not([variant]),
wa-button[variant="m"] {
   ...
}
```

I'm slating this for the Web Awesome 4 milestone because this is a breaking change even though its impact on end users will likely be small.

justinlaf:
Fixes #2488

### Changes made
Added `useDefault: true` to all string `@property` decorators with `reflect: true` across 52 component files, following Lit's best practices for reflected attributes.

This prevents "attribute sprouting" where default values are written back to the DOM as attributes on initialization. For example, previously:
```html
<wa-button>  =>  <wa-button variant="neutral" appearance="accent" size="m">
```
After this change, default values are no longer reflected as attributes, which fixes compatibility issues with DOM morphing libraries (e.g. Alpine.js, Turbo, htmx).

Boolean and Number properties were intentionally left unchanged as `useDefault` only applies to string attributes.

### Note
This is a breaking change slated for the Web Awesome 4.0 milestone as noted in the issue. CSS selectors targeting default attribute values will need to be updated to also account for the absence of the attribute, as documented in the issue description.

### DIFF
--- a/packages/webawesome/src/components/accordion-item/accordion-item.ts
+++ b/packages/webawesome/src/components/accordion-item/accordion-item.ts
@@ -62,16 +62,16 @@ export default class WaAccordionItem extends WebAwesomeElement {
   @property({ type: Boolean, reflect: true }) disabled = false;
 
   /** @internal Set by the parent accordion to control the heading level of the trigger. */
-  @property({ attribute: 'heading-level', reflect: true }) headingLevel = '3';
+  @property({ attribute: 'heading-level', reflect: true, useDefault: true }) headingLevel = '3';
 
   /** @internal Set by the parent accordion to control the roving tab index. */
   @property({ type: Boolean, attribute: false }) isTabbable = true;
 
   /** @internal Set by the parent accordion to control icon placement. */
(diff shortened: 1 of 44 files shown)


claviska:
Closing do to https://github.com/shoelace-style/webawesome/issues/2488#issuecomment-4752877024

PR #2537, opened 2026-06-22
fix page stretching

### LINKED ISSUE #2528: <wa-page> slot="footer" always overflows the viewport by the footer's height
> **Remark:** This bug report is based on a deep analysis of an **AI agent** comparing outcome of using WA Pro v3.7.0 in one application and v3.9.0 in another application that has been created more recently. By sticking the latter to 3.7.0, the issue is gone.

# `<wa-page>` `slot="footer"` always overflows the viewport by the footer's height (regression of the 3.1.0 fix)

## Summary

On pages whose content is shorter than the viewport, a footer placed in the
`footer` slot of `<wa-page>` is stacked **below** the viewport instead of being
pinned to its bottom. The document scrolls by exactly the footer's height, even
when there is almost no content — a permanent, spurious scrollbar.

This is a **regression**. Per the changelog this exact issue was fixed in
**3.1.0** ("Fixed a few sizing bugs in `<wa-page>` and `slot="footer"` no longer
will always 'overflow' the container."). It is back as of **3.9.0**.

## Affected versions

- **Broken:** 3.9.0 (and at least one earlier 3.8.x — undocumented; the 3.8.0
  notes mention only a `navigation-toggle` slot fix, and 3.9.0 lists no `page`
  changes).
- **Last known good:** 3.7.0.

## Expected behaviour (3.7.0)

`wa-page` sizes the body region as `viewport − header − footer`, so the footer
sits at the bottom of the viewport (sticky footer) and the document does not
scroll until the actual content exceeds `viewport − header − footer`.

## Actual behaviour (3.9.0)

The footer height is **not** subtracted from the body region. The body fills
`viewport − header`, and the footer is laid out *below* it, so the document is
always `header + (viewport − header) + footer = viewport + footer` tall →
overflow equals the full footer height, independent of content.

## Minimal reproduction

```html
<wa-page>
  <header slot="header" style="padding:1rem">Brand</header>
  <main><p>Short content</p></main>
  <footer slot="footer" style="padding:1rem">© Footer</footer>
</wa-page>
```

Load in a viewport taller than the content (e.g. 1440×900). The document shows a
vertical scrollbar; scrolling reveals the footer sitting just below the fold.

## Measured evidence

Identical app, footer height 109px, viewport height 900px, only the WebAwesome
version changed (clean install + cleared bundler cache):

| content height | WA 3.7.0 — body / overflow | WA 3.9.0 — body / overflow |
|---|---|---|
| 50px  | 717px / **0** ✓        | 826px / **109** ✗ |
| 400px | 717px / **0** ✓        | 826px / **109** ✗ |
| 700px | 796px / 79 (genuinely taller than fit) | 826px / 109 |

On 3.7.0 the footer height (109px) is subtracted (`900 − 74 header − 109 footer
= 717` body); on 3.9.0 it is not (`900 − 74 = 826` body), so the footer overflows
by its full 109px regardless of how little content is present.

## Environment

- `@web.awesome.me/webawesome-pro` 3.9.0 (Pro, premium theme)
- Vue 3 SPA, `<wa-page>` as the app shell, footer in the `footer` slot
- Chromium, desktop viewport
- Global CSS only zeroes html/body margins (`html,body{min-height:100%;margin:0;padding:0}`)

## Notes

- No new attribute/API appears to be required in 3.9.0 (nothing documented).
- Workaround: move the footer out of `slot="footer"` into the default slot and
  build a sticky footer manually, or pin to 3.7.0.

KonnorRogers:
fixes #2528


packages/webawesome/src/components/page/page.styles.ts
```diff
@@ -226,14 +226,10 @@ export default css`
...
     z-index: 4;
-    height: min(
-      var(--main-height, 100dvh),
-      calc(100dvh - var(--header-top) - var(--banner-top) - var(--subheader-top))
-    );
-    max-height: min(
-      var(--main-height, 100dvh),
-      calc(100dvh - var(--header-top) - var(--banner-top) - var(--subheader-top))
-    );
+    min-height: 0;
```
claviska, line 229: If this was a critical part of the fix, maybe it should be tagged with a comment.

claviska (dismissed):

packages/webawesome/src/components/page/page.styles.ts
KonnorRogers: Shouldn't be critical, the critical part was the `align-self: stretch;` the min-height is just a nice to have.

PR #2541, opened 2026-06-23
Reorder and cleanup component reference pages

claviska:
This PR reorders component reference skill files, so the overview comes first, followed by the API and then examples. This allows agents to get to the API information without having to parse long examples, potentially saving short-circuiting agents a large number of tokens for every request.

The order is now:

- Component name + summary
- API docs (slots, properties, events, methods, etc.)
- Examples


talbs (approved):
Very nice! The 🤖 overlords will love this!

PR #2543, opened 2026-06-23
Align Segmented-Field Picker Trailing Icons

talbs:
The trailing icons in `<wa-time-input>` — the clock toggle and the clear button — didn't sit on the same vertical axis as `<wa-select>`'s chevron or `<wa-input>`'s end-slot icons. So a row of mixed form controls looked subtly misaligned down the right edge. 

This adds a shared `segmented-field.styles.ts` module that places the picker buttons on that shared axis while preserving their padded hit targets and focus rings, and wires `<wa-time-input>` up to it. 

`<wa-date-input>` adopts the same module in the companion https://github.com/shoelace-style/webawesome-pro/pull/245 PR.

### Adds Shared Trailing-Affordance Module
`styles/component/segmented-field.styles.ts` is imported by both segmented-field pickers (`<wa-time-input>` here, `<wa-date-input>` via the `$webawesome` alias in pro). 

- Instead of the previous per-glyph nudge, each button gets a fixed `inline-size` equal to its natural width (`1.75em` expand, `1.5em` clear) with the glyph centered, so the icon stays on the axis regardless of its intrinsic width — a custom-slotted `expand-icon` no longer drifts. `
- margin-inline-end: -0.25em` lets the button padding overhang the trailing edge rather than displacing the glyph, and the leading gap is expressed against `--wa-form-control-padding-inline` so the alignment holds across sizes and themes.

### Revises `<wa-time-input>`
- Adopts the shared module and drops its duplicated trailing-button rules. 
- The `start`/`end` decoration slots now use the standard `--wa-form-control-padding-inline` gap to match `<wa-input>` and `<wa-select>` (previously a tighter `0.25em`).

### Adds Alignment Guards
Adds tests that render each control next to `<wa-select>` and assert their expand/clear glyph centers match within a pixel: one for `<wa-time-input>`, and an anchor test pinning `<wa-select>` and `<wa-input>` to the same axis — so the reference the other guards compare against can't silently drift.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -40,6 +40,9 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 - Fixed a bug in `<wa-carousel>` with `loop` enabled that displayed the wrong slide, briefly flashing it on load, when the carousel was initialized inside a hidden container such as an inactive tab panel [issue:1163]
 - Fixed component API tables in the `webawesome` Agent Skill by generating them from the CEM instead of scraping the rendered HTML [issue:2475]
 - Fixed a bug in `<wa-toast-item>` where the documented `--padding` custom property was unused in component styles
+- Aligned the `start` and `end` slot region in `<wa-date-input>` and `<wa-time-input>` with `<wa-input>` and `<wa-select>`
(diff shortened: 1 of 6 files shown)


talbs:
@claude, review this PR.

claviska (approved):

PR #2544, opened 2026-06-23
Fix known date validation bugs

claviska:
This PR fixes some bugs in the experimental known date component we recently released:

- Validation was incorrectly being shown inline below the form control
- When an invalid month, day, or year was entered, the validation focus would always point to the first input

### How to test

It's probably easiest to reproduce the buggy behavior first. To do that, go here: https://webawesome.com/docs/components/known-date

Then:

1. Enter invalid and incomplete dates and you'll see the feedback appear below the inputs.

CleanShot 2026-06-23 at 15 31 58@2x

2. Go to the required example and submit the date with an invalid day.

CleanShot 2026-06-23 at 15 31 05@2x

Follow the same steps for testing in this branch. You should observe no inline error for (1) and the invalid day should be correctly focused on submit for (2).

### DIFF
--- a/packages/webawesome/docs/docs/components/known-date.md
+++ b/packages/webawesome/docs/docs/components/known-date.md
@@ -113,7 +113,7 @@ Constrain the accepted range with `min` and `max`. Values outside the range are 
 
 ### Required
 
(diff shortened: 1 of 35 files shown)


kelseythejackson (dismissed):
Looks good

PR #2545, opened 2026-06-23
Docs: Revise Page Metadata Details

talbs:
This work tightens head metadata in the docs framework after a single-page SEO audit flagged missing canonical tags, an `og:url` that leaked the internal template file path into social shares, no `twitter:card` declaration, and no `og:image` dimensions. 

The standalone webawesome docs site and any downstream consumer of `docs/.eleventy.js` (including `webawesome-app`) all pick this up automatically.

### Fix `og:url`
Two bugs in the `ogUrl` data function in `.eleventy.js` collapsed into one rewrite:

- A page declaring `permalink: /foo.njk` had its template extension leak into `page.url`, which the data function passed straight into `og:url`. Worst case was `/account/signup.njk` showing up in social shares and pointing at a 404. The function now strips `.njk` before composing the URL.
- Front-matter overrides like `ogUrl: /signup` were emitted verbatim, producing relative-href `og:url` and `canonical` values. The function now always returns an absolute URL, prefixing `siteMetadata.url` when the value isn't already `http(s)://`.

### Add Canonical Link
`_includes/head.njk` now emits `<link rel="canonical" href="{{ ogUrl }}" />` directly after `<title>`, outside the `{% if not noindex %}` guard. Canonical is intentionally emitted on noindex pages too so duplicate-URL consolidation still works.

### Set `og:image` Dimensions
`siteMetadata` declares `imageWidth: 2400` / `imageHeight: 1260` matching the bundled default OG image. New `ogImageWidth` / `ogImageHeight` data functions emit those defaults when the page is using the default `ogImage`, allow explicit per-page overrides, and emit nothing when a page supplies a custom `ogImage` without dimensions — safer than lying with the defaults.

### Set `twitter:card`
`_includes/head.njk` declares `<meta name="twitter:card" content="summary_large_image">`. The bundled OG image is well above the 600×314 minimum for the large card variant.

---

Companion PR - https://github.com/shoelace-style/webawesome-app/pull/393.

### DIFF
--- a/packages/webawesome/docs/.eleventy.js
+++ b/packages/webawesome/docs/.eleventy.js
@@ -101,6 +101,8 @@ export default async function (eleventyConfig) {
     name: 'Web Awesome',
     description: 'Build better with Web Awesome, the open source library of web components from Font Awesome.',
     image: 'https://webawesome.com/assets/images/open-graph/default.png',
+    imageWidth: 2400,
+    imageHeight: 1260,
   };
 
   // Title composition/stripping config - single source of truth
@@ -146,6 +148,10 @@ export default async function (eleventyConfig) {
     ogTitle: data => composePageTitle(data.ogTitle || data.title),
     ogDescription: data => data.ogDescription || data.description,
     ogImage: data => data.ogImage || siteMetadata.image,
+    // Only emit dimensions when we know them: use the default if the page is using the
+    // default image, the explicit override if provided, otherwise null (suppresses emission).
+    ogImageWidth: data => data.ogImageWidth || (data.ogImage ? null : siteMetadata.imageWidth),
+    ogImageHeight: data => data.ogImageHeight || (data.ogImage ? null : siteMetadata.imageHeight),
     ogUrl: data => {
       // Strip template extensions: a page setting `permalink: /foo.njk` leaks the template
       // file path into page.url, which is never a valid public URL for og:url or canonical.
--- a/packages/webawesome/docs/.eleventy.js
+++ b/packages/webawesome/docs/.eleventy.js
@@ -153,8 +153,9 @@ export default async function (eleventyConfig) {
     ogImageWidth: data => data.ogImageWidth || (data.ogImage ? null : siteMetadata.imageWidth),
     ogImageHeight: data => data.ogImageHeight || (data.ogImage ? null : siteMetadata.imageHeight),
     ogUrl: data => {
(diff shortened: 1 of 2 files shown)


talbs:
@claude review this PR.

claviska (approved):

packages/webawesome/docs/.eleventy.js
```diff
@@ -146,10 +148,19 @@ export default async function (eleventyConfig) {
...
     ogImage: data => data.ogImage || siteMetadata.image,
+    // Only emit dimensions when we know them: use the default if the page is using the
+    // default image, the explicit override if provided, otherwise null (suppresses emission).
+    ogImageWidth: data => data.ogImageWidth || (data.ogImage ? null : siteMetadata.imageWidth),
+    ogImageHeight: data => data.ogImageHeight || (data.ogImage ? null : siteMetadata.imageHeight),
     ogUrl: data => {
-      if (data.ogUrl) return data.ogUrl;
-      const url = data.page?.url || '';
-      return url ? `${siteMetadata.url}${url}` : siteMetadata.url;
+      // Strip template extensions: a page setting `permalink: /foo.njk` leaks the template
```
claviska: Are we sure about this claim? If you set a permalink with an extension and visit that URL, will the page render or not?

claviska (dismissed):

packages/webawesome/docs/.eleventy.js
claviska: I guess the question is, should the `.njk` ever appear in a permalink? If not, would removing them help?

talbs: Fair — "template file path" is the misleading bit. 

The `.njk` permalinks don't exist in this repo's docs at all (zero), so the strip is a no-op here. They live in `webawesome-app`, which sets `permalink: /foo.njk` on ~35 pages on purpose: it's load-bearing for the two-pass SSR, where the page-router resolves a clean URL like `/account/claim` to the compiled `.njk` file. The clean URL is what's served; `page.url` just carries the `.njk` along.

So `.njk` *should* be in those permalinks, and removing them isn't a safe cleanup — they're in the other repo and the runtime resolution depends on them. The shared data function can't reach them anyway, so stripping at that seam is the right spot. And whether the `.njk` URL renders or not, canonical/og:url should point at the clean URL — that's the duplicate it exists to consolidate.

I'll reword the code comment to call the `.njk` an SSR-resolution artifact rather than a "template file path."

PR #2547, opened 2026-06-24
Improve `<wa-random-content>`: Accessibility, API Consistency, and Docs

talbs:
### Live Preview
https://webawesome-git-talbs-rando-calrissian-font-awesome.vercel.app/docs/components/random-content

---

Builds on #2513, refining the experimental `<wa-random-content>` component before it ships in the following ways...

### API
- Default `mode` is now `unique` (was `random`), so repeated selections don't show the same item twice in a row — the right default for tip rotators and timed loops.
- Replaced the single `interval` number with `autoplay` (Boolean) + `autoplay-interval` (ms), matching `<wa-carousel>`. Toggling `autoplay` doubles as the pause hook.
- Added a `wa-content-change` event (`detail.items`), emitted on first render, on `randomize()`, and on each autoplay tick. Follows the library's `wa-<noun>-change` convention (there is no bare `wa-change`).
- `randomize()` now returns the elements it selected.
- Unselected children are now hidden with the `hidden` attribute instead of inline `display:none`, exposing a `:not([hidden])` styling hook for the current item. A `::slotted([hidden]) { display: none !important }` rule keeps hiding reliable even when a child sets its own `display` (e.g. `.wa-flank`, a component's `:host`).

### Accessibility
- Autoplay now uses carousel's `AutoplayController`, so it pauses on hover, focus, and touch.
- Reduced motion skips the entrance animation but keeps autoplay running, matching `<wa-carousel>`'s behavior (motion is removed, content still advances).
- Added a visually-hidden `role="status"` live region (the `copy-button`/`textarea` pattern) so screen readers hear rotations.
- Documented that autoplay needs a visible pause control for keyboard users, since hover/focus pausing can't help when the rotating content isn't focusable.

### Fixes
- Call `super.firstUpdated()` — the base class dispatches the initial `slotchange` there as an SSR workaround, so skipping it left server-rendered content unselected. Caught by the `ssr-client-hydrated` test variant.
- Strip the leading *and* trailing block margins of the shown set (previously only the trailing one), so a shown `<p>`/`<wa-callout>` doesn't leave a phantom gap above or below the component.

### Docs
Rewrote `random-content.md` to follow our conventions and approachable structure: 

- Title-Case task-oriented headings in a logical order
-  the `<wa-divider>` example-control convention from carousel
- asset-free examples
- a modes comparison table
- notes for SSR (`wa-cloak`) and framework usage. 

### Tests
54 tests across CSR and SSR covering the new default, the event, autoplay pause/reduced-motion, attribute-based hiding (including children with their own `display`), the symmetric margin handling, and `randomize()`'s return value.

### DIFF
--- a/packages/webawesome/docs/docs/components/random-content.md
+++ b/packages/webawesome/docs/docs/components/random-content.md
@@ -2,282 +2,314 @@
 title: Random Content
 layout: component
 category: Helpers
+synonyms:
+  - rotate
+  - shuffle
+  - variety
+  - rotator
+use-cases:
+  - rotating testimonials
+  - tip of the day
+  - featured content
 ---
 
-Randomly selects and displays one or more of its slotted children. Transparent to layout by default (`display: contents`), so it composes naturally in both block and inline contexts.
+Randomly picks and displays one or more of its slotted children, hiding the rest. Use it to rotate testimonials, surface featured content, show a tip of the day, or add variety to an otherwise static page.
 
 ```html {.example}
-<wa-random-content id="initial" class="wa-grid" items="1" mode="unique" animation="fade-right" style="--animation-easing: ease-in-out; --animation-duration: 500ms">
-  <article class="wa-stack">
-    <div class="wa-frame wa-border-radius-l">
-      <img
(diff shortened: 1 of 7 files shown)


talbs:
@claude, review this PR.

claviska (approved):
Outstanding.

PR #2551, opened 2026-06-25
Docs: Standardizing Example Controls

talbs:
This work standardizes how component docs present examples that pair a live demo with separate controls, rolling out the convention established in the `<wa-random-content>` PR (#2547): 

- a `<wa-divider>` separates the demo from the controls that drive it
- multi-control rows align with `wa-cluster`, and control buttons normalize to `appearance="filled"`.

### What changed

14 component docs, covering every example where a demo is driven by *separate* controls:

- `popup` (12 examples — switches/sliders/comboboxes configuring the popup), `carousel`, `accordion`, `animation`, `checkbox-group`, `format-bytes`, `format-number`, `progress-bar`, `progress-ring`, `qr-code`, `select` (lazy-loading), `split-panel`, `tooltip`, `tree`.
- Multi-control rows use a bare `wa-cluster` (it centers by default — no inline alignment needed). Single controls sit directly after the divider.
- A few examples whose scripts located controls via `nextElementSibling` traversal were switched to scoped queries, since the new wrapper changes sibling order. Behavior is unchanged.

### What's intentionally excluded

The divider separates a *visible demo* from its controls, so examples without that shape are left alone: intrinsic triggers (`dialog`/`drawer`/`toast` "Open"/"Show" buttons with nothing persistent above them), and pages where the interactive element *is* the demo (`button`, `switch`, `select`, `slider`, etc.). `animation` keeps its divider but not `wa-cluster` — its controls have a deliberate fixed-width column layout.

## Companion PRs

-  https://github.com/shoelace-style/webawesome-pro/pull/246 — same convention across `chart`, `file-input`, and `video-playlist` docs (branch `talbs/example-control-sweep`).

### DIFF
--- a/packages/webawesome/docs/docs/components/accordion.md
+++ b/packages/webawesome/docs/docs/components/accordion.md
@@ -270,17 +270,19 @@ To place HTML in an accordion item's header, use the `label` slot instead of the
 Use the `expandAll()` and `collapseAll()` methods to programmatically control all items at once. Note that `expandAll()` is a no-op when `mode` is `single` or `single-collapsible`.
 
 ```html {.example}
-<wa-accordion id="accordion-methods">
-  <wa-accordion-item label="Section one">Content for the first section.</wa-accordion-item>
-  <wa-accordion-item label="Section two">Content for the second section.</wa-accordion-item>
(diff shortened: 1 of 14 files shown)


claviska (approved):
This PR left me...divided 🙈

PR #2554, opened 2026-06-25
Docs: Add  Code Examples Color-Scheme + Text-Direction Toggles

talbs:
Each docs code example now has its own light/dark and LTR/RTL toggles in the example toolbar, so you can preview any example in dark mode or RTL without flipping the whole page.

| Default | Toggled |
|--------|--------|
| image | image |
| image | image | 
| image | image | 

It builds on the existing `.code-example` chrome rather than adding anything new: two buttons in the toolbar, a couple of branches in the client script, and some CSS. No component changes, and the pro docs inherit it automatically through the 11ty config chain.

### Color Scheme
- Toggling repaints the whole preview surface — background and the components on it — while the toolbar, source view, and resizer stay in the page's theme. 
- It applies an explicit `wa-light`/`wa-dark` to the preview, so it overrides the page either way (works whether the docs are in light or dark mode). 
- Floating content like dialogs and dropdowns inherits it too, since top-layer rendering doesn't change CSS inheritance. 
- The resizer keeps the page colors by reading custom properties captured above the themed element.

### Direction
- Toggling stamps `dir` on the preview content and every descendant. 
- That breadth is necessary: our components resolve direction from their own `dir` (or the document's), never an ancestor's, so a wrapper-only attribute would flip the text but not the components. 
- A `MutationObserver` keeps dynamically-added content (like the carousel "add slide" demo) in sync while RTL is active.

### Opting Out
- Use `no-theme` / `no-dir` to suppress a toggle where it'd be redundant — already applied to the theming and RTL demos. - Examples that embed an iframe opt out automatically, since a scoped theme or direction can't reach into a separate document.

### Notes
- The toggles are per-example and don't persist. 
- They use an action-label pattern (the icon's label reads "Show in dark mode", etc.) rather than `aria-pressed`. 
- And a chart drawn to `<canvas>` won't re-color on toggle, since it's already painted.


talbs:
@claude, review this PR.

claviska:
Great addition!

CleanShot 2026-06-25 at 16 06 36@2x
<br>
CleanShot 2026-06-25 at 16 07 28@2x

I respectfully request to use the same light/dark mode icons as we use on the website.

claviska:
Total nit: when you toggle light/dark mode in a specific example and then toggle the main light/dark mode for the website, I think it should cancel out the selection for all example and sync them to the main mode. (Otherwise there's no way to reset them outside of refreshing.) It feels pretty natural to have it absorb back into the main theme.

talbs:
Okay, @claviska - this has all of your solid feedback/requests tackled. Mind giving it another look?

claviska (approved):
💥

PR #2561, opened 2026-06-26
Fix XML tags in SKILL.md YAML description for Claude.ai compatibility

### LINKED ISSUE #2558: Claude doesn't like XML tags in SKILL.md
[From Discord:](https://discord.com/channels/739437527907303535/1020332605599662181/1519773409376993532)

---

I am just starting to experiment with the Web Awesome AI skills. I ran into one issue right away. When I tried to upload the `webawesome-design` skill in Claude.ai > Customize > Skills, I got an error:

SKILL.md description cannot contain XML tags

The problem turned out to be the use of angle brackets in mentions of `<wa-tag>` in the SKILL.md YAML description. Apparently the Claude.ai validator didn't like that. The solution is to remove the angle brackets and just use `wa-tag` in the skill description. Note the angle brackets are only an issue in the YAML description not in the body.

This is probably not an issue for using the skill locally in Claude Code. But it is an issue for using the skill in Claude.ai Projects and possibly in Claude Design. Here are the details in a Claude conversation:

https://claude.ai/share/860bc4b3-bbe0-4935-ba69-802b2f53af1c

WaterXiao-git:
## Description

Claude.ai's skill validator rejects SKILL.md files whose YAML description contains XML/HTML angle brackets (e.g. \<wa-page>\). This prevents the webawesome-design skill from being used in Claude.ai Projects and Claude Design.

## Changes

- Removed angle brackets from \<wa-page>\ references in the SKILL.md YAML description field
- The component name \wa-page\ remains unambiguous without brackets

## Related Issue

Closes #2558

### DIFF
--- a/packages/webawesome/scripts/design-skill/SKILL.md
+++ b/packages/webawesome/scripts/design-skill/SKILL.md
@@ -4,9 +4,9 @@ description: >
   Design and lay out user interfaces with Web Awesome. Use this when building or styling a PAGE,
   LAYOUT, or SECTION; choosing or customizing a THEME; applying brand COLORS or design tokens; or
   composing a polished, good-looking UI with Web Awesome components and utilities. Triggers on requests
-  like "build a landing page", "make an app layout", "set up <wa-page>", "add a sidebar", "apply a
+  like "build a landing page", "make an app layout", "set up wa-page", "add a sidebar", "apply a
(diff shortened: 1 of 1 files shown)


claviska (changes requested):
The file has a lot of places where these brackets exist. This only covers one tag.

PR #2564, opened 2026-06-26
Use `@scope` in `prose.css`

omnifides:
### Summary
Wraps `prose.css` rules in `@scope (.wa-prose) to (.wa-not-prose)` so that the classes are proximity aware. That way, we don't need to fight specificity or precedence wars to opt-in or opt-out of these styles.

Prose styles reliably apply to `wa-prose` elements, even when nested in `wa-not-prose` containers, and the styles are reliably ignored for `wa-not-prose` elements, even when nested in `wa-prose` containers.

---

### The Problem

There are two issues at play with our current approach for opting in and out of `prose.css` styles:

1. **Layer precedence:** The `wa-not-prose` class in `prose.css` specifies `revert-layer` for the properties styled otherwise in with `wa-prose`. However, because these rules are scoped to `@layer wa-utilities`, `wa-not-prose` reverts the properties across the entire `wa-utilities` layer. This renders many essential styles for other utilities — like `font-size` for text utility classes and `margin-block` resets for layout utility classes — inapplicable.

2. **Specificity:** While using `wa-not-prose` within a `wa-prose` container ignores the prose styles as expected, using `wa-prose` within a `wa-not-prose` container doesn't yield the same results. `wa-prose` always loses the specificity battle with `wa-not-prose`, so we can't opt back in to `wa-prose` styling.

### Example

Each of our docs uses `wa-prose`, but each of our code examples uses `wa-not-prose`. Because of this, the two problems described above rear their heads in different contexts:

1. Code examples that use text or layout utility classes don't get the intended font sizes or spacing from those classes. In first example below, note how the classes `wa-stack` and `wa-heading-l` don't apply all intended styles in the _Before_ image.
2. Code examples that are meant to demonstrate `wa-prose` can't actually show the intended `wa-prose` styles. In the second example below, note how `wa-prose`'s bespoke spacing isn't applied in the _Before_ image.

| Before | After |
| - | - |
| Screenshot 2026-06-26 at 4 45 53 PM | Screenshot 2026-06-26 at 4 45 45 PM |
| Screenshot 2026-06-26 at 6 07 57 PM | Screenshot 2026-06-26 at 6 07 51 PM |

---

### Browser Support

[`@scope` has longstanding support](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@scope#browser_compatibility) in Chrome, Edge, and Safari, though it was only enabled by default in Firefox less than a year ago.

Firefox is currently on version 152, and `@scope` support was enabled in version 146. That satisfies our "latest two major versions" requirement and feels comfortable for an opt-in, progressive enhancement–style feature like `prose.css`.

### DIFF
--- a/packages/webawesome/src/styles/utilities/prose.css
+++ b/packages/webawesome/src/styles/utilities/prose.css
@@ -6,7 +6,7 @@
      sits at 0,0,0 specificity — any utility class or element rule the author
      applies inside the container takes precedence without weight tricks. */
   @scope (.wa-prose) to (.wa-not-prose) {
-    :scope {
+    :where(:scope) {
       /* #region Tokens ~~~~~~~~~~~~ */
 
       /* Em-based rhythm scale: each token resolves at the use site, so margins
--- a/packages/webawesome/src/styles/utilities/prose.css
+++ b/packages/webawesome/src/styles/utilities/prose.css
@@ -5,7 +5,7 @@
      The class and its element selectors are wrapped in `:where()` so every rule
      sits at 0,0,0 specificity — any utility class or element rule the author
      applies inside the container takes precedence without weight tricks. */
-     
+
   /* @scope is Newly Available in Baseline 2025. As such, it should be reserved for
      non-essential, progressive enhancement features until @scope is established
      as Widely Available. https://caniuse.com/css-cascade-scope */
--- a/packages/webawesome/src/styles/utilities/prose.css
+++ b/packages/webawesome/src/styles/utilities/prose.css
@@ -1,168 +1,157 @@
-@layer wa-prose {
+@layer wa-utilities {
   /* `wa-prose` applies hierarchical, asymmetric typographic rhythm to a block
      of long-form content. Builds on `wa-native`'s base block styles instead of
      duplicating them. Em-based spacing scales when composed with `wa-font-size-*`.
      The class and its element selectors are wrapped in `:where()` so every rule
      sits at 0,0,0 specificity — any utility class or element rule the author
      applies inside the container takes precedence without weight tricks. */
-  :where(.wa-prose) {
-    /* #region Tokens ~~~~~~~~~~~~ */
+  @scope (.wa-prose) to (.wa-not-prose) {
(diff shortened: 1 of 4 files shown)


talbs:
Thank you so much for the fix and the continued work to make prose rosy (and less thorny).

claviska (dismissed):
Had to do some reading for this one 😅 

> Firefox is currently on version 152, and @scope support was enabled in version 146. That satisfies our "latest two major versions" requirement and feels comfortable for an opt-in, progressive enhancement–style feature like prose.css.

This alleviates my main concern, since this is Baseline 2025.

My only other concern is using this in non-PE ways. Now that `@scope` appears in the source, there's a precedent and it's easy for us (and contributors) to forget that it may be OK in some places and not in others. Is it worth a comment and a [link to caniuse](https://caniuse.com/css-cascade-scope)?

packages/webawesome/src/styles/utilities/prose.css
```diff
@@ -5,164 +5,153 @@
...
-    /* #endregion */
-
-    /* #region Headings ~~~~~~~~~~~~ */
-
-    /* Em sizes inherit through the cascade — re-applying `--wa-font-size-scale`
-       here would double-scale. Authors overriding should stay em-based. */
-    & :where(h1) {
-      font-size: 2.5625em;
+  @scope (.wa-prose) to (.wa-not-prose) {
+    :scope {
```
talbs:
```suggestion
    :where(:scope) {
```

The previous rule kept specificity at `0,0,0`. Can we use the above to maintain that (rather than `:scope {}` moving the specificity to `0,1,0`? 

If you'd rather not, we'll probably need to update the docs' specificity mention at https://webawesome.com/docs/utilities/prose#composing-with-other-utilities.

talbs (dismissed):
Looks good and thanks again for the fix.

I left one suggestion/comment around specificity. Feel free to take or leave it - but if you leave it, we should update the specificity note in Prose's docs.

talbs (approved):

PR #2568, opened 2026-06-29
fix `name` property typing on `wa-checkbox`

roydukkey:
The `name` property on `wa-checkbox` should accept strings. Currently, the property is only accepting `null`, as seen in the [generated typedefs](https://app.unpkg.com/@awesome.me/webawesome@3.9.0/files/dist/components/checkbox/checkbox.d.ts#L51-52).

```
    /** The name of the checkbox, submitted as a name/value pair with form data. */
    name: null;
```

### DIFF
--- a/packages/webawesome/src/components/checkbox/checkbox.ts
+++ b/packages/webawesome/src/components/checkbox/checkbox.ts
@@ -75,9 +75,6 @@ export default class WaCheckbox extends WebAwesomeFormAssociatedElement {
 
   @property() title = ''; // make reactive to pass through
 
-  /** The name of the checkbox, submitted as a name/value pair with form data. */
-  @property({ reflect: true }) name = null;
-
   private _value: string | null = this.getAttribute('value') ?? null;
 
   /** The value of the checkbox, submitted as a name/value pair with form data. */


packages/webawesome/src/components/checkbox/checkbox.ts
```diff
@@ -76,7 +76,7 @@ export default class WaCheckbox extends WebAwesomeFormAssociatedElement {
   @property() title = ''; // make reactive to pass through
 
   /** The name of the checkbox, submitted as a name/value pair with form data. */
-  @property({ reflect: true }) name = null;
+  @property({ reflect: true }) name: string | null = null;
```
claviska: Just a passing glance review, but I think this should be removed completely because `name` should be inherited from `WebAwesomeFormAssociatedElement`. Maybe @KonnorRogers can check my math.

KonnorRogers: Yup, double checked, that would be the way to do it as this is defined the exact same way there.

roydukkey:
Not sure if this warrants a callout in the changelog or not, but if anything I would focus on the fact that this corrects a typing issue for my team:

image

claviska:
@roydukkey We would typically put it in there as a bug fix. You're welcome to add it, if not, we'll add it when we merge :)

KonnorRogers:
Added note here:

https://github.com/shoelace-style/webawesome/pull/2572

PR #2573, opened 2026-06-30
Fix #2571: keep color-picker swatches the same size when only a few are present

### LINKED ISSUE #2571: `<wa-color-picker>` draws swatches too big when only a few are present
Repro:

```html
<wa-color-picker label="Select a color" swatches="red; green; blue;"></wa-color-picker>
```

Result:

Image

Expected behavior: swatches should be the same size as shown when the row is full, even when there is only one.

Image

dev48v:
## Summary

`<wa-color-picker>` drew its swatches much larger than normal when only a few were present.

Fixes #2571

## Cause

The `.swatches` grid uses:

```css
grid-template-columns: repeat(auto-fit, minmax(min(1.5em, 100%), 1fr));
```

`auto-fit` collapses the empty tracks and hands their space to the filled tracks (each able to grow to `1fr`), so with only a few swatches each one stretched to a large fraction of the container width.

## Fix

Switch `auto-fit` to `auto-fill`. `auto-fill` keeps the empty tracks, so filled swatches keep the same size whether the row is full or holds a single swatch — which is the expected behavior described in the issue. One-character CSS change, plus a changelog entry.

## Testing

- `npm run build` — passes
- `CI=true npx web-test-runner --group color-picker` — **51 passed** on both Chromium and Firefox
- Reproduced the issue case (`swatches="red; green; blue;"`): before the change, 3 swatches rendered ~3× larger than in a full row (75px vs 27px in a 240px container); after, they render at the same size as a full row.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -54,6 +54,7 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 - Fixed a bug in `<wa-known-date>` that showed validation errors while typing instead of on form submission like other form controls
 - Fixed a bug in `<wa-known-date>` where the validation tooltip always pointed to the first input regardless of which one was invalid
 - Fixed a bug in `<wa-toast-item>` where the documented `--padding` custom property was unused in component styles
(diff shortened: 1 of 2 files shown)


claviska (approved):
Outstanding. Thanks! 🙌

PR #2582, opened 2026-07-01
Replace Google Fonts URL syntax with Bunny Fonts syntax in themes

omnifides:
At some point, loading multiple fonts in a single `@import` using Google Fonts syntax via Bunny Fonts stopped working. The first font in the query string loads, but additional fonts do not.

This work replaces the Google Fonts syntax in the query strings for theme font `@import`s with Bunny Fonts' preferred syntax. For safety, this also replaces the single font URLs in our themer schema with matching syntax.


talbs (dismissed):
🐇 🥕🥕🥕🥕🥕🥕🥕

PR #2583, opened 2026-07-01
Fix Anchor-Link Scrolling for Sticky Header/Subheader

talbs:
Deep-linking to a heading in the docs (e.g. `/docs/resources/changelog#wa_3100`) often landed in the wrong place, with the target hidden behind the sticky header. This fixes the landing position and tidies up the surrounding scroll behavior.

### Landing on the Right Spot
On a cold load to a `#hash`, the page scrolled to the anchor before the sticky header had finished measuring its height, so the target ended up tucked behind it. We now wait until the header/subheader is measured, then align to the target. The offset itself already comes from `wa-page`.

### Smooth-Link Clicks
Simplified in-page link clicks to a plain `scrollIntoView()`, which also fixes a related bug where clicks didn't account for the subheader height.

### Outline Offset
Fixed the right-hand "On this page" outline (`outline.css`), whose sticky offset referenced an undefined variable and silently did nothing.

### Reduced Motion
Smooth scrolling now lives in CSS (`scroll-behavior` in the shared `utils.css`) instead of a hard-coded JS option, so it respects `prefers-reduced-motion` and themes can override it. The load and reload scroll corrections stay instant.

### Follow-up
The root cause lives in `<wa-page>`'s measurement timing, so it affects every `wa-page` consumer, not just our docs. We should consider making the fix to `<wa-page>` in the future.

### DIFF
--- a/packages/webawesome/docs/assets/scripts/scroll.js
+++ b/packages/webawesome/docs/assets/scripts/scroll.js
@@ -25,16 +25,12 @@ document.addEventListener('click', event => {
 
   // Only handle smooth scroll if there's a hash and the link points to the current page
   if (id && link.pathname === window.location.pathname) {
-    const target = document.getElementById(id);
-    const headerHeight = document.querySelector('wa-page > header').clientHeight;
+    const target = document.getElementById(decodeURIComponent(id));
 
     if (target) {
       event.preventDefault();
-      window.scroll({
-        top: target.offsetTop - headerHeight,
-        behavior: 'smooth',
-      });
-
+      // Offset comes from scroll-margin-top on wa-page descendants (library layers.css).
+      target.scrollIntoView({ behavior: 'smooth' });
       history.replaceState(history.state, '', `#${id}`);
     }
   }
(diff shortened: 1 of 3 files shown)


talbs:
@claude, review this PR.

packages/webawesome/docs/assets/scripts/scroll.js
```diff
@@ -49,15 +45,38 @@ window.addEventListener('scroll', updateScrollClass);
...
+      // Snap, never animate — this corrects a stale load position, not a user action.
+      target.scrollIntoView({ behavior: 'instant' });
+    } else {
+      requestAnimationFrame(tick);
+    }
+  };
+  requestAnimationFrame(tick);
+}
+
 // Restore scroll position after components are defined
```
KonnorRogers, line 68: This is a huge red flag to me. Its easier to just wait for things to settle.

packages/webawesome/docs/assets/scripts/scroll.js
```diff
@@ -49,15 +45,38 @@ window.addEventListener('scroll', updateScrollClass);
...
   const key = `wa-scroll-y-[${location.pathname}]`;
   const scrollY = sessionStorage.getItem(key);
+  const hashTarget = location.hash ? document.getElementById(decodeURIComponent(location.hash.slice(1))) : null;
 
-  // Only restore when reloading, otherwise clear it
-  if (navigationType === 'reload' && scrollY) {
-    window.scrollTo(0, scrollY);
+  if (hashTarget) {
+    // Re-align after hydration; the browser's initial hash jump landed at a stale position.
+    alignToHashTarget(hashTarget);
```
KonnorRogers, line 77: try:

```
setTimeout(() => {
  hashTarget?.scrollIntoView({ behavior: "instant" })
})
```

KonnorRogers (commented):
`scrollIntoView()` definitely seems more correct behavior. I'm not sure constantly polling via `rAf` is the way to go. We should be able to tell when everything has "rendered"

talbs:
Closing this - I think the root issue here (and this fix's direction) deserves a longer look.

PR #2584, opened 2026-07-02
<wa-otp-input> Component

kelseythejackson:
## Add `wa-otp-input` component

A form-associated OTP/passcode input that renders a fixed number of individual character segments.

### Architecture

Uses a **single hidden `<input>`** as the form and focus target, the usual accessibility pattern for OTP. Visual segment cells are decorative (`aria-hidden="true"`), with `delegatesFocus: true` on the shadow root, so clicking anywhere on the field routes focus to the real input. This avoids the semantic and screen reader problems that come with multi-input implementations.

### Properties

  | Property | Description |
  |---|---|
  | `length` | Number of segments (default: 6) |
  | `type` | Character filter: `numeric` (default), `alpha`, `alphanumeric` |
  | `format` | Group segments with separators, e.g. `"### ###"` or `"####-####"` |
  | `appearance` | `outlined` (default), `filled`, `filled-outlined`, `contained` |
  | `mask` | Display entered characters as bullets |
  | `case` | `preserve` (default), `upper`, `lower` |
  | `size` | `small`, `medium` (default), `large` |
  | `placeholder` | Hint character shown in empty segments (e.g. `·`, `0`) |
  | `readonly` | Display-only; still focusable and submitted |
  | `autofocus` | Passed through to the hidden input |
  | `autocomplete` | Defaults to `one-time-code` for SMS autofill |
  | `required` | Standard constraint validation |

### Behavior

  - **Keyboard**: Arrow keys and Tab move between segments; Backspace clears the current slot in-place (no character shift); Delete clears without moving the cursor
  - **Paste**: Fills from the active segment forward, preserving characters before and after the pasted region; non-matching characters (per `type`) are silently dropped
  - **`wa-complete`**: Fires once when the last segment is filled, useful for auto-submit flows
  - **Validation**: `required` and `tooShort` (partial fill is always invalid regardless of `required`) via `MirrorValidator`; `validationTarget` anchors the browser constraint popup to the visible segments
  - **Prop reactivity**: Changing `type`, `case`, `length`, or `format` at runtime re-filters and truncates the current value immediately

### Accessibility

  - `<label for="hidden-input">` + `aria-describedby="hint"` on the hidden input
  - `role="group"` + `aria-labelledby` on the segments container
  - `autocomplete="one-time-code"` default for OS-level SMS autofill
  - `inputmode="numeric"` on numeric type for appropriate mobile keyboard

  ### CSS

  **Parts**: `label`, `hint`, `segments`, `segment`, `segment-separator`

  **Custom properties**: `--segment-size`, `--segment-gap`, `--segment-border-radius`

  **Custom states**: `--blank`, `--filled`, `disabled`, `readonly`, `user-invalid`


packages/webawesome/src/components/otp-input/otp-input.ts
```diff
@@ -0,0 +1,515 @@
...
+            </div>
+          `;
+        })}
+
+        <input
+          id="hidden-input"
+          class="hidden-input"
+          type="text"
+          .value=${live(this._value)}
+          maxlength=${this.effectiveLength}
```
KonnorRogers: Possibly worth adding here:

```js
private getPattern () {
  const length = this.effectiveLength

  if (this.type === "alphanumeric") {
    return `[a-zA-Z0-9]{${length}}`
  }

  if (this.type === "alpha") {
    return `[a-zA-Z]{${length}}`
  }

  return `\d{${length}}`
}


pattern="${this.getPattern()}"
```

This will be a further check for proper formatted characters

packages/webawesome/src/components/otp-input/otp-input.ts
```diff
@@ -0,0 +1,515 @@
...
+
+    // Update the active segment to wherever the cursor is now
+    this._activeIndex = Math.min(cursorAfterFilter, this.effectiveLength - 1);
+
+    const prevLength = this._value.length;
+    const oldValue = this._value;
+    this._value = filtered;
+    this.setValue(filtered);
+
+    this.dispatchEvent(new InputEvent('input', { bubbles: true, composed: true }));
```
KonnorRogers: Should we use `this.relayNativeEvent()` here?

packages/webawesome/src/components/otp-input/otp-input.ts
```diff
@@ -0,0 +1,515 @@
...
+    if (this.input) this.input.value = next;
+    // When value changes while focused, move cursor to the end of the new value
+    if (this._focused) {
+      this._activeIndex = Math.min(next.length, this.effectiveLength - 1);
+    }
+    this.requestUpdate('value', oldValue);
+  }
+
+  /** The default value. Used to restore the field on form reset. Reflects the `value` HTML attribute. */
+  @property({ attribute: 'value', reflect: true }) defaultValue = this.getAttribute('value') ?? '';
```
KonnorRogers:
```suggestion
@property({ attribute: 'value', reflect: true }) defaultValue = this.getAttribute('value') ?? null
```

`?? '' ` will write `<wa-otp-input value="">` which is kind of wrong.

KonnorRogers (commented):
Things I found didn't work:

~~`ESC` key does not clear the input like you would expect~~ this only applies to `type="search"`

`cmd+a` does not select all characters and doesn't let you delete them

Things that worked well:

- `paste` seemed to work as expected
- Read well in a screen reader
- Appreciate that its a single input under the hood

Additional features:

- `autosubmit`, it seems small, but many otps will auto-submit when you have matched the criteria. (this may not be necessary, and i would definitely debounce it + opt in if you get there, perhaps future improvement?)



Other notes:

Have you tested this on a mobile device at all? I'm curious how it behaves there.

packages/webawesome/src/components/otp-input/otp-input.ts
kelseythejackson: @KonnorRogers It'd be a little redundant. `filterAndTransform()` already strips every character that doesn't match the type's class on every keystroke/paste, and minlength/maxlength (both already set to `effectiveLength`) already enforce exact length. Since `_value` can never actually contain an invalid character or be of the wrong length by the time it reaches `input.value`, a mismatch would never fire in practice; it'd be dead code that also duplicates the logic in a second place, so someone updating `filterAndTransform` later has to remember to keep `getPattern()` in sync.

packages/webawesome/src/components/otp-input/otp-input.ts
kelseythejackson: I think we can skip it, but I did have a bug in `handleInput`

packages/webawesome/src/components/otp-input/otp-input.ts
kelseythejackson: added

packages/webawesome/src/components/otp-input/otp-input.ts
KonnorRogers: makes sense 👍

claviska:
This is looking really good! I found a minor things that can be polished up, but not much.

---

Small a11y concern: The read-only example doesn't have a focus ring. Also, when you're focused into that example, you can see a subtle black outline around the active "text box", and pressing left and right arrow keys lets you move it around. I would expect not to see that interaction for a read-only control.

Screen cap of clicking into it, then pressing arrow keys:

CleanShot 2026-07-23 at 12 17 31

---

A minor improvement can be made when clicking on a pre-filled OTP. The last box always gets focused briefly if you select an earlier box:

CleanShot 2026-07-23 at 12 08 54

---

This example focuses on size, but we should probably show how to customize borders and backgrounds and custom states that may be interesting to the user.

CleanShot 2026-07-23 at 12 13 28@2x

---

While `:user-invalid` is preferred, this example should probably show `:invalid` since the customizations otherwise don't show. Users will have to type something invalid, otherwise they won't see a difference.

CleanShot 2026-07-23 at 12 14 19@2x

---

- Nice touch with autosubmit and the event 👌 

---

We may want to rename `placeholder` to `mask` or something similar. In every other form control placeholder is a string of text and, while this isn't documented in OTP, I wanted to see if I could get it to show an actual placeholder, e.g. "012345" and this was the result:

CleanShot 2026-07-23 at 12 24 03@2x

So if we're going to accept a single character, let's rename the property and trim the string. The other option is to support a more proper placeholder, which means `placeholder="012345"` would show each character, one per box. In that case, you'd also need to deal with placeholders that are shorter and longer than the number of boxes, which may feel awkward.

I'd lean towards something like `with-mask` that turns the mask on and defaults to what a password field shows. The character can then be configurable either via CSS, e.g. `--mask-char: '*'` or attribute, e.g. `mark-char="*"`.

packages/webawesome/src/components/otp-input/otp-input.ts
```diff
@@ -0,0 +1,602 @@
...
+    this.input.setSelectionRange(start, end);
+  }
+
+  formResetCallback() {
+    super.formResetCallback();
+    const reset = this.filterAndTransform(this.defaultValue ?? '').slice(0, this.effectiveLength);
+    const oldValue = this._value;
+    this._value = reset;
+    this.setValue(reset);
+    this._lastChangeValue = '';
```
talbs: After a form reset, focus→blur with no edit fires a spurious `change` (native inputs don't).  Should track the reset value:
```suggestion
    this._lastChangeValue = reset;
```

kelseythejackson: applied

packages/webawesome/src/components/otp-input/otp-input.ts
```diff
@@ -0,0 +1,631 @@
...
+import formControlStyles from '../../styles/component/form-control.styles.js';
+import sizeStyles from '../../styles/component/size.styles.js';
+import styles from './otp-input.styles.js';
+
+/**
+ * @summary OTP inputs collect one-time passcodes, PINs, and other fixed-length codes, one character per segment.
+ * Use them for SMS verification, two-factor authentication, and invite codes.
+ * @documentation https://webawesome.com/docs/components/otp-input
+ * @status experimental
+ * @since 3.10
```
claviska:
```suggestion
 * @since 3.11
```

claviska (approved):
Beautiful! :shipit:

PR #2586, opened 2026-07-02
Fix compact color picker swatches

### LINKED ISSUE #2571: `<wa-color-picker>` draws swatches too big when only a few are present

sjsjsjjs534:
## Summary

- Prevent color picker swatches from stretching when only a few presets are provided.
- Add a regression test covering compact swatch sizing.

Fixes #2571

## Test Plan

- npm run build
- CI=true FAIL_FAST=true npx web-test-runner --group color-picker

### DIFF
--- a/packages/webawesome/src/components/color-picker/color-picker.styles.ts
+++ b/packages/webawesome/src/components/color-picker/color-picker.styles.ts
@@ -216,7 +216,7 @@ export default css`
 
   .swatches {
     display: grid;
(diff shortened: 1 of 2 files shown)


claviska:
Thanks for the PR. https://github.com/shoelace-style/webawesome/pull/2573 landed first so I'm closing this one out.

PR #2588, opened 2026-07-02
Preserve color picker HEXA initial values

### LINKED ISSUE #2550: Colour Picker doesn't correctly load with a HEXA value
### Describe the bug
It looks like if you supply a HEXA value to the colour picker, that it won't interpret it correctly and the opacity value is lost.

### Demo
https://codepen.io/brendon-the-animator/pen/XJpggNd

You'll notice that the opacity value gets lost in the interface and the form will submit just a a HEX value unless the opacity is changed.

### Browser / OS
 - OS: Mac
 - Browser: Safari, Chrome
 - Browser version: Latest

sjsjsjjs534:
## Summary

- Preserve HEXA alpha values when a color picker initializes with `opacity`.
- Add a regression test covering the submitted form value for an initial HEXA color.

Fixes #2550

## Test Plan

- npm run build
- CI=true FAIL_FAST=true npx web-test-runner --group color-picker

### DIFF
--- a/packages/webawesome/src/components/color-picker/color-picker.test.ts
+++ b/packages/webawesome/src/components/color-picker/color-picker.test.ts
@@ -318,6 +318,18 @@ describe('<wa-color-picker>', () => {
       });
 
       describe('form integration', () => {
+        it('should preserve alpha when initialized with a HEXA value and opacity enabled', async () => {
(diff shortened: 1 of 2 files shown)


claviska:
Thanks for the fix! Applied in #2661, which is based off your branch.

PR #2589, opened 2026-07-02
Remove unsupported scroller aria orientation

### LINKED ISSUE #2364: wa-scroller: aria-orientation on role="region" violates aria-allowed-attr
## Summary

`<wa-scroller>` renders an inner `<div role="region" aria-orientation="...">` in its shadow DOM. The combination violates the ARIA spec: `aria-orientation` is only allowed on a small set of roles (`scrollbar`, `slider`, `separator`, `tablist`, `toolbar`, `tree`, `menu`, `radiogroup`, `listbox`, `select`). `region` is not one of them, so axe-core flags it as a critical `aria-allowed-attr` violation under WCAG 4.1.2.

## Environment

- Web Awesome: 3.6.0
- Browser: Chrome 130 (also reproduces in Firefox 132, Safari 17.6)
- axe-core: 4.11

## Reproduction

```html
<wa-scroller>
    <div style="width: 1500px;">Scrollable content</div>
</wa-scroller>
```

Inspect the rendered shadow DOM — the `#content` element looks like:

```html
<div id="content"
     part="content"
     role="region"
     aria-label="Scrollable region"
     aria-orientation="horizontal"
     tabindex="0">
    <slot></slot>
</div>
```

Run axe-core against the host element (or any ancestor):

```js
const { violations } = await axe.run(document.querySelector('wa-scroller'));
console.log(violations);
```

Output:

```json
[
    {
        "id": "aria-allowed-attr",
        "impact": "critical",
        "description": "Ensure an element's role supports its ARIA attributes",
        "help": "Elements must only use supported ARIA attributes",
        "helpUrl": "https://dequeuniversity.com/rules/axe/4.11/aria-allowed-attr",
        "nodes": [{
            "html": "<div id=\"content\" part=\"content\" role=\"region\" aria-label=\"Scrollable region\" aria-orientation=\"horizontal\" tabindex=\"-1\">",
            "failureSummary": "ARIA attribute is not allowed: aria-orientation=\"horizontal\""
        }]
    }
]
```

## Source

The attribute is set in [`scroller.styles.js`](https://github.com/shoelace-style/webawesome/blob/main/src/components/scroller/scroller.ts) (paraphrased from the bundled chunk):

```ts
<div
    id="content"
    part="content"
    role="region"
    aria-label=${this.localize.term('scrollableRegion')}
    aria-orientation=${this.orientation}   // ← here
    tabindex=${this.canScroll ? '0' : '-1'}
    @keydown=${this.handleKeyDown}
    @scroll=${this.updateScroll}
>
```

## Spec references

- [ARIA 1.2 §6.6.5: aria-orientation valid roles](https://www.w3.org/TR/wai-aria-1.2/#aria-orientation) — explicitly enumerates the supported roles; `region` is not among them.
- [ARIA region role](https://www.w3.org/TR/wai-aria-1.2/#region) — supports `aria-label` / `aria-labelledby` / global states, but not `aria-orientation`.

sjsjsjjs534:
## Summary

- Remove the unsupported `aria-orientation` attribute from the scroller region.
- Update the scroller accessibility test to cover the allowed ARIA attribute set.

Fixes #2364

## Test Plan

- npm run build
- CI=true FAIL_FAST=true npx web-test-runner --group scroller

### DIFF
--- a/packages/webawesome/src/components/scroller/scroller.test.ts
+++ b/packages/webawesome/src/components/scroller/scroller.test.ts
@@ -27,11 +27,11 @@ describe('<wa-scroller>', () => {
           expect(content).to.have.attribute('aria-label');
         });
 
-        it('should set aria-orientation based on orientation property', async () => {
+        it('should not set aria-orientation on the region content container', async () => {
           const el = await fixture<WaScroller>(html` <wa-scroller orientation="vertical"></wa-scroller> `);
 
           const content = el.shadowRoot!.querySelector('#content')!;
-          expect(content).to.have.attribute('aria-orientation', 'vertical');
+          expect(content).to.not.have.attribute('aria-orientation');
         });
       });
 
--- a/packages/webawesome/src/components/scroller/scroller.test.ts
+++ b/packages/webawesome/src/components/scroller/scroller.test.ts
@@ -26,13 +26,6 @@ describe('<wa-scroller>', () => {
           const content = el.shadowRoot!.querySelector('#content')!;
           expect(content).to.have.attribute('aria-label');
         });
-
-        it('should not set aria-orientation on the region content container', async () => {
-          const el = await fixture<WaScroller>(html` <wa-scroller orientation="vertical"></wa-scroller> `);
-
-          const content = el.shadowRoot!.querySelector('#content')!;
-          expect(content).to.not.have.attribute('aria-orientation');
-        });
       });
 
       describe('properties', () => {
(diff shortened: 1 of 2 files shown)


packages/webawesome/src/components/scroller/scroller.test.ts
```diff
@@ -27,11 +27,11 @@ describe('<wa-scroller>', () => {
...
 
-        it('should set aria-orientation based on orientation property', async () => {
+        it('should not set aria-orientation on the region content container', async () => {
           const el = await fixture<WaScroller>(html` <wa-scroller orientation="vertical"></wa-scroller> `);
 
           const content = el.shadowRoot!.querySelector('#content')!;
-          expect(content).to.have.attribute('aria-orientation', 'vertical');
+          expect(content).to.not.have.attribute('aria-orientation');
         });
       });
```
KonnorRogers: I think we can just delete this test, no need to test this in my opinion. Its kind of just noise.

KonnorRogers (changes requested):

KonnorRogers (approved):

PR #2594, opened 2026-07-02
Apply inverted colors to native button variants

### LINKED ISSUE #2533: .wa-invert Does Not Work On Some Native Elements
### Describe the bug
I'm not sure if this is actually a bug, just needs documentation, or I missed the documentation about it, but using the `".wa-invert"` class does not seem to affect at least some native elements. The extent of my testing has been using `button` with a wrapping container. I also tested whether putting `"wa-invert"` and `"wa-theme-default wa-invert"` directly on the button would work, and it does not. Notably, all these cases work fine with `h2` elements (as seen in the Codepen below).

I'm not sure exactly _why_ it doesn't work with `<button>`. It may have to do with the way the coloration is handled with nesting in the CSS, in which case I would expect other form controls to have similar behavior.

### To Reproduce

1. Go to https://codepen.io/stevematney/pen/wBgdgJL?editors=1000
2. Review the variance between Web Component and native buttons.

### Demo

See button example here: https://codepen.io/stevematney/pen/wBgdgJL?editors=1000

### Browser / OS
 - OS: Mac
 - Browser: Chrome
 - Browser version: 149.0.7827.115

sjsjsjjs534:
## Summary

- Let `.wa-invert` establish the default neutral variant aliases used by native styles.
- Add native style regression tests for buttons inside `.wa-invert` and buttons with `.wa-invert` directly applied.

Fixes #2533

## Test Plan

- `npm run build`
- `CI=true FAIL_FAST=true npx web-test-runner --group native-styles`

### DIFF
--- a/packages/webawesome/src/internal/native-styles.test.ts
+++ b/packages/webawesome/src/internal/native-styles.test.ts
@@ -1,0 +1,59 @@
+import { expect, fixture } from '@open-wc/testing';
+import { html } from 'lit';
+
+async function loadNativeStyles() {
+  const existing = document.querySelector<HTMLLinkElement>('link[data-test-native-styles]');
+  if (existing) return;
+
+  const link = document.createElement('link');
+  link.rel = 'stylesheet';
+  link.href = '/dist/styles/native.css';
+  link.dataset.testNativeStyles = '';
+  document.head.append(link);
+
+  await new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


claviska (approved):
Just need a changelog and this one is ready per https://github.com/shoelace-style/webawesome/issues/2533

claviska:
Thanks for the PR. Merged as part of https://github.com/shoelace-style/webawesome/pull/2678

PR #2605, opened 2026-07-06
Page: Fix Empty `navigation-footer` in Mobile Drawer

talbs:
### Issue
`wa-page` unconditionally forwards a `<slot slot="footer">` into its mobile `<wa-drawer>`. 

Because `wa-drawer` decides footer visibility structurally, that forwarded slot counts as content even when it projects nothing. So the drawer renders an empty `.footer` whose padding steals the bottom of the nav area. 

FYI, on desktop the footer lives in a grid row that collapses to zero, so only the mobile drawer shows the dead space.

### Proposed Fix
Detect a slotted `navigation-footer` with a `HasSlotController` and only forward the mobile footer slot when it has content, following the same idiom as `wa-drawer`, `wa-dialog`, and `wa-card`.

With no forwarded slot, `wa-drawer`'s `hasFooter` is correctly `false` and its footer stays `hidden`, so the nav fills the drawer.

### Notes

- Added `mobile drawer footer` tests (CSR + SSR-hydrated), asserting the footer slot is forwarded only when `navigation-footer` is slotted. All 60-page tests pass in Chromium/Firefox/WebKit.

## Related PRs

- Docs band-aid that hides the empty footer until this lands: https://github.com/shoelace-style/webawesome/pull/2604

### DIFF
--- a/packages/webawesome/src/components/page/page.ts
+++ b/packages/webawesome/src/components/page/page.ts
@@ -1,8 +1,9 @@
 import type { PropertyValues } from 'lit';
-import { html, isServer } from 'lit';
+import { html, isServer, nothing } from 'lit';
 import { customElement, property, query } from 'lit/decorators.js';
 import { live } from 'lit/directives/live.js';
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import { HasSlotController } from '../../internal/slot.js';
 import WebAwesomeElement from '../../internal/webawesome-element.js';
 import visuallyHidden from '../../styles/component/visually-hidden.styles.js';
 import '../button/button.js';
@@ -172,6 +173,8 @@ export default class WaPage extends WebAwesomeElement {
   @query("[part~='drawer']") navigationDrawer: WaDrawer;
   @query("slot[name~='navigation-toggle']") navigationToggleSlot: HTMLSlotElement;
 
+  private readonly hasSlotController = new HasSlotController(this, 'navigation-footer');
+
   /**
    * The view is a reflection of the "mobileBreakpoint", when the page is larger than the `mobile-breakpoint` (768px by
    * default), it is considered to be a "desktop" view. The view is merely a way to distinguish when to show/hide the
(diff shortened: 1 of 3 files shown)


talbs:
@claude review this PR.

packages/webawesome/src/components/page/page.ts
```diff
@@ -435,9 +447,11 @@ export default class WaPage extends WebAwesomeElement {
...
         </slot>
 
-        <slot slot="footer" name="mobile-navigation-footer">
-          <slot part="navigation-footer" name=${this.view === 'mobile' ? 'navigation-footer' : '___'}></slot>
-        </slot>
+        ${hasNavigationFooter
+          ? html`<slot slot="footer" name="mobile-navigation-footer">
+              <slot part="navigation-footer" name=${this.view === 'mobile' ? 'navigation-footer' : '___'}></slot>
+            </slot>`
+          : nothing}
```
KonnorRogers, line 454: Conditional rendering like this breaks because the slotchange event never fires because the slot is never rendered.

It'd have to be more like:

```suggestion
<slot slot="footer" name="mobile-navigation-footer" ?hidden="${!hasNavigationFooter}">
  <slot part="navigation-footer" name=${this.view === 'mobile' ? 'navigation-footer' : '___'}></slot>
</slot>
```

and just hide it that way instead, that way it still lives in the DOM and can detect slot changes.

KonnorRogers (changes requested):

packages/webawesome/src/components/page/page.ts
KonnorRogers: The `hidden` _should_ be respected by the hasSlotController in `<wa-dialog>`

talbs:
Backing this out. 

The fix is a component-design call: `wa-page` forwards an empty footer slot, and `wa-drawer` counts it structurally. 

Every simple fix handles the static case, but none restore runtime add/remove without watching `hidden` mutations. Wrote it up for the team to decide - https://github.com/shoelace-style/webawesome/issues/2609. 

The Docs band-aid that landed (#2604) holds until then.

PR #2613, opened 2026-07-07
Make `tabular-nums` opt-in for tables

### LINKED ISSUE #2599: tabular-nums affects `a.wa-button` in table
### Describe the bug
#2459 added `font-variant-numeric: tabular-nums` to `<table>`. This probably should not affect native link buttons `a.wa-button`. `<input>` and  `<button>` are not affected.

### To Reproduce

1. Inside a `<table>` add `<a href="" class="wa-button">Call 911</a>`
1. See that '911' is using tabular numbers

### Demo
https://codepen.io/roydukkey/pen/PwWOOvR?editors=1000

### Browser / OS
 - OS: Mac
 - Browser: Chrome
 - Browser version: 149.0.7827.197 (Official Build) (arm64)

omnifides:
Closes #2599 

We encountered two concerns with applying `font-variant-numeric: tabular-nums;` to `<table>` elements by default:
1. The set `font-variant-numeric` doesn't uniformly affect elements within tables because user agents set their own values for specific elements, like `<button>` and `<input>`
2. There are plenty of potential table data where `tabular-nums` isn't appropriate, such as usernames, button labels, and numbers within paragraphs

This work removes `font-variant-numeric: tabular-nums;` from the default `<table>` styles in `native.css` and instead adds an opt-in convenience class, `wa-tabular-nums`

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -40,6 +40,11 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 - Fixed a lifecycle issue in `<wa-option>` [pr:2591]
 - Fixed a bug with extra margin on the first button of a `<wa-button-group>` [pr:2592]
 
+:::
+
+:::changed
+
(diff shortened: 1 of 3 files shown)


talbs (approved):
Great solution and solid additions/revisions to the docs.

PR #2621, opened 2026-07-09
Docs: Add Visual Anatomy Diagram to Component Parts

talbs:
This work adds an interactive **anatomy diagram** to every component doc page.

The anatomy diagram is a live instance of the component on a stage where hovering the CSS Parts table highlights each part on the component (and hovering a part highlights its row). 

It's progressive enhancement — the CSS Parts table stays the source of truth and the no-JS fallback.

| Highlighted Part | Highlighted Part + Scroll | Neutral Part | 
|--------|--------|--------|
| CleanShot 2026-07-13 at 14 02 32@2x | CleanShot 2026-07-13 at 14 03 22@2x |  CleanShot 2026-07-13 at 14 04 01@2x | 

---

| Live Previews |
|--------|
| [Select](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/select#css-parts) |
| [Button](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/button#css-parts) |
| [Input](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/input#css-parts) |
| [Tag](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/tag#css-parts) |
| [Slider](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/slider#css-parts) |
| [Avatar](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/avatar#css-parts) - Minimal Example | 
| [Color Picker](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/color-picker#css-parts) - Opt-Out Example |

### The Visuals
A docs-only `<component-anatomy>` element (`docs/assets/scripts/component-anatomy.js` + `anatomy.css`), emitted from `component.njk` next to the CSS Parts table. 

It clones the page's example onto a dot-grid stage and, for each part, draws an accent ring + wash and a `wa-badge` part-name label on hover/focus. Hovering a part highlights its table row; clicking a part scrolls to its row (handy for long tables).

### Choosing the Anatomy Example
Two fence flags in the `code-examples` transformer:
- `{.example .anatomy}` — use an existing visible example as the diagram subject (unchanged content).
- `{.example .anatomy-only}` — a dedicated richer example that renders but is **hidden from the visible examples list**, so the page's real examples stay simple. Used where the common parts (`hint`, `start`, `clear-button`, …) aren't shown by the basic example.

### State Toggles
The stage sits in a `wa-card`; the header offers toggles states, which show more situational parts. Toggles are auto-offered only when the component has such a part — derived from a small `part → state` map.

### What's Excluded
- `hasAnatomy: false` front matter opts a component out.
- Overlays like `dialog`/`tooltip`/`dropdown`, layout wrappers like `page`/`scroller` are set to `hasAnatomy: false (see below on why).
- Sub-components (anything with a front-matter `parent` — `tab`, `option`, `tree-item`, …) are also opted out automatically as they don't render meaningfully standalone.
- If nothing is highlightable, the JS drops the stage at runtime and leaves the table alone.

### What This PR Doesn't Cover
The diagram only measures parts that render **on the stage** — the component's normal shadow DOM. 

Parts that appear in an *open* state and render in a popup or the top layer (a `select`'s `listbox`, a `time-input`'s picker) aren't highlighted; they stay in the CSS Parts table, just without a region. 

Reaching them means driving each component open *and* following the part outside the stage's coordinate space — per-component work, left for a follow-up.

## Companion PRs
- https://github.com/shoelace-style/webawesome-pro/pull/260


talbs:
@claude, review this PR.

packages/webawesome/docs/docs/components/drawer.md
```diff
@@ -2,6 +2,7 @@
 title: Drawer
 layout: component
 category: Layout
+hasAnatomy: false
```
claviska, line 5: Might be better to opt in instead of opting out.

claviska (commented):
This addresses a valid pain point when styling custom elements.

I only had a chance to look at this briefly, but it seems very promising and I think we should continue exploring here. The tab-like interface for showing various states is fantastic.

One thing I noticed is that there's no way to tab into the anatomy. So a roving tab index that allows keyboard users to explore would be a good next step.

This is extremely promising. Great work, @talbs! ❤️

packages/webawesome/docs/docs/components/drawer.md
talbs: I went back and forth on this. I think I'd lean towards **opt-out**.

The value is that this Anatomy UI auto-generates from the CEM and covers new components for free. 

Opt-in kills that: every new component needs a flag, and coverage quietly lags whatever people remember to tag.

The guards catch the structural cases (sub-components, no `@csspart`, nothing-to-measure drops itself at runtime). They don't catch "renders but ugly." So we'd need to proof new stuff as we build the docs and add a manual `hasAnatomy: false` if this new UI looks off/isn't appropriate.

☝️ The upside is the opt-out list stays short and stable (overlays, page/scroller). Everything else just works.

talbs:
Okay, gang. Took @claviska's good [proactive feedback](https://github.com/shoelace-style/webawesome/pull/2621#pullrequestreview-4675173925) and made a few tweaks.

**Non-Hover Case Support**
Hover isn't the only way to highlight part info in the anatomy diagram. The part names in the CSS Parts table are buttons (progressive enhancement, so no-JS still gets a plain table). Focus or tap one and it highlights that part on the diagram — so keyboard and touch users get the same highlight that mouse users already had.

**Sticky Anatomy Diagram**
On long part tables (e.g. the Slider component), the diagram pins as you scroll, so it stays paired with the row you're reading for easy visual association.

This is now ready for review - happy to make the UI or code behind this better.

talbs:
@claude, review my changes since your last review.

talbs:
Okay, @claude. All caught up - please re-review.

lindsaym-fa (dismissed):
This is awesome! 🤩

I'd love to see additional states/variations someday (`wa-avatar` is a good one — `icon` vs. `image` vs. `initials`), but those definitely aren't necessary for this to be valuable.

It would also be nice to figure out a solution for child components that don't currently feature any examples or show an anatomy diagram for their parts, like `wa-radio` and `wa-option`. For future consideration, I wonder if we should bake those into the parent component's diagram somehow, or at least offer an `.anatomy-only` example on those docs.

We do have at least one part that seems to be documented with the wrong name: `wa-checkbox` documents `::part(checked-icon)`, and that likely is the name it _should_ be to match other components like `wa-radio`, but the part is actually named `check-icon`. Let me know if you want me to address this or if you want to roll this into any part-naming-consistency efforts.

talbs:
Thanks for the review, @lindsaym-fa. Here's a quick update on some work done since that look...

### Child Components
These are now done/covered. Added `breadcrumb-item`, `accordion-item`, and `dropdown-item` on top of radio and option, and gave them the parent context so they're actually legible.

### Avatar Variations
These options are now documented. The icon/initials/image modes are the `Default`/`Initials`/`Image` toggles. They can't coexist, so a toggle is the only way to show all three.

### Default Anatomy Preview, Toggle States, and No Anatomy
I tightened up some logic and rules to help govern the following:

- The base example should look like something you'd actually ship, not a kitchen sink with every part crammed 
- Mutually-exclusive parts and non-default modes go behind toggles
- Toggle States are validated - so any dead/untrue ones are gone
- Parts that can't render statically stay table-only, and the probe enforces that on its own. 

### Fixes + Polish
- Part-name buttons don't shift the table row anymore.
- Trimmed the stage's vertical whitespace. 
- The diagram also stays hidden until the JS probe finishes, so it doesn't flicker on load.

### Checkbox Bug
I've got a fix in a separate branch for this - once it is merged in, this work will pick it up automatically and correctly.

talbs:
> We do have at least one part that seems to be documented with the wrong name: wa-checkbox documents ::part(checked-icon), and that likely is the name it should be to match other components like wa-radio, but the part is actually named check-icon. Let me know if you want me to address this or if you want to roll this into any part-naming-consistency efforts.

@lindsaym-fa, this should be tackled by https://github.com/shoelace-style/webawesome/pull/2646.

claviska:
[Tag has](https://webawesome-git-talbs-component-anatomy-font-awesome.vercel.app/docs/components/tag#css-parts) a `display: contents` on the `content` part which means you can't reliably get the coordinates. Without diving into the code, I believe that's causing this (the `content` part is actually the slot itself):

CleanShot 2026-07-20 at 16 16 56@2x

That said, we probably _shouldn't_ be putting parts on slots directly. I now consider this an antipattern, so perhaps the solution is adding a `content` wrapper around the slot.

claviska:
Minor visual thing with the corners here:

CleanShot 2026-07-20 at 16 26 23@2x

omnifides (dismissed):
Great updates! ✨

talbs:
Okay, @claviska - those good catches are now addressed. Ready for the rest of your review when you can.

packages/webawesome/docs/assets/scripts/component-anatomy.js
```diff
@@ -0,0 +1,583 @@
...
+// measuring assigned *elements* alone would box just an icon and miss the words beside it.
+function nodeRect(node) {
+  if (node.nodeType !== Node.TEXT_NODE) return node.getBoundingClientRect?.();
+  if (!node.textContent.trim()) return null;
+  const range = document.createRange();
+  range.selectNodeContents(node);
+  return range.getBoundingClientRect();
+}
+
+// Bounding box for a part; a degenerate rect (a `display: contents` slot, or slider's 0-height `markers`
```
claviska, line 57: degenerate rect! 😆

claviska:
This is great! Component anatomy (paired with new concepts such as `::part()` and `::state()` remain the biggest hurdles for users trying to customize web components.

Just a nit:

CleanShot 2026-07-21 at 17 30 39@2x

When I click on a part and then hover to another, I see both highlighted but only for a second or two, then it corrects. Is that expected?

On second look, it seems simply clicking a part will cause it to focus/show and then remain, even after hovering outside of the part. Normal hovering works more like a tooltip.

If you really want to polish this off:

- Ensure only one part is ever highlighted at a time
- Add a hover delay (similar to tooltips) so hovering briefly outside of the boundary doesn't hide the selection abruptly

And I'm not sure about this one, but is it better or worse if the translucent orange background on selected parts is removed?

---

My previous comment about a roving tab index isn't relevant, since users can tab through the part list which will perform the highlight. This seems better, since there's less redundant tabbing.

---

How do these get generated? I don't see any changes to the Slider component itself and the markdown file doesn't seem to include them. How does it know to add these states?

CleanShot 2026-07-21 at 17 38 08@2x

claviska (dismissed):
Approved with some comments, nits, and a question. Nothing blocking, mostly polish.

This is a huge step forward for component customization. Outstanding work!

packages/webawesome/docs/assets/scripts/component-anatomy.js
talbs: That's the name of my new Swedish Punk Rock Band!

talbs:
@claviska, Good catches — fixed the highlight behavior.

- The two-at-once was two separate mechanisms that never cleared each other: hover and focus set one class, click set another. Collapsed them into a single active highlight resolved from hover, click, and focus.
- On the fill: I actually like it, so I kept it but dropped it from 12% to 8% and matched it to the table-row hover so they read as one thing. Much quieter over a photo now.
- Roving tab index: agreed, the part list covers it.

---

> How do these get generated?

Nothing lives in the Slider component or its markdown. It's a `STATE_MAP` in `component-anatomy.js`, keyed by part name: `markers: { label: 'Markers', attrs: { 'with-markers': '' } }`.

PR #2623, opened 2026-07-10
Note Rendered-But-Undocumented Parts + Add Guard

talbs:
Several components render a `part="…"` attribute that they never declared with a matching `@csspart`. So those parts are invisible in the Docs' CSS Parts table (and anything built from it). 

This work documents the ones that were unambiguously just missing, and adds a build guard so the rendered and documented part surfaces can't silently drift apart again. 

### Add Documented Parts
Added the missing `@csspart` tags for parts that already render:

- `card`: `actions`
- `color-picker`: `trigger-container`
- `textarea`: `textarea-adjuster`
- `scroller`:`start-shadow`, `end-shadow`
- `page`: `main`, `drawer`, `navigation-desktop`

### Add Coverage Guard
- New `scripts/check-css-parts.js`, wired as `npm run test:parts`. 
- It fails when a component renders a `part=` with no matching `@csspart`. 
- Static analysis only — no build. 
- Multi-part attributes (`part="foo bar"`) are split and checked per token.

### What's Deliberately Deferred
Three findings are API-shape questions, not mechanical fixes, so they're held behind the guard's `ALLOWLIST` and tracked for a decision in a follow-up issue at https://github.com/shoelace-style/webawesome/issues/2624.

FYI: The allowlist entries are self-expiring. The guard reports any that become stale, so they can't quietly turn into a rubber stamp.


talbs:
@claude review this PR.

talbs:
@claude, feedback addressed!

packages/webawesome/scripts/check-css-parts.js
```diff
@@ -0,0 +1,112 @@
...
+import { basename, dirname } from 'node:path';
+import process from 'node:process';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const root = dirname(__dirname);
+
+// Static `part=` attributes only. The lookbehind rejects `exportparts=` and `[part=…]` selectors.
+const RENDERED_PART = /(?<![\w[])part=(["'])([^"'$]+)\1/g;
+const DECLARED_PART = /@csspart\s+(\S+)/g;
```
KonnorRogers, line 19: What could possibly go wrong (though this is probably good enough, and we can ignore the bits that slip) 😅

KonnorRogers (approved):

Issue #2624, opened 2026-07-10
Inconsistent + Undocumented Part Surfaces Across Several Components


talbs:
While auditing CSS parts, I found several components that render a `part="…"` attribute with no matching `@csspart` tag, plus one systemic inconsistency in how form controls name their label part. 

The purely mechanical gaps (a rendered part that's just missing its tag) are easy to fix. 

**But three of them are decisions about the public `::part()` surface.** So I'm raising them here rather than just opening a PR.

### 1. The Form-Control Label Part is Documented 5 Different Ways
Every form control renders a label element, but the rendered part name(s) and the documented tags drift per component:

| Component | Renders on the label | Documents |
| --- | --- | --- |
| `input` | `form-control-label label` | `label`, `hint` |
| `textarea` | `form-control-label label` | `label`, `form-control-input`, `hint` |
| `number-input` | `form-control-label label` | `label`, `form-control-label`, `hint` |
| `select` | `form-control-label label` | `form-control-label`, `form-control`, `form-control-input`, `hint` |
| `time-input` | `form-control-label label` | `form-control-label`, `form-control`, `form-control-input`, `hint` |
| `checkbox-group` | `form-control-label` | full set |
| `radio-group` | `form-control-label` | full set |
| `color-picker` | `form-control-label` | *(none)* |
| `known-date` | `field-label form-control-label label` | full set + `label` |

Two things are tangled:

#### (a) Which label name is canonical?
Single-value controls render *both* `form-control-label` and `label` on the same element, so both already work as `::part()` targets. 

Groups render only `form-control-label`. The docs then scatter: `input`/`textarea` document only `label`, `select`/`time-input` only `form-control-label`, `number-input` both, `color-picker` neither.

#### (b) Shared Parts
`form-control`, `form-control-input`, `hint` are documented on most controls and missing on others (`input` has no `form-control-input`; `color-picker` documents nothing).

**Options**
- **A — Document both names everywhere they render; standardize the shared set**
- **B — Pick one canonical label name, deprecate the other**
- **C — Make groups symmetric** by adding `part="label"` to `checkbox-group`/`radio-group`/`color-picker`

**Proposed:** A now (unblocks documentation with zero risk), with B/C tracked separately as the real "canonical form-control part vocabulary" call for the future. 

Note: `known-date` also renders an undocumented `field-label` — fold it into whatever's decided.

---

### 2. `page` Documents a Skip Link that Doesn't Exist
[...]

Issue #2626, opened 2026-07-10
Remove the base part. Finally.


claviska:
For Web Awesome 4, it's [finally] about time to finally get rid of the `base` part convention and use part names that are actually helpful to people.

PR #2632, opened 2026-07-13
Accordion: Expose `base` Part

talbs:
This work resolves item 3 of https://github.com/shoelace-style/webawesome/issues/2624...

>  `<wa-accordion>` rendered only a bare `<slot>` and exposed no `::part()` hook for its container, the one gap among its container siblings. 

☝️ This adds a `base` part.

### Y U NO Use Wrapper?
The issue floated wrapping the slot in a `<div part="base">`. Putting the part on the existing `<slot>` instead — the way `<wa-button-group>` already does it — means zero new DOM, so there's no risk of disturbing the accordion's layout, which was the open concern in the thread.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -33,7 +33,7 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 
 :::added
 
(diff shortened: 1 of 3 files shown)


claviska (commented):
Can we get away from using `base` as a part name in preparation for WA 4? See https://github.com/shoelace-style/webawesome/issues/2626

talbs:
Closing this as we'll revisit it with another named part other than `base`.

PR #2633, opened 2026-07-13
Docs: Fix Page `skip-to-content` Part

talbs:
This resolves item 2 of https://github.com/shoelace-style/webawesome/issues/2624...

> `<wa-page>` documented `@csspart skip-links` and `@csspart skip-link`, but neither renders. 

The element only ever exposes `part="skip-to-content"`, which went undocumented. 

☝️ This points the docs at the part that actually exists. And this also drops the now-stale `page: skip-to-content` entry from the `check-css-parts.js` allowlist which had been suppressing the "rendered but undocumented" warning for exactly this part.


claviska (dismissed):
Nice catch! looks good

PR #2638, opened 2026-07-14
Fix: Safari Visual Seam on Tooltip and Popover Arrows

talbs:
Our Tooltip and Popover-based visual arrows are rotated squares with `clip-path` plus `border-bottom`/`border-right`.

But Safari leaves a transparent gap between a `border` and the background fill of a clipped, rotated element.

To sidestep this bug, this work paints the arrow's border with `box-shadow: inset` instead of `border`, on both components. No border means no border/fill boundary, so nothing seams. 

| Safari (pre-fix)  | Safari (post-fix) |
|--------|--------|
| CleanShot 2026-07-14 at 12 36 00@2x | CleanShot 2026-07-14 at 12 35 48@2x | 

It reads the same tokens as before (`--wa-tooltip-border-color`, `--wa-tooltip-border-width`, and the popover's `--wa-color-surface-border` / `--wa-panel-border-width`), so nothing changes outside Safari.

### One Tradeoff
`--wa-tooltip-border-style` and `--wa-panel-border-style` no longer reach the arrow. `box-shadow` is always solid. This may be okay because:

- Every shipped theme currently uses `solid`, so there's no visual change. 
- And a dashed arrow never actually worked: the rotated arrow's dashes fall out of phase with the body and collide at the base. We dropped a broken mode, not a real one.

### Verifications
- Tested across all placements
- Tested across Core and Pro themes
- Tested across Light and Dark modes
- Tested across size/color overrides.
- `::part(base__arrow)` stays fully overridable (users can clear it with `box-shadow: none` or paint their own). 
- Tooltip and Popover suites pass in Chromium, Firefox, and WebKit.

### Docs
Updated the `wa-popup` "border on popup" example to the `box-shadow` pattern so nobody copies the seaming version.

### DIFF
--- a/packages/webawesome/docs/docs/components/popup.md
+++ b/packages/webawesome/docs/docs/components/popup.md
@@ -544,9 +544,10 @@ When adding borders to the popup element which has an arrow, make sure to set th
       --arrow-color: var(--wa-color-brand-on-loud);
       --popup-border-width: var(--wa-panel-border-width);
 
+      /* Inset box-shadow, not a border: Safari seams a clip-path edge that runs along a border. */
       &::part(arrow) {
-        border-bottom: var(--wa-panel-border-width) var(--wa-panel-border-style) var(--wa-color-brand-border-loud);
-        border-right: var(--wa-panel-border-width) var(--wa-panel-border-style) var(--wa-color-brand-border-loud);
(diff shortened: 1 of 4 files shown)


lindsaym-fa (approved):
Works well!

PR #2641, opened 2026-07-14
Docs: Highlight "Docs" in Site Nav

talbs:
This work fixes the subheader nav's active state so it reflects where you actually are. 

Previously, `Docs` and `Components` only showed as current/highlighted on their exact URL, so any nested docs page — `/docs/usage`, a component page, or the changelog — left the nav empty.

### Behavior
- Component pages (`/docs/components/**`) light **Components**.
- Every other docs page (`/docs`, `/docs/usage`, `/docs/frameworks/**`, resources, …) lights **Docs**.

### How
- The `current-link` transformer gains a `data-match="section"` opt-in (a link matches its whole subtree without changing its `href`) plus a most-specific-wins tie-break scoped to the subheader nav — so `Docs` yields to the deeper `Components` under `/docs/components/**`.
- `Components` and `Docs` opt in via `data-match="section"`. `href`s are unchanged, so navigation targets and analytics destinations are untouched, and the sidebar's exact-match behavior is unaffected.

### DIFF
--- a/packages/webawesome/docs/_includes/subheader-links.njk
+++ b/packages/webawesome/docs/_includes/subheader-links.njk
@@ -3,10 +3,10 @@
   <wa-button appearance="plain" size="s" href="/docs" data-track-event="navigation:header_link_click" data-track-destination="start">
     Start
   </wa-button>
-  <wa-button appearance="plain" size="s" href="/docs/components" data-track-event="navigation:header_link_click" data-track-destination="components">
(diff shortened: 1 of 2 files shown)


talbs:
@claude, review this PR.

talbs:
@claude,

- **Specificity comment** — expanded to spell out the invariant it relies on.
- **`Start`/`Docs` double-highlight on `/docs`** — agreed it's pre-existing, so I'm leaving it out of this PR and will track it separately.
- **Equal-depth section tie** — intentional: ties staying lit is what keeps `Start` + `Docs` both correct on `/docs`.

KonnorRogers (approved):

PR #2642, opened 2026-07-14
allow passing ids to wa-include

KonnorRogers:


claviska (commented):
Nice enhancement! Just needs a changelog entry and a docs example.

claviska:
@KonnorRogers I added support for non-template content, updated the docs, added a changelog entry, and merged the latest in. Since I added functionality, would you mind verifying?

PR #2644, opened 2026-07-15
Replace Generic `base` with Component-Named Parts

talbs:
This work supersedes the closed https://github.com/shoelace-style/webawesome/pull/2632 .

It attempts to settle the naming question from https://github.com/shoelace-style/webawesome/issues/2624 and https://github.com/shoelace-style/webawesome/issues/2626. 

`base` is a generic name we've been stuck with since v3. This gives every component a part named after itself (`accordion`, `details`, `tab-panel`) and quietly demotes `base` to an alias that still works.

### The Convention
- Every component that renders a `base` wrapper now renders its own name alongside it. 
- `<wa-details>` becomes `part="base details"`, `<wa-tab-panel>` becomes `part="base tab-panel"`, and so on. 
- Where the component name is already spoken for — `input` and `textarea` use it for their native control:  the wrapper takes a `-wrapper` suffix instead: `input-wrapper`, `textarea-wrapper`.
- `accordion` never had a `base` (it was a bare `<slot>`), so it just gets `part="accordion"`, no `base` at all. 
- New components work the same way: the Plop scaffold now emits the canonical part and drops `base` entirely, so it stops spreading.
- Components with no wrapper to name are left alone. `card`, `dropdown`, and `tag` are the container themselves, so you style them on the host. The headless utilities (`format-*`, the observers) render nothing and get nothing.

### `base` Doesn't Go Away
- `base` is still a real `::part()` target. 
- Every `wa-foo::part(base)` selector out there keeps working, and it stays in the custom-elements manifest so tooling and IDEs still see it. 
- The only change: it's filtered out of the generated CSS Parts table in the docs, so the canonical name is what people actually read. 
- Deprecating it is a v4 job, removing it a v5 one.

### Enforcement
`check-css-parts.js` fails the build if a `base`-rendering component forgets its canonical part, so this can't rot back to `base`-only.  Its allowlist is parameterized so the Pro package reuses the same checker instead of duplicating it.

### One Bug Fix
`dropdown` was grabbing its trigger button's base with an exact `[part="base"]` selector. 

That broke the second the button started rendering `part="base button"` — the query returned null and `aria` attributes stopped getting set. Switched it to `[part~="base"]`.

## Companion PRs
- https://github.com/shoelace-style/webawesome-pro/pull/263

### DIFF
--- a/packages/webawesome/scripts/plop/templates/component/component.hbs
+++ b/packages/webawesome/scripts/plop/templates/component/component.hbs
@@ -15,8 +15,6 @@ import styles from './{{ tagWithoutPrefix tag }}.styles.js';
  * @slot - The default slot.
  * @slot example - An example slot.
  *
- * @csspart base - The component's base wrapper.
- *
  * @cssproperty --example - An example CSS custom property.
  */
  @customElement("{{ tag }}")
--- a/packages/webawesome/scripts/plop/templates/component/component.hbs
+++ b/packages/webawesome/scripts/plop/templates/component/component.hbs
@@ -15,7 +15,7 @@ import styles from './{{ tagWithoutPrefix tag }}.styles.js';
  * @slot - The default slot.
  * @slot example - An example slot.
  *
- * @csspart base - The component's base wrapper.
+ * @csspart {{ tagWithoutPrefix tag }} - The component's outer wrapper.
  *
  * @cssproperty --example - An example CSS custom property.
  */
@@ -32,7 +32,7 @@ export default class {{ properCase tag }} extends WebAwesomeElement {
   }
 
   render() {
-    return html` <slot></slot> `;
+    return html` <slot part="{{ tagWithoutPrefix tag }}"></slot> `;
   }
 }
 
--- a/packages/webawesome/custom-elements-manifest.js
+++ b/packages/webawesome/custom-elements-manifest.js
@@ -135,6 +135,22 @@ export default {
       },
     },
     {
+      // Flag `base` in the manifest so editors and other CEM consumers see the deprecation, not just the docs.
+      name: 'wa-deprecate-base-part',
+      packageLinkPhase({ customElementsManifest }) {
+        customElementsManifest?.modules?.forEach(mod => {
+          mod.declarations?.forEach(declaration => {
+            declaration.cssParts?.forEach(part => {
+              if (part.name === 'base') {
+                part.deprecated =
+                  'Use the part named after the component instead. This part will be removed in a future major version.';
+              }
+            });
+          });
+        });
+      },
+    },
+    {
       name: 'wa-translate-module-paths',
       packageLinkPhase({ customElementsManifest }) {
         customElementsManifest?.modules?.forEach(mod => {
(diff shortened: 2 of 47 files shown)


talbs:
@claude, review this PR.

talbs:
@claude, branch updated.

kelseythejackson (dismissed):
Looks good

packages/webawesome/scripts/plop/templates/component/component.hbs
```diff
@@ -15,7 +15,7 @@ import styles from './{{ tagWithoutPrefix tag }}.styles.js';
  * @slot - The default slot.
  * @slot example - An example slot.
  *
- * @csspart base - The component's base wrapper.
+ * @csspart {{ tagWithoutPrefix tag }} - The component's outer wrapper.
```
claviska: It's probably best to hard code the names here.

packages/webawesome/scripts/plop/templates/component/component.hbs
```diff
@@ -32,7 +32,7 @@ export default class {{ properCase tag }} extends WebAwesomeElement {
   }
 
   render() {
-    return html` <slot></slot> `;
+    return html` <slot part="{{ tagWithoutPrefix tag }}"></slot> `;
```
claviska: Here too.

talbs:
Okay, gang. After getting a bit more clarity, the scope of this changed.

The first pass named a part after every component. But @claviska's philosophy of **let the host do the work, and wrap in a part only when you have to** changed that.

### What's the New Outcome/Diff?
- **`accordion` exposes no part now.** It's a bare `<slot>`, so you style `wa-accordion { ... }`. Same for `card`, `dropdown`, `tag`.
- **Nothing goes on a `<slot>`.** `button-group` and `tab-panel` keep their legacy `base` and gain nothing (until a WA4 clean up)
- Components that had preexisting `base` parts (26 free, 6 Pro) now also have a component-based part name alongside them. Only these components have this addition.
- New components scaffold with a bare slot. No `base`, no wrapper part.

### What Happens to `base`?
- `base` sticks around but stops hiding. 
- It's flagged `deprecated` in `custom-elements.json` so editors and `llms.txt` pick it up
- its description points at the replacement (`Deprecated. Use the `[componentName]` part instead.`)
- `base` and other deprecated parts sort to the bottom of the parts table and are visually muted 
- There's an explainer in `customizing.md` alongside updated Parts recommendations
- There's a `:::deprecated` changelog entry noting the soft deprecation (supported, but no longer recommended) . 
- Docs examples no longer teach `::part(base)`.

---

### One Bug from the Fallout
`dropdown` grabbed its trigger button's base with an exact `[part="base"]` selector. Add a second token and that returns null, so `aria-haspopup` and `aria-expanded` quietly stopped getting set. This has been fixed.

claviska:
> Deprecating it is a v4 job, removing it a v5 one.

Looks like we've deprecated them now, which is preferable. It's a trivial fix for users, so risk is low. I've opened #2655 to track the removal in the 4.0 milestone.

claviska (approved):
Excellent. I've wanted to decommission the `base` part pattern for a long time. 😅

talbs:
> Looks like we've deprecated them now, which is preferable. It's a trivial fix for users, so risk is low. I've opened https://github.com/shoelace-style/webawesome/issues/2655 to track the removal in the 4.0 milestone.

Soft deprecated - `base` is still supported, but we're demoting it in favor of the component-based part names in Docs and getting the word out about it being dropped in a future major version.

PR #2646, opened 2026-07-16
Checkbox: Render the Documented `checked-icon` CSS Part

talbs:
`<wa-checkbox>` has always documented a `checked-icon` CSS part — it's in the JSDoc, it's in the CEM, and `<wa-radio>` uses that same name. 

But a checked checkbox was actually rendering that icon under `check-icon`, a part name that shows up nowhere in the docs. Follow the docs, write `::part(checked-icon)`, get nothing.

### The Fix 
- The part name comes from `iconState`, which was `'check'` for the checked state. 
- It's now `'checked'`, so the rendered part is `checked-icon`. 
- The icon glyph itself (`iconName`) didn't change: only the part label was wrong, and only for the checked state.
- Also updated the test that was still asserting the old `check-icon`
- Added a changelog line

FYI - Found during work and review on #2621.


omnifides (approved):
Works great!

PR #2647, opened 2026-07-16
Improve Utilities Documentation

talbs:
A House Rules pass is applied over all 17 CSS Utility Docs, following the same playbook as the component-docs passes (Actions, Forms, Feedback, Media). 

### Conventions
- Singularized heading drift: `Sizing` → `Size` on `flank` and `grid`.
- Named the utility-exposed custom properties as such: `--flank-size`, `--content-percentage`, `--min-column-size`. `text`'s description now calls the theme values design tokens, not custom properties.

### Examples
- `color`'s bare variant-class list is now presented in a table with more details.
- Each section keeps its in-context live example and links out to the canonical reference.

### Redundancy
- Trimmed the repeated `gap`, `align-items`, and `border-radius` class lists on the layout pages. They duplicated the canonical `gap`/`align-items`/`rounding` pages (full tables, previews, the works) and had drifted: `flank`, `grid`, and `split` stopped at `3xl` while `stack` and `cluster` listed `5xl`. 

### Layout Consistency
- `native` and `prose` moved from `layout: page-outline` to `layout: docs`, matching the other 15 utilities. Adds the docs section nav alongside the outline.

### Bug Fixes
- Fixed a broken `{ .example }` fence on `color` (the stray spaces stopped it rendering live)
- Fixed a `::: info` callout on `text` that wasn't parsing.
- `prose` used `wa-align-items-flex-start` in two examples. That class doesn't exist. Fixed to use the real one - `wa-align-items-start`
- Fixed some typos: `native` "numeric date" → "numeric data", `prose` "adopt" → "adopts".

PR #2648, opened 2026-07-16
Improve Tokens Documentation

talbs:
This work applies the same House Rules pass as https://github.com/shoelace-style/webawesome/pull/2647, this time over the 8 design-token docs.

### Layout Consistency
- Unified all 8 pages on `layout: docs`, dropping the redundant `hasOutline: true` and `layout: page-outline`. 
- Every token page now gets the same outline and section nav instead of three different setups.

### Descriptions
- Rewrote the front-matter descriptions to neutral reference voice
- `space`'s was also just broken ("Lock down consistent spacing Web Awesome's space tokens") and now reads like a sentence.

### Bug Fixes
- fixed a stray-quote typo in the variant-hue class example
- `component-groups` had two Button rows, both carrying `id="token-wa-panel-border-style"`, copy-pasted from Panels. Now `token-wa-button-transform-hover` and `-active`

### Links and Copy
- Converted the vestigial `?active_tab=…` cross-links to site-relative `/docs/tokens/…` paths. Nothing consumes that query param anymore, and each token has its own URL
- Dropped "We recommend" from the text-color contrast note

### DIFF
--- a/packages/webawesome/docs/docs/tokens/borders.md
+++ b/packages/webawesome/docs/docs/tokens/borders.md
@@ -9,7 +9,7 @@ use-cases:
   - border width
   - border style
   - border color
-hasOutline: true
+layout: docs
 ---
 
 <style>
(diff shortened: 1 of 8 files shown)


omnifides (approved):
All good! The descriptions are _much_ better. 💪

PR #2650, opened 2026-07-20
OTP: Improve Docs + Apply House Rules

talbs:
Applies the `.house-rules` Docs Rubric to the new `<wa-otp-input>` from https://github.com/shoelace-style/webawesome/pull/2584.

### Structure Changes
- Renamed sections to the shared lexicon (`Labels` → `Label`, `Sizes` → `Size`, `Prefilling a Value` → `Initial Value`, `Custom Styles` → `Customizing`)
- Reordered simple → complex sections, so the plain attribute demos sit up top and the script-driven ones land last
- Cut three examples that weren't earning their place: 
  - the interactive length slider (two static fields show it better)
  - the URL-param prefill script (rendered an empty field on the Docs site)
  - the `Form Submission` / `Form Attribute` sections (no other form control documents these and API table and form-controls already cover them).

### Accuracy Fixes
- `autosubmit` shipped in the component, but the docs never mentioned it. The old auto-submit section hand-rolled the behavior around a commented-out `requestSubmit()`. It's a real section now, and it notes that `wa-complete` is cancelable
- The paste demo also had a dead icon: it used `slot="prefix"`, but `wa-button`'s slot is `start`, so the clipboard icon never rendered. That demo now uses `<wa-copy-button>` with a slotted trigger, which killed the custom clipboard script too

### Docs Additions
- A `type` option table with the default badge, per the rubric
- A WebOTP example for the Android side of SMS autofill (`autocomplete="one-time-code"` handles iOS)
- A `user-invalid` styling example under Customizing
- An `Accessibility Considerations` section covering the `single-hidden-input` model and the keyboard map

### JSDoc & Changelog Edits
- `@summary` reworded task-first, since it renders as the docs page intro
- Added the missing `@event wa-clear`.
- Changelog entry tightened

### DIFF
--- a/packages/webawesome/src/components/otp-input/otp-input.ts
+++ b/packages/webawesome/src/components/otp-input/otp-input.ts
@@ -16,7 +16,8 @@ import sizeStyles from '../../styles/component/size.styles.js';
 import styles from './otp-input.styles.js';
 
 /**
- * @summary OTP inputs collect one-time passcodes, PINs, and other fixed-length codes, one character per segment. Use them for SMS verification, two-factor authentication, and invite codes.
+ * @summary OTP inputs collect one-time passcodes, PINs, and other fixed-length codes, one character per segment.
+ * Use them for SMS verification, two-factor authentication, and invite codes.
  * @documentation https://webawesome.com/docs/components/otp-input
  * @status experimental
  * @since 3.10
--- a/packages/webawesome/src/components/otp-input/otp-input.ts
+++ b/packages/webawesome/src/components/otp-input/otp-input.ts
@@ -16,7 +16,7 @@ import sizeStyles from '../../styles/component/size.styles.js';
 import styles from './otp-input.styles.js';
 
 /**
- * @summary A form-associated OTP/passcode input that displays a fixed number of character segments.
+ * @summary OTP inputs collect one-time passcodes, PINs, and other fixed-length codes, one character per segment. Use them for SMS verification, two-factor authentication, and invite codes.
  * @documentation https://webawesome.com/docs/components/otp-input
  * @status experimental
  * @since 3.10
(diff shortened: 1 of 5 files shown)


talbs:
@claude review this PR.

kelseythejackson (approved):
all good here

talbs:
@claude, review this PR.

packages/webawesome/src/components/otp-input/otp-input.ts
```diff
@@ -328,15 +330,10 @@ export default class WaOtpInput extends WebAwesomeFormAssociatedElement {
...
-        e.preventDefault();
-        this.setCaretIndex(this._activeIndex + 1);
-      } else if (e.shiftKey && this._activeIndex > 0) {
-        e.preventDefault();
-        this.setCaretIndex(this._activeIndex - 1);
-      }
-      // else: Tab/Shift-Tab at boundary → propagate out of component
+    // Tab is deliberately not handled — it moves focus to the next control like any single input;
+    // arrow keys handle movement between segments.
+    if (e.key === 'Enter') {
```
claviska, line 335: Small pre-existing nit: we most commonly use `event.key` and `event.target` instead of `e`. Might be something to consider for house rules. Obviously not a blocker, more of a consistency thing :)

claviska (approved):
This brings more consistency to the OTP input. Docs are looking great too!

packages/webawesome/src/components/otp-input/otp-input.ts
talbs: Pushed a quick commit (https://github.com/shoelace-style/webawesome/pull/2584/changes/17ca84acecc347ba5b6ac6f90cac99e12d7655fc) to cover this post-merge.

PR #2651, opened 2026-07-20
OTP: Keyboard, RTL, and Autofill Fixes

talbs:
(Currently built on top of/based on https://github.com/shoelace-style/webawesome/pull/2650...)

This work includes behavior fixes for the `<wa-otp-input>` component added in https://github.com/shoelace-style/webawesome/pull/2584:

### Keyboard Interactions
- `Enter` now submits the containing form through the shared `submitOnEnter` helper, same as `wa-input`.  Previously, nothing happened because the hidden input wasn't the form owner, and native implicit submission never fired.
- `Tab` no longer walks the segments; arrows already handle movement within the field, and `Tab` should leave it like any single input. **That reverses a deliberate choice from the original PR, so push back if you disagree**, but a keyboard user landing on an empty six-segment field was pressing `Tab` six times to escape it.

### WebKit-Based Fix
`Backspace` in a `readonly` field navigated the page back. The readonly branch blocked mutating keys without `preventDefault()`, and WebKit reads an unhandled `Backspace` in a non-editable context as the history-back gesture. 

This also broke the test suite, which had zero coverage in Safari as a result. This is fixed, and the suite now runs 178/178 on all three browsers.

### RTL Support
Segments stay left-to-right under `dir="rtl"`. Codes read the same in any locale, but the flex row was flipping while the string indexes weren't, so typing filled right-to-left. One `direction: ltr` on `.segments`.

### Autofill Updates
Dropped `maxlength` from the hidden input. Autofill isn't a paste event, so a separator-laden value could truncate before the filter ran. `handleInput` already filters and slices, and `minlength` stays to drive `tooShort`.

### Visual Polish
Hid the cursor caret over a filled segment. A filled active segment is in replace mode, and the ring already marks it; a caret drawn through the character is confusing and unfinished. 

### API Syncing
- Renamed the `segment-separator` part to `segment-literal` to match `time-input`'s name for the same thing
- switched the `--segment-*` custom properties to bracketed JSDoc defaults so they render in the API table.

### DIFF
--- a/packages/webawesome/docs/docs/components/otp-input.md
+++ b/packages/webawesome/docs/docs/components/otp-input.md
@@ -319,9 +319,10 @@ The component uses a single visually hidden `<input>` as the focus and form targ
 
 Keyboard interaction follows the single-input model:
 
-| Key                                   | Behavior                                                             |
-| ------------------------------------- | -------------------------------------------------------------------- |
-| <kbd>←</kbd> <kbd>→</kbd>             | Move between segments                                                |
-| <kbd>Tab</kbd> / <kbd>Shift+Tab</kbd> | Move between segments; leaves the field at the first or last segment |
-| <kbd>Backspace</kbd>                  | Clears the current segment and moves back (no character shift)       |
-| <kbd>Delete</kbd>                     | Clears the current segment without moving                            |
+| Key                       | Behavior                                                       |
+| ------------------------- | -------------------------------------------------------------- |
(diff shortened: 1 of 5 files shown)


talbs:
@claude, review this PR.

claviska:
@kelseythejackson it doesn't look like this was reviewed yet. Moving forward, please refrain from merging other folks' PRs until all of their requested reviews are in :)

Issue #2634, opened 2026-07-13
Site is permanently scroll-locked when a cookie-notice blocker hides the consent dialog


mekedron:
### Describe the bug

On https://webawesome.com/, the cookie consent notice is a **modal** `wa-dialog#cookie-consent-dialog`. Cookie-notice blockers — "I don't care about cookies", uBlock Origin's *Cookie notices* list, AdGuard, Brave — hide it with a cosmetic filter (`display: none !important`). The dialog stays `open`, so `wa-scroll-lock` stays on `<html>` and `body { overflow: hidden }` sticks.

The notice disappears. The scroll lock does not. **The whole site becomes unscrollable, with nothing on screen to explain why.**

There's also no way out of it. The dialog can't be dismissed programmatically — the site cancels `wa-hide` (`defaultPrevented: true`), presumably to force a choice between *Accept All* / *Necessary Only* / *Customize*. But those buttons are inside the element the extension just hid, so they're unreachable, no choice is ever recorded, no consent cookie is written — and the lock comes back on **every** page load.

### To Reproduce

1. Install [I don't care about cookies](https://chromewebstore.google.com/detail/i-dont-care-about-cookies/fihnjjcciajhdojfnbdddfaoknhalnja) (or enable uBlock Origin's *Annoyances → Cookie notices* list).
2. Go to https://webawesome.com/ with no consent cookie stored.
3. Try to scroll.

The page doesn't scroll.

### Browser / OS

- OS: macOS
- Browser: Chrome 150
- Web Awesome: 3.10.0 (`data-version` on `<html>` at webawesome.com)
- Extension: I don't care about cookies 3.5.1

### Additional information

State on a normal page load with the extension active:

```js
document.documentElement.className   // "... wa-scroll-lock ... idc0_350"   ← locked; idc0_350 = extension marker
getComputedStyle(document.body).overflow                       // "hidden"

const d = document.getElementById('cookie-consent-dialog');
d.open                                                         // true
getComputedStyle(d).display                                    // "none"    ← hidden by the extension
```

An open modal that isn't rendered, still holding a document-wide scroll lock.

**Impact.** "I don't care about cookies" alone is past a million installs (800,000 users on the Chrome Web Store, 236,146 on Firefox, 290,000+ on Edge), and the other blockers listed above hide banners the same way — so the affected population is larger still. For all of them the site silently refuses to scroll, and the symptom reads as *"Web Awesome's dialog component breaks page scrolling"*.

[...]

PR #2653, opened 2026-07-21
Add RenderedWatcher to dialog and drawer

claviska:
Adds a `RenderedWatcher` class that detects when an element or its ancestor is hidden through, e.g. `display: none` which many ad/cookie banner blockers do. This prevents the dialogue from going into a modal state that cannot be escaped, as described in #2634.

Regression tests were added to dialog and drawer since they both have a shared history and were both susceptible to the same problem.

While this isn't technically a bug in the components, we decided to add this logic so end users with ad blockers don't get stranded when a dialog or drawer is "blocked" by such an extension.

@talbs Since you were able to reproduce this, would you mind testing this branch out when you get a chance?


talbs (dismissed):
I tested locally using the Chrome Extension the reporting user mentioned alongside some VPNing and **this approach works** - our Home view is no longer gated by the visually hidden modal.

- I can scroll down the view
- I can interact with view elements
- Disabling the extension brings back the Cookie Prompt/dialogue

PR #2656, opened 2026-07-22
Document Mechanical Code Conventions in `CLAUDE.md`

talbs:
This work adds a `## Code Conventions` section to `packages/webawesome/CLAUDE.md` capturing the mechanical, easy-to-miss conventions a contributor (or coding agent) tends to get wrong. 

It grew out of [#2650](https://github.com/shoelace-style/webawesome/pull/2650), where the `event`-not-`e` handler-param preference came up in review.

Rather than leave it as one-off feedback, this writes it down alongside the other conventions at the same level.

The section leads with a pointer to the published [contributing guide](https://webawesome.com/docs/resources/contributing/) as the canonical convention doc, then lists only rules that are genuinely undocumented there and verifiable across the codebase:

### What's Documented
- Event handler parameters are named `event`, not `e`
- Event handlers are named `handle<Subject>` (`handleInput`, `handleClearClick`)
- Relative imports end in `.js` (NodeNext ESM)
- Custom events are one `class Wa<Name>Event extends Event` per file in `src/events/`, dispatched with `new Wa…Event()`
- Multi-word properties declare an explicit kebab `attribute:`, since Lit lowercases attribute names.

### DIFF
--- a/packages/webawesome/CLAUDE.md
+++ b/packages/webawesome/CLAUDE.md
@@ -49,7 +49,13 @@ Instantiate in the class body (not constructor):
 
 ## Code Conventions
 
-- Event handler parameters are named `event`, not `e` — read from it directly (`event.key`, `event.target`, `event.preventDefault()`). Keeps handlers consistent across the codebase.
+The published [contributing guide](docs/docs/resources/contributing.md) is the canonical convention doc (component structure, BEM class names, `with-*`/`without-*` boolean props, event naming). The rules below are the mechanical ones most often gotten wrong:
+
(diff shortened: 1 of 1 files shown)


talbs:
@claviska, here's a follow-up to address that good comment you left on https://github.com/shoelace-style/webawesome/pull/2650#discussion_r3625915356. Instead of House Rules, this felt most appropriate adding to `Claude.md`.

I also added a few more things that felt in the same ballpark without duping Contributing-level docs. Shout if you want any edits.

claviska (approved):
Good call and great initiative.

PR #2661, opened 2026-07-23
Initial HEXA value bug

claviska:
This builds off PR https://github.com/shoelace-style/webawesome/pull/2588 to solve https://github.com/shoelace-style/webawesome/issues/2550.

The PR fixes a bug where `<wa-color-picker>` would lose the alpha channel of an initial HEXA/RGBA/HSLA value, causing the form to submit a plain hex value. The color picker now reads its `opacity`, `format`, and `uppercase` attributes before the initial value sync so the value is never re-serialized with default settings, and no longer resets alpha to 100 on first render.

Tested using the original test case found in #2550.


kelseythejackson (approved):
All good

PR #2667, opened 2026-07-24
fix seo metadata stuff

KonnorRogers:
Also worth noting `og:description` is undefined for all our component docs since that comes later in the build. That is displayed on a page using `{{ component.summary }}`, it never sets a `description` or `ogDescription`, so things like:

https://webawesome.com/docs/components/button

have no `og:description` as a result.

I'll leave it to our SEO expert @talbs if we should fix this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2733,9 +2733,9 @@
       }
     },
     "node_modules/@shoelace-style/localize": {
-      "version": "3.2.2",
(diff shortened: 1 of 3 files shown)


talbs (dismissed):
OG image fix looks great! Approved!

Not sure if you need the `package-lock.json` diff, but you know best on that.

---

And yeah, we should have `og:description` point to the `component.summary` - that was my original intent. Thanks if you fix that up too.

KonnorRogers:
@talbs we do because localize was upgraded

talbs (approved):

PR #2669, opened 2026-07-27
Update changelog; fixes #2637

### LINKED ISSUE #2637: wa-date-input's calendar or wa-time-input closes an ancestor wa-dialog when dismissed (Escape / date pick / outside click)
### Describe the bug
Hello! I’m new to WebAwesome and GitHub, so I apologize if I’ve done anything incorrectly.

When a `<wa-date-input>` or  `<wa-time-input>` is placed inside a `<wa-dialog>` and its calendar popup is open, dismissing the calendar — either by pressing Escape or by picking a date (which closes the calendar automatically once the selection is complete) — also closes the parent `<wa-dialog>`, removing it from the DOM.

Per the "Dismissible Overlays" convention described in CONTRIBUTING.md, only the topmost dismissible should respond to a dismiss gesture. This breaks that contract: the calendar is the topmost overlay, but closing it also closes its ancestor dialog.

Separately, dismissing the calendar with Escape also throws a stack overflow (`RangeError: Maximum call stack size exceeded` in Chromium, `InternalError: too much recursion` in Firefox). I traced this to `wa-date-input`'s own Escape handler, which dispatches a synthetic, bubbling + composed `keydown` Escape event onto its own calendar before hiding:

  ```js
  this.calendar?.dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true, composed: true }));
  this.hide();
```

Because the event is composed and bubbling, it re-reaches the same document-level Escape listener that dispatched it, re-entering the handler recursively. I could reproduce this crash reliably in a real (non-sandboxed) browser tab — both an automated Chromium/Playwright run and a manual Firefox test — but not inside the CodePen iframe linked below, likely due to a different stack depth limit or error handling in that sandboxed context. The "closes the parent dialog" symptom itself reproduces fine on CodePen regardless.

### To Reproduce

1. Open the demo below.
2. Click "Open dialog".
3. Click the date field to open the calendar popup.
4. Either: press Escape, OR click any day in the calendar, OR click anywhere else inside the dialog (outside the popup).
5. Observe: the calendar closes, but the  `<wa-dialog> ` closes too and is removed from the DOM. For the Escape case, also check the console for Maximum call stack size exceeded.

### Demo
I've made a codepen to replicate the error:
[Open CodePen](https://codepen.io/editor/seveninformatica/pen/019f60a1-437b-78b5-8104-3d4985c65134?console=true&file=%2Findex.html&orientation=left&panel=files&show=split)

### Screenshots
Before (calendar open inside the dialog) / after pressing Escape (dialog and everything else gone):

Image
Image

### Browser / OS
- OS: Windows 11
- Browser: tested in Chrome/Chromium and Firefox, both affected
- Web Awesome version: 3.10.0

### Additional information
Reproduced independently in three environments: an automated Playwright/Chromium harness, a manual Firefox test against a real local app, and the CodePen demo linked above (Chrome and Firefox). The "dialog closes" symptom is consistent across all three; the stack-overflow crash specifically was observed in the first two (non-sandboxed) environments but not inside CodePen's iframe.

claviska:
Updates the changelog for pro [companion PR 272](https://github.com/shoelace-style/webawesome-pro/pull/272).

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -58,6 +58,7 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 - Fixed a bug in `<wa-dialog>` and `<wa-drawer>` that left the page permanently scroll locked and inert when third-party CSS, e.g. from ad blockers, hides an open dialog or drawer with `display: none`
 - Fixed `<wa-page>` documenting `skip-links` and `skip-link` CSS parts that don't render; replaced them with the `skip-to-content` part it actually exposes [pr:2633]
 - Fixed `x-label` and `y-label` attributes not working on `<wa-chart>` and its variants
+- Fixed a bug in `<wa-date-input>` where pressing [[Escape]] to dismiss the open calendar threw a "Maximum call stack size exceeded" error [issue:2637]
 - Fixed a bug in `<wa-color-picker>` where an initial value with alpha, e.g. `#f5a62315` with `opacity`, would lose its alpha channel on load [issue:2550]
 
 :::


talbs (approved):
ch-ch-changelog ch-ch-changes look good!

PR #2672, opened 2026-07-27
canvas to svg testing

KonnorRogers:


claviska:
I'm not opposed to this as a hook for consumers to use, but we'll need to document it and add a test + changelog entry first.

Based on the [discussion here](https://github.com/shoelace-style/webawesome/discussions/2668), I wonder if it makes more sense to adapt the dependency to generate SVG instead of canvas.

PR #2674, opened 2026-07-28
fix(tree): don't throw on Enter/Space when focus is inside a tree item

### LINKED ISSUE #2673: wa-tree: Enter/Space throws "Cannot read properties of undefined (reading 'disabled')" when focus is inside a tree item
### Describe the bug

`WaTree.handleKeyDown()` resolves the focused node with `items.findIndex(item => item.matches(':focus'))`. When focus is **inside** a tree item rather than **on** it — e.g. on a link or a button slotted into the item — no `<wa-tree-item>` matches `:focus` (it only matches `:focus-within`), so `findIndex()` returns `-1` and `activeItem` is `undefined`.

Every branch of the handler guards for that (`if (!activeItem || activeItem.disabled || …)`) except the Enter/Space branch:

https://github.com/shoelace-style/webawesome/blob/next/packages/webawesome/src/components/tree/tree.ts#L298-L303

```ts
} else if (event.key === 'Enter' || event.key === ' ') {
  // Selects the focused node.
  if (!activeItem.disabled) {   // 💥 activeItem is undefined
    this.selectItem(activeItem);
  }
}
```

so the keypress throws:

```
TypeError: Cannot read properties of undefined (reading 'disabled')
    at WaTree.handleKeyDown (chunk.7H62LFN3.js:228:25)
    at L.handleEvent (chunk.BKE5EYM3.js:230:48)
```

There's a second, related symptom: `event.preventDefault()` runs [before](https://github.com/shoelace-style/webawesome/blob/next/packages/webawesome/src/components/tree/tree.ts#L252) `activeItem` is even resolved, so the keystroke is swallowed too. Even with the crash fixed, a link inside a tree item can never be activated with Enter — the tree cancels the event without having a node to act on.

### To Reproduce

1. Load the demo below.
2. Press <kbd>Tab</kbd> to move focus to the link inside the first tree item (clicking it works too).
3. Press <kbd>Enter</kbd> (or <kbd>Space</kbd>).
4. `TypeError: Cannot read properties of undefined (reading 'disabled')` is thrown, and the link is not followed.

### Demo

```html
<link rel="stylesheet" href="https://ka-f.webawesome.com/webawesome@3.10.0/styles/webawesome.css" />
<script type="module" src="https://ka-f.webawesome.com/webawesome@3.10.0/webawesome.loader.js"></script>

<wa-tree>
  <wa-tree-item>
    <a href="https://webawesome.com/">Tab to me, then press Enter</a>
  </wa-tree-item>
  <wa-tree-item>Node 2</wa-tree-item>
</wa-tree>
```

It also reproduces as a unit test in this repo — this fails on `next` and passes with the guard added:

```ts
it('should not throw when Enter is pressed with focus inside a tree item', async () => {
  const tree = await fixture<WaTree>(html`
    <wa-tree>
      <wa-tree-item><a href="#node-1">Node 1</a></wa-tree-item>
      <wa-tree-item>Node 2</wa-tree-item>
    </wa-tree>
  `);

  tree.querySelector('a')!.focus();
  await tree.updateComplete;

  await sendKeys({ press: 'Enter' }); // Uncaught TypeError: Cannot read properties of undefined (reading 'disabled')
});
```

### Browser / OS

Not browser-specific — it's a plain logic error, and there is no `activeItem` in any browser when focus sits on slotted content.

- Reproduced locally in Chromium 3.10.0 / `next` via the repo's test runner (macOS).
- Observed in production on Windows 10 / Chrome 150, Web Awesome 3.10.0.

### Additional information

We hit this in a real app: an application sidebar built as `<wa-tree selection="leaf">` where every `<wa-tree-item>` wraps an `<a>` for the nav destination. Users keyboard-navigating the sidebar generate a steady trickle of these `TypeError`s in our error reporting, and their <kbd>Enter</kbd> press does nothing because of the `preventDefault()` above.

Slotting interactive content into tree items looks intended: `handleKeyDown()` already skips keydowns originating in `input`/`textarea` so that "nested form controls that exist inside tree items" aren't hijacked, and #1552 covers anchors inside tree items. Links and buttons just aren't covered by that escape hatch.

Suggested fix — bail out for Enter/Space when there is no active item, before `preventDefault()`, so the focused element keeps its default behavior, plus the missing `activeItem &&` guard in the branch itself. Happy to open a PR (tests included); one is on its way.

gregjotau:
Fixes #2673.

### The bug

`WaTree.handleKeyDown()` resolves the focused node with `items.findIndex(item => item.matches(':focus'))`. When focus is **inside** a tree item rather than **on** it — e.g. on a link or a button slotted into the item — no `<wa-tree-item>` matches `:focus` (it only matches `:focus-within`), `findIndex()` returns `-1`, and `activeItem` is `undefined`.

Every branch of the handler guards for that (`if (!activeItem || activeItem.disabled || …)`) except Enter/Space, so pressing either threw:

```
TypeError: Cannot read properties of undefined (reading 'disabled')
    at WaTree.handleKeyDown (chunk.7H62LFN3.js:228:25)
```

`event.preventDefault()` also ran *before* `activeItem` was resolved, so the keypress was swallowed as well — a link inside a tree item could never be activated with <kbd>Enter</kbd>, error or not.

Repro (also included as a test):

```html
<wa-tree>
  <wa-tree-item><a href="https://webawesome.com/">Tab to me, then press Enter</a></wa-tree-item>
  <wa-tree-item>Node 2</wa-tree-item>
</wa-tree>
```

### The change

`packages/webawesome/src/components/tree/tree.ts`

- Resolve `activeItem` before calling `event.preventDefault()`, and return early for <kbd>Enter</kbd>/<kbd>Space</kbd> when there is no active item. Nothing can be selected in that case, so the event is left alone and the focused element keeps its default behavior (the link is followed, the button is clicked).
- Add the missing `activeItem &&` guard in the Enter/Space branch, matching the ArrowLeft/ArrowRight branches. Redundant after the early return, but it keeps the branch type-safe and consistent with its neighbors.

Arrow keys, <kbd>Home</kbd> and <kbd>End</kbd> are deliberately untouched: they still `preventDefault()` and still fall back to `focusItemAt(-1 + 1)` / `clamp()` when nothing is focused, which is how focus enters the tree today.

This lines up with the existing escape hatch in the same handler for `input` and `textarea` ("prevents the component from hijacking nested form controls that exist inside tree items") — links and buttons simply weren't covered by it.

### Tests

`packages/webawesome/src/components/tree/tree.test.ts` — new `keyboard navigation > when focus is inside a tree item instead of on it` block:

- <kbd>Enter</kbd> with focus on a slotted link doesn't throw and selects nothing
- <kbd>Space</kbd> with focus on a slotted link doesn't throw and selects nothing
- <kbd>Enter</kbd> reaches the link, i.e. the click isn't cancelled by the tree

All three fail on `next` (the first two as `Uncaught TypeError: Cannot read properties of undefined (reading 'disabled')`) and pass with the fix.

Verified locally in Chromium, both rendering modes: `CSR_ONLY` across the whole suite (2458 passing, 0 failures) and `SSR_ONLY` for `tree.test.ts` (44 passing). `prettier --check` and `cspell` are clean on the changed files.

### Changelog

Entry added under **Unreleased → Fixed** referencing #2673.

### Context

We hit this in production in an app sidebar built as `<wa-tree selection="leaf">` where each `<wa-tree-item>` wraps an `<a>`: users navigating the sidebar by keyboard generated a steady trickle of these `TypeError`s, and their <kbd>Enter</kbd> press did nothing.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -63,6 +63,7 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 - Fixed a bug in `<wa-color-picker>` where an initial value with alpha, e.g. `#f5a62315` with `opacity`, would lose its alpha channel on load [issue:2550]
 - Fixed a bug in `<wa-color-picker>` where selecting a swatch with the keyboard didn't emit `change` and `input` events [pr:2665]
 - Fixed a bug in `<wa-color-picker>` where an initial value with alpha but without `opacity` would render the trigger as transparent even though the value was opaque [pr:2665]
+- Fixed a bug in `<wa-tree>` where pressing [[Enter]] or [[Space]] while focus was inside a tree item, e.g. on a link or a button slotted into it, threw a "Cannot read properties of undefined" error and swallowed the keypress instead of letting the focused element handle it [issue:2673]
 
 :::
 
--- a/packages/webawesome/src/components/tree/tree.test.ts
+++ b/packages/webawesome/src/components/tree/tree.test.ts
@@ -468,6 +468,66 @@ describe('<wa-tree>', () => {
             });
           });
         });
+
+        describe('when focus is inside a tree item instead of on it', () => {
+          const uncaughtErrors: string[] = [];
+          const collectUncaughtError = (event: ErrorEvent) => {
+            event.preventDefault();
+            uncaughtErrors.push(event.message);
+          };
+
+          beforeEach(() => {
+            uncaughtErrors.length = 0;
+            window.addEventListener('error', collectUncaughtError);
+          });
+
+          afterEach(() => {
+            window.removeEventListener('error', collectUncaughtError);
+          });
+
+          async function fixtureWithLink() {
+            const tree = await fixture<WaTree>(html`
+              <wa-tree>
+                <wa-tree-item><a href="#node-1">Node 1</a></wa-tree-item>
+                <wa-tree-item>Node 2</wa-tree-item>
+              </wa-tree>
+            `);
+            const link = tree.querySelector('a')!;
+
+            link.focus();
+            await tree.updateComplete;
+
+            return { tree, link };
+          }
+
+          it('should not throw when Enter is pressed', async () => {
+            const { tree } = await fixtureWithLink();
+
+            await sendKeys({ press: 'Enter' });
+
+            expect(uncaughtErrors).to.be.empty;
+            expect(tree.selectedItems.length).to.eq(0);
+          });
+
+          it('should not throw when Space is pressed', async () => {
+            const { tree } = await fixtureWithLink();
+
+            await sendKeys({ press: ' ' });
+
+            expect(uncaughtErrors).to.be.empty;
+            expect(tree.selectedItems.length).to.eq(0);
+          });
+
+          it('should let the focused element handle Enter', async () => {
+            const { link } = await fixtureWithLink();
+            const clickSpy = sinon.spy((event: MouseEvent) => event.preventDefault());
+            link.addEventListener('click', clickSpy);
+
+            await sendKeys({ press: 'Enter' });
+
+            expect(clickSpy).to.have.been.calledOnce;
+          });
+        });
       });
 
       describe('events', () => {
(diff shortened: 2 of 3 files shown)


claviska (approved):
Thanks for the bug report and PR. This looks good to me!

PR #2677, opened 2026-07-28
Fix the fixed icons

### LINKED ISSUE #2636: wa-icon fill is forcefully set to currentColor
### Describe the bug
The changes between 3.7.0 and 3.8.0 reintroduced a `fill: currentColor` on wa-icon, which cause the same issue as was reported in https://github.com/shoelace-style/webawesome/issues/1733.
The changes can be seen in this commit https://github.com/shoelace-style/webawesome/commit/50bdfbc5d1bb7eb98aabbaaa81cb38bce1e924ec#diff-b16bf5a6ad3354c71c0a5325a767574d3477618d14559750a8d555b4fa104ab0.
I couldn't override it using `mutator` as shown in docs either.

### To Reproduce

1. Set up lucide icons in `registerIconLibrary`
2. Add a wa-icon

Actual behavior: Fill is set to currentColor
Expected behavior: Fill is not set

### Demo

N/A

### Screenshots
Same as in #1733

### Browser / OS
 - OS: [e.g. Mac, Windows]
 - Browser: [e.g. Chrome, Firefox, Safari]
 - Browser version: [e.g. 22]

### Additional information
N/A

claviska:
Fixes #2636. This was the same root cause as #1733, which was fixed in #1784 and reintroduced in 3.8.0 so we could, uh, fix it again!

The component stylesheet set `fill: currentColor` on the internal SVG, causing stroke-based libraries like Lucide to render with filled blobs. The fill now gets applied by the default library's mutator instead, and only when the SVG doesn't declare its own.

To prevent having to fix this fix again again 🙃 , I've added regression tests for both.

@calahatras thanks for the [bug report](https://github.com/shoelace-style/webawesome/issues/2636)! Can you verify the PR preview is rendering what you'd expect now?

Also pinging @talbs for his 🦅 👁️  on this one.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -63,7 +63,7 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 - Fixed a bug in `<wa-color-picker>` where an initial value with alpha, e.g. `#f5a62315` with `opacity`, would lose its alpha channel on load [issue:2550]
 - Fixed a bug in `<wa-color-picker>` where selecting a swatch with the keyboard didn't emit `change` and `input` events [pr:2665]
 - Fixed a bug in `<wa-color-picker>` where an initial value with alpha but without `opacity` would render the trigger as transparent even though the value was opaque [pr:2665]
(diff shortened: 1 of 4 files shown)


talbs (dismissed):
Looking good. Getting 🎁🎁🎁.

Nice notes and tests to make sure this doesn't regress again. 🧑‍🍳😘

calahatras:
Verified on my end, thank you! 👏

PR #2680, opened 2026-07-29
Add supporting files for the pro data grid component

claviska:
Pro companion PR: https://github.com/shoelace-style/webawesome-pro/pull/276

PR #2683, opened 2026-07-29
Pagination: Docs Polish

talbs:
This work is a pass over the pagination Docs (following up on https://github.com/shoelace-style/webawesome/pull/2680 where the page first landed). 

**Preview: https://webawesome-git-talbs-a-real-page-turner-font-awesome.vercel.app/docs/components/pagination**

### Prose and Callouts
- The lead example has been moved to the top of the page, allowing users to see the component first rather than reading a description of it.
- All three `:::info` callouts now follow a consistent format: bold lead text followed by a `<br />` line break, similar to about twenty other component pages, and each has been trimmed for brevity.
- Two factual gaps have been addressed: the `boundary-count` now specifies its default value of `1`, and the section title "Styling Paginations" has been changed to "Customizing" to align with related pages.

### Examples
- Multi-control examples now use `wa-stack` for layout instead of relying on `<br />` tags.
- The interactive demos are contained within their own wrapper elements, allowing for scoped queries.
- The Custom Icons section replaces `arrow-left` and `arrow-right` with `backward-step` and `forward-step` to better fit the overall icon set.
- In the `Customizing` demo, the properties `sibling-count="1"` and `flex-wrap: nowrap` are applied to `::part(pages)` to ensure the pill treatment remains in one row at narrower widths.

### Accessibility Considerations
A new section documents important behaviors of the component that were previously unlisted: page changes are announced through a shared live region, focus shifts to the new current page upon activation, `aria-current="page"` is assigned to the active item, `aria-disabled` is applied to boundary controls, and direction-aware icons are automatically flipped for RTL layouts. 

The section concludes with guidance on setting the `label` attribute when multiple pagination controls share the same page. It closes with guidance to set `label` when more than one pagination shares a page.

### DIFF
--- a/packages/webawesome/docs/docs/components/pagination.md
+++ b/packages/webawesome/docs/docs/components/pagination.md
@@ -13,14 +13,15 @@ use-cases:
   - large lists
 ---
 
-Set `total` and `page-size` to generate numbered page buttons with previous and next controls. Use the `label` attribute to give the control an accessible name, which is especially helpful when more than one appears on the same page.
-
 ```html {.example}
 <wa-pagination total="237" page-size="10" page="3" label="Search results"></wa-pagination>
 ```
 
+Set `total` and `page-size` to generate numbered page buttons with previous and next controls. Use the `label` attribute to give the control an accessible name, which is especially helpful when more than one appears on the same page.
+
 :::info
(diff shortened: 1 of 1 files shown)


claviska (approved):
https://www.youtube.com/watch?v=3khH9ih2XJg

PR #2685, opened 2026-07-30
OTP Input: Align Focus Ring with Other Inputs

talbs:
### The Issue

`<wa-otp-input>` was the only form control animating its focus ring's geometry: `outline-width` and `outline-offset` alongside `outline-color`. 

Because the ring lives on the active segment, that animation re-ran on every keystroke as the caret advanced: a 6-digit code meant a dozen grow/shrink animations in about two seconds, with no `prefers-reduced-motion` gate. 

Every other input keeps a full-width transparent outline and fades only the color, a convention set in https://github.com/shoelace-style/webawesome/pull/2625.

### The Fix
- `.segment` now carries `outline: var(--wa-focus-ring-style) var(--wa-focus-ring-width) transparent` with a static `outline-offset`, and transitions `outline-color` only.
- The `contained` appearance's inward offset moved off `:focus-within` onto the base rule, so ring geometry is static per appearance and only the color animates
- Dropped two `aTimeout(200)` waits in the tests that existed to let the offset transition settle

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -29,6 +29,14 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 
 {% include "changelog-email-signup.njk" %}
 
+## Unreleased
+
+:::fixed
+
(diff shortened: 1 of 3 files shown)


claviska (approved):
Great catch. That feels more natural.

kelseythejackson (approved):

PR #2689, opened 2026-07-30
Fix pagination ellipsis color in dark mode

### LINKED ISSUE #2688: Pagination dark mode ... is the wrong color
Image

[Reported via Discord](https://discord.com/channels/739437527907303535/1532499891954258010/1532499891954258010)

happykawayigt:
## Summary
- make the system ellipsis SVG inherit `currentColor`
- add a regression test for the system icon
- fix the pagination ellipsis color in dark mode

Fixes #2688

## Validation
- Chromium icon tests: 142 passed
- Chromium pagination tests: 116 passed
- focused Prettier check
- `git diff --check`
- minimal build with `SKIP_SLOW_STEPS=true`

## Notes
The repository's full local build reaches an existing Windows React-wrapper path generation issue; CI will run the full Linux and browser matrix.


packages/webawesome/src/components/icon/library.system.ts
```diff
@@ -21,7 +21,7 @@ export const icons: { [key: string]: { [key: string]: string } } = {
...
+    ellipsis: `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 640 640"><!--!Font Awesome Free v7.3.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2026 Fonticons, Inc.--><path fill="currentColor" d="M96 320C96 289.1 121.1 264 152 264C182.9 264 208 289.1 208 320C208 350.9 182.9 376 152 376C121.1 376 96 350.9 96 320zM264 320C264 289.1 289.1 264 320 264C350.9 264 376 289.1 376 320C376 350.9 350.9 376 320 376C289.1 376 264 350.9 264 320zM488 264C518.9 264 544 289.1 544 320C544 350.9 518.9 376 488 376C457.1 376 432 350.9 432 320C432 289.1 457.1 264 488 264z"/></svg>`,
```
claviska, line 24: We shouldn't have to hand edit these icons, as they come directly from FA. This _might_ have been fixed in https://github.com/shoelace-style/webawesome/pull/2677 but I'll need to verify.

claviska:
Thanks so much for the PR! While this solved the symptom, the bigger problem was that we've been adding `fill="currentColor"` to the system lib's SVGs, but that attribute isn't present in the raw official SVGs.

The correct solution is to remove that attribute from all the SVGs and apply it in the system icon library's mutator, which I did in PR #2718.

Thanks again!

PR #2690, opened 2026-07-30
Fix fallback navigation toggle header background

### LINKED ISSUE #2666: <wa-page> fallback navigation toggle leaves transparent gap in sticky header
### Describe the bug

On mobile viewports, `<wa-page>` places its built-in fallback navigation toggle before the slotted `header`.

The slotted header receives `background-color: var(--wa-color-surface-default)`, but the internal sticky `[part~='header']` wrapper does not. This leaves the area behind the fallback hamburger transparent.

When page content scrolls beneath the sticky header, it becomes visible through the hamburger area while the rest of the header remains opaque.

### Minimal reproduction

```html
<!doctype html>
<html lang="en" class="wa-cloak">
  <head>
    <meta charset="UTF-8" />

    <link rel="stylesheet" href="/dist/styles/themes/default.css" />
    <link rel="stylesheet" href="/dist/styles/webawesome.css" />
    <script type="module" src="/dist/webawesome.loader.js"></script>
  </head>

  <body>
    <wa-page>
      <header slot="header">
        <strong>Example application</strong>
      </header>

      <nav slot="navigation">
        <a href="#">Navigation item</a>
      </nav>

      <main>
        <h1>Page content</h1>
        <p>Scroll this content beneath the sticky header.</p>
        <div style="min-height: 200vh"></div>
      </main>
    </wa-page>
  </body>
</html>
```

1. Open the page at a width below the default mobile breakpoint.
2. Confirm that the built-in hamburger is visible.
3. Scroll the main content underneath the sticky header.
4. Observe the page content through the area behind the hamburger.

### Expected behavior

The entire sticky header row should have the same surface background, including the area occupied by the built-in navigation toggle.

### Actual behavior

Only the slotted header is opaque. The internal header wrapper and the area containing the fallback toggle remain transparent.

The resulting layout is effectively:

```text
| transparent fallback-toggle area | opaque slotted header |
```

### Current workaround

```css
wa-page::part(header) {
  background-color: var(--wa-color-surface-default);
}
```

Using a custom `data-toggle-nav` button inside the slotted header also avoids the problem because the built-in fallback is removed. However, the built-in fallback should work without requiring custom markup or CSS.

### Environment

- Web Awesome: 3.10.0
- Browser: Chrome
- Theme: Default-derived theme
- Reproduces with the built-in navigation toggle

happykawayigt:
## Summary
- give the full `wa-page` header part the default surface background
- cover the built-in mobile navigation toggle area instead of only the slotted header
- add a regression test for the header part's computed background

Fixes #2666

## Validation
- Chromium page tests: 60 passed
- focused Prettier check
- `git diff --check`
- minimal build with `SKIP_SLOW_STEPS=true`

## Notes
The repository's full local build reaches an existing Windows React-wrapper path generation issue; CI will run the full Linux and browser matrix.


claviska:
Assigning this one to @KonnorRogers for review. If we're good with the change, this just needs a changelog entry.

claviska:
@KonnorRogers this one too, please :D

packages/webawesome/src/components/page/page.styles.ts
```diff
@@ -168,6 +168,7 @@ export default css`
   }
   [part~='header'] {
     top: var(--banner-top);
+    background-color: var(--wa-color-surface-default);
```
KonnorRogers, line 171: I _think_ this is better to put on the slotted button.

KonnorRogers (changes requested):

packages/webawesome/src/components/page/page.styles.ts
KonnorRogers: After looking again...yea this is the right fix. No way around this. We'll have to direct users to set the background color on the `::part(header)`

KonnorRogers (dismissed):

claviska:
Maintainer note: needs a changelog, otherwise ready to 🚢

PR #2694, opened 2026-07-31
Keep the button label accessible while loading

### LINKED ISSUE #2693: `<wa-button>`: loading state removes the label from the accessibility tree
### Describe the bug

While `loading` is set, `<wa-button>` hides its content with `visibility: hidden`:

https://github.com/shoelace-style/webawesome/blob/1eef0e82915ab8c2c468f4db5c46777feb97e8a3/packages/webawesome/src/components/button/button.styles.ts#L312-L321

`visibility: hidden` removes content from the accessibility tree, so the label stops contributing to the button's accessible name. The rendered `<button>` also never gets `aria-busy`. While an action is in flight, a screen reader user therefore hears an unlabelled button, or just "Loading" from the spinner.

Two details make this harder to spot:

- `visibility` is inherited, and slotted content inherits from the slot it is assigned to ([CSS Scoping, flat tree](https://drafts.csswg.org/css-shadow-1/#flattening)). The rule reaches across the shadow boundary into markup the consumer wrote.
- Whether the name survives depends on how the consumer wrote the label. A bare text node keeps its name in Chrome. Wrap that same text in an element, or project it through a wrapper component's slot, and the name is gone.

### To Reproduce

```html
<link rel="stylesheet" href="https://ka-f.webawesome.com/webawesome@3.11.0/styles/themes/default.css">
<script type="module" src="https://ka-f.webawesome.com/webawesome@3.11.0/webawesome.loader.js"></script>

<wa-button loading><span>Save changes</span></wa-button>
<wa-button><span>Save changes</span></wa-button>
```

Read the accessible names from Chrome's accessibility tree (DevTools → Elements → Accessibility, or CDP `Accessibility.getFullAXTree`):

| markup | accessible name |
| --- | --- |
| `<wa-button loading><span>Save changes</span></wa-button>` | `"Loading"` |
| `<wa-button><span>Save changes</span></wa-button>` | `"Save changes"` |

The `"Loading"` comes from the spinner's own `role="progressbar"` label, not from the button. Remove the `<wa-spinner>` from the shadow root and the loading button's name is `""`.

### Additional information

Suggested fix, in two parts:

1. Hide the content with `opacity: 0` instead of `visibility: hidden`. The accessible name computation spec explicitly does not treat `opacity: 0` as hidden ([accname, "Hidden Not Referenced"](https://w3c.github.io/accname/#comp_hidden_not_referenced)), so the label stays in the tree. I measured the button box across 12 cases on 3.7.0 and 3.11.0: the difference is 0.000 px in both dimensions, so the documented promise that "its width stays the same, so adjacent elements don't shift" still holds.
2. Set `aria-busy` on the rendered `<button>`/`<a>` while loading. `<wa-tree-item>` and `<wa-carousel>` already do this.

One behaviour difference worth naming: `opacity: 0` leaves the content hit-testable, which `visibility: hidden` does not. The `<button>` is the click target and `handleClick` already suppresses clicks while loading, so I did not find an effect, but it is a real difference.

`button.test.ts` asserts `to.be.accessible()` for each variant and for `disabled`. The `loading` test only checks that a spinner renders, which is probably why this went unnoticed.

We hit this in Home Assistant, where buttons are commonly wrapped, so the label reaches `<wa-button>` through a second slot and the name is always lost. I am happy to open a PR for both parts.

### Browser / OS

- OS: macOS
- Browser: Chromium 151.0.7922.34
- Web Awesome: reproduced on 3.11.0 and 3.7.0

I measured accessible names in Chrome only. The CSS that causes it is not browser specific.

MindFreeze:
Fixes #2693.

While `loading` is set, `<wa-button>` hid its content with `visibility: hidden`. That removes the content from the accessibility tree, so the button could end up with no accessible name, and nothing reported the busy state.

## What changed

- `button.styles.ts`: hide the content with `opacity: 0` instead of `visibility: hidden`.
- `button.ts`: set `aria-busy` on the rendered `<button>`/`<a>` while loading, matching `<wa-tree-item>` and `<wa-carousel>`.
- `button.test.ts`: three cases in the existing `accessibility` block.
- Changelog entry.

## Why `opacity`

The accessible name computation spec treats `visibility: hidden` as hidden and explicitly does not treat `opacity: 0` as hidden ([accname, "Hidden Not Referenced"](https://w3c.github.io/accname/#comp_hidden_not_referenced)), so the label stays in the tree.

`opacity: 0` also keeps the content in flow, so the documented promise that "its width stays the same, so adjacent elements don't shift" still holds. I measured the host and the internal `.button` box across 12 label shapes on 3.7.0 and 3.11.0: the difference is 0.000 px in both dimensions.

One behaviour difference worth naming: `opacity: 0` leaves the content hit-testable, which `visibility: hidden` does not. The `<button>` is the click target and `handleClick` already returns early while loading, so I did not find an effect, but it is a real difference and I would rather flag it than hide it.

## Verification

`npx web-test-runner --group button` passes in Chromium and Firefox: 188 passed, 6 skipped, both the client-only and SSR-hydrated fixtures.

With the CSS change reverted, `should keep the label in the accessibility tree when loading` fails in both fixtures and both browsers, so the test is a real guard.

Worth knowing: the `to.be.accessible()` axe check does **not** catch this on its own. With the label hidden, the spinner's own `role="progressbar"` label becomes the button's accessible name, so axe still sees a named button. That is probably why this went unnoticed. The direct assertion is the one that matters.

Verified against Chrome's native accessibility tree via CDP `Accessibility.getFullAXTree`, not a test-runner approximation.

| markup | before | after |
| --- | --- | --- |
| `<wa-button loading><span>Save changes</span></wa-button>` | `"Loading"` | `"Save changes Loading"` |
| same, with the spinner removed | `""` | `"Save changes"` |
| `<wa-button loading>Save changes</wa-button>` | `"Save changes Loading"` | `"Save changes Loading"` |

The third row is the case that looks fine today. A bare text node has no element-level computed style, so Chrome's hidden check does not reject it. Wrap the text in an element, or project it through a wrapper component's slot, and the name is gone. Relying on that leniency did not seem safe.

Happy to adjust the approach or drop the changelog entry if you would rather write it yourself.


claviska (changes requested):
Thanks for the PR!

Looks like we should also add `pointer-events: none` to the hidden parts. Unlike `visibility: hidden`, `opacity: 0` content stays clickable, so clicks on the invisible label will fire consumer click listeners before the internal guard stops the event.

packages/webawesome/src/components/button/button.ts
```diff
@@ -328,6 +328,7 @@ export default class WaButton extends WebAwesomeFormAssociatedElement {
         rel=${ifDefined(isLink && this.rel ? this.rel : undefined)}
         role=${ifDefined(isLink ? undefined : 'button')}
         aria-disabled=${ifDefined(isLink && this.disabled ? 'true' : undefined)}
+        aria-busy=${this.loading ? 'true' : 'false'}
```
KonnorRogers, line 331: one thing to call out with `aria-busy` is it makes the contents not update until busy is gone, so if the content inside the button changes, it won't be reported until loading is done.

claviska (dismissed):

PR #2695, opened 2026-07-31
Docs: Fix the CSS Part Anatomy Diagram for Property-Driven components

talbs:
The anatomy diagram on the Data Grid page was rendering an empty grid.

### The Diagram's Clone had no Data
The diagram clones a live example onto its stage. `cloneNode` copies attributes but not properties. So the clone came up empty and stayed empty.

`#copyLiveProperties` copies those across once both sides have settled.

### Width
The Anatomy subject was capped at 24rem, which is what clipped the Data Grid example. The tab group was clipping before this branch, too. This updated CSS should resolve those.


talbs:
@claude, review this PR.

claviska (approved):

PR #2702, opened 2026-08-04
Fix VoiceOver reports the wrong dropdown item count

### LINKED ISSUE #2697: <wa-divider> causes screen readers to miscount the number of <wa-dropdown> items
If a divider is used in `<wa-dropdown>`, subsequent items are removed from the total count. (e.g: 5 of 5 despite there being 7 items). This misrepresents the total items in the list.

Adding aria-hidden restores the correct and intended behavior.

Shivajith754:
fix voiceover announcing the wrong dropdown item count when a dropdown, includes non-item elements like a divider. Root cause: mixed children caused assistive technologies to count the wrong items. 
 Fix: sets aria-posinset and aria-setsize on top-level wa-dropdown-item elements using only filtered dropdown
 Testing: added regression test, browser verification, manual voiceover check confirmed 3 of 4.
 Fixes #2697

### DIFF
--- a/packages/webawesome/src/components/dropdown/dropdown.test.ts
+++ b/packages/webawesome/src/components/dropdown/dropdown.test.ts
@@ -59,6 +59,26 @@ describe('<wa-dropdown>', () => {
           const menu = el.shadowRoot!.querySelector('#menu')!;
           expect(menu.getAttribute('role')).to.equal('menu');
         });
+
+        it('should exclude labels and dividers from the reported menu item count', async () => {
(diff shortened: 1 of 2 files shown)


claviska (approved):
This change makes sense per spec, but it doesn't actually change the behavior _today_ because of this bug.

https://bugs.webkit.org/show_bug.cgi?id=297598

We'll merge and add a changelog entry, but it's important to note that a proper solution to #2697 is blocked by an upstream browser bug.

PR #2704, opened 2026-08-04
Fix Centered toast placements are offset on narrow mobile viewports

### LINKED ISSUE #2701: Centered Toast placements get offset on narrow (mobile) screen sizes
I'm seeing that both the `top-center` and `bottom-center` placements for Toast look wonky on a narrow viewport…it's like the left edge is itself a bit past 50% and then it paints way past the right edge. 

### To Reproduce

Use the Toast example here: https://webawesome.com/docs/components/toast/#placement
when your browser window width is pretty narrow and choose `top-center` or `bottom-center`

I was getting this both in Safari iOS and in Vivaldi on Linux, so doesn't seem platform-specific.

### Screenshots

Image

Shivajith754:
Fixes centered toast placement on narrow viewports and adds regression coverage for top-center and bottom-center.

Fixes #2701

### DIFF
--- a/packages/webawesome/src/components/toast/toast.styles.ts
+++ b/packages/webawesome/src/components/toast/toast.styles.ts
@@ -97,10 +97,5 @@ export default css`
       width: 100%;
       padding: var(--wa-space-s);
     }
-
-    :host([placement='top-center']),
-    :host([placement='bottom-center']) {
-      translate: 0;
-    }
   }
 `;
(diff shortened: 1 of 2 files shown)


claviska (approved):
Thanks for the PR! I was also able to replicate this and verify the fix work. I'll add a changelog entry on my end.

claviska:
Changed merged as part of #2706

PR #2706, opened 2026-08-04
Fix centered toast items on narrow screens

### LINKED ISSUE #2701: Centered Toast placements get offset on narrow (mobile) screen sizes

claviska:
Branches off #2704 to add a changelog entry and fix #2701.


talbs (approved):
[Your fix works great and as intended](https://www.youtube.com/watch?v=pJHbJ5LCl5A) - nice work!

Issue #2707, opened 2026-08-04
.wa-invert breaks button variants when applied directly


stevematney:
### Describe the bug
When applying the `wa-invert` class to the `wa-button` Web Component or to a native `<button>` directly, it breaks the variant coloration

### To Reproduce

1. Go to https://codepen.io/editor/stevematney/pen/019fce56-2d7e-7109-92fe-ea1aba1abd6c
2. See coloration differences around the application of the `wa-invert` class.

### Browser / OS
 - OS: Windows
 - Browser: Chrome
 - Browser version: 150.0.7871.188

claviska:
Thanks for the detailed repro!

This is expected behavior. `.wa-invert` (like `.wa-light` and `.wa-dark`) is a scoping class meant for containers, not for controls themselves. Under the hood, it resets the semantic color tokens on the element it's applied to so they can re-resolve against the flipped palette, and that reset collides with the element's own variant styles.

For custom elements this is a hard CSS limitation, since page styles targeting a host always beat the component's internal `:host` styles. Your third example is the intended usage: apply `.wa-invert` to a wrapper and everything inside will invert correctly.

I'll update the docs to make this more explicit!

claviska:
https://github.com/shoelace-style/webawesome/pull/2709

PR #2709, opened 2026-08-04
Add callout about `wa-invert`

claviska:
[As promised here](https://github.com/shoelace-style/webawesome/issues/2707#issuecomment-5184381638)

### DIFF
--- a/packages/webawesome/docs/docs/customizing.md
+++ b/packages/webawesome/docs/docs/customizing.md
@@ -146,7 +146,7 @@ You can force a section to behave like `.wa-dark` in light mode and like `.wa-li
 ```
 
 :::warning
-Apply `wa-invert` (and other color scheme classes like `wa-light` and `wa-dark`) to a container, not directly to a component. These classes reset the variant color tokens on the element they're applied to, which overrides the element's own variant styles.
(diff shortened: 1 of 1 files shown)


packages/webawesome/docs/docs/customizing.md
```diff
@@ -145,6 +145,10 @@ You can force a section to behave like `.wa-dark` in light mode and like `.wa-li
 </wa-card>
 ```
 
+:::warning
+Apply `wa-invert` (and other color scheme classes like `wa-light` and `wa-dark`) to a container, not directly to a component. These classes reset the variant color tokens on the element they're applied to, which overrides the element's own variant styles.
```
talbs: <strong>Apply Color Scheme Classes to a Container</strong><br />
Avoid adding `wa-invert` (and other color scheme classes like `wa-light` and `wa-dark`) directly to a component. These classes reset the variant color tokens on the element they're applied to, which overrides the element's own variant styles.

talbs (dismissed):
Looks good! Left one formatting suggestion - take it or leave it.

packages/webawesome/docs/docs/customizing.md
claviska:
```suggestion
Avoid adding `wa-invert` (and other color scheme classes like `wa-light` and `wa-dark`) directly to a component. These classes reset the variant color tokens on the element they're applied to, which overrides the element's own variant styles.
```

PR #2710, opened 2026-08-04
Docs: Keep Sidebar's Pro Badge Scoped to Pro Components

talbs:
Every page under `/docs/patterns/*/` badges *every* component in the sidebar as Pro, free ones included. It's baked in at build time, so signed-out visitors see it too.

### Cause
The `sidebarComponent` macro sets a local named `isPro`. For free components that expression evaluates to `undefined` rather than `false`, and Nunjucks treats an undefined local as unset and falls through to the page context — where pattern pages carry their own `isPro: true` from the data cascade.

### Fix
Renamed the local to `hasProBadge` and coerced it to a real boolean. The rename fixes today's collision; the coercion looks redundant but isn't — it's what stops an undefined local from reaching the context again under some future variable name.


talbs:
@claude, review this PR.

KonnorRogers (approved):

PR #2712, opened 2026-08-05
fix super calls in `firstUpdated`

KonnorRogers:
I noticed `slotchange` wasn't firing for `dropdown` and tests were failing...so naturally I got carried away...


claviska (commented):
From the [Lit docs](https://lit.dev/docs/components/lifecycle/#firstupdated):

Image

Are you sure this is necessary?

KonnorRogers:
@claviska any components which rely on `slotchange` events will be broken without it when they SSR.

https://github.com/shoelace-style/webawesome/blob/eed8125a0e9b13e4200f013f887ce2392a601bbf/packages/webawesome/src/internal/webawesome-element.ts#L147-L157

claviska:
@KonnorRogers ahhh, OK. Didn't realize we're overloading `firstUpdated()` now. Can we lint for this so it doesn't become a footgun?

claviska (approved):

PR #2717, opened 2026-08-06
Tag: Make Remove Button Keyboard Operable

talbs:
You could remove a `<wa-tag>` with the mouse, but not the keyboard - the remove button was skipped when you tabbed through.

Now you can use Tab to select it, then press Enter or Space to remove the `<wa-tag>`.

### Why it wasn't just a one-line fix

The remove button was intentionally skipped. `<wa-select multiple>` puts its tags inside the select itself, so making the remove button focusable everywhere would mean a `<wa-select>` with eight tags becomes nine Tab stops instead of one. 

That breaks how form controls are supposed to work, and it's the same reason the `<wa-select>`'s clear button and its dropdown list are skipped as well.

There's a second problem in `<wa-select>`. When it has tags, the select shows focus as an outline around the whole control. Tab to a `<wa-tag>'s remove button, and that outline stays put while a tiny second outline appears — so the focus ring looks like it didn't move.

### How it works

`<wa-tag>` gets an internal setting that controls whether its remove button is a Tab stop. It's on by default, so a `<wa-tag>` on its own behaves correctly. `<wa-select>` turns it off for the tags it renders, so the select stays a single Tab stop.

`<wa-select>` turns it off in two places. 
1. Once in the template, which covers the normal case. 
1. Once after every render, which covers people who've customized their tags with `getTag` — that option can return raw HTML, and a template setting can't reach into raw HTML.

The setting is a JavaScript property with no matching HTML attribute. That's deliberate: an attribute would show up in the public API docs and people would start using it, and then we couldn't change it.

### Revised tests

For `<wa-tag>`: Tab reaches the ×, Enter removes the tag, Space removes the tag, and turning the setting off skips the × again. Each test presses a real Tab key and checks where focus actually went, rather than reading an attribute and assuming.

For `<wa-select>`: tags don't become Tab stops, including when someone customizes `getTag`. Both tests also check that the `<wa-tag>`s actually rendered and that focus actually reached the select first, so they can't pass by accident.

### Docs
The Removable example now actually removes the `<wa-tag>` and moves focus to the container, instead of just fading the tag out and leaving focus on an invisible element. There's a Reset button so you can try it more than once.

## Companion PRs
- https://github.com/shoelace-style/webawesome-pro/pull/287

### DIFF
--- a/packages/webawesome/docs/docs/components/tag.md
+++ b/packages/webawesome/docs/docs/components/tag.md
@@ -102,10 +102,15 @@ Use the `pill` attribute to give tags rounded edges.
 
 ### Removable
 
-Use the `with-remove` attribute to add a remove button to the tag. The button carries a built-in `Remove` label for assistive technology, and activating it emits the `wa-remove` event so you can handle the removal.
+Use the `with-remove` attribute to add a remove button to the tag. The button carries a built-in `Remove` label for assistive technology, and activating it emits the `wa-remove` event so you can handle the removal. The button is reachable with <kbd>Tab</kbd> and activates with <kbd>Enter</kbd> or <kbd>Space</kbd>.
+
+:::info
+<strong>`wa-remove` only announces the intent</strong><br />
+Removing the tag is up to you. Move focus to the tag's container afterwards (give it `tabindex="-1"`) so keyboard users aren't left on a removed element.
+:::
 
 ```html {.example}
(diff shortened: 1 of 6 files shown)


talbs:
@claude, review this PR.

talbs:
I think this use case needs more discussion planning, so closing this approach for now.

PR #2718, opened 2026-08-06
Fix system icon fill color

### LINKED ISSUE #2688: Pagination dark mode ... is the wrong color

claviska:
This issue was identified in #2688. PR #2689 attempted to solve it by adding the fill color to the SVG, but the correct place to do this is in the library's mutator.

The `fill` isn't set in Font Awesome SVGs by default, so I removed it and added a mutator test. To review this, we need to look at various components that use the system library to ensure fills are appropriate applied (i.e. not just the ellipsis in pagination, as that was just one example of the bug).

Closes #2688


kelseythejackson (dismissed):
Looks good

PR #2720, opened 2026-08-06
Fix textarea structure to support custom styles like other form controls

### LINKED ISSUE #2614: <wa-input> Customizing label position does not work
### Describe the bug
An example to show how to customizing label position simply does not work.

### To Reproduce

1. Go to https://webawesome.com/docs/components/input/
2. Scroll down to Customizing Label Position

### Demo
https://webawesome.com/docs/components/input/

### Browser / OS
 - OS: Windows
 - Browser: Firefox
 - Browser version: latest

claviska:
The textarea has a `<div class="footer">` that surrounds the hint. The `hint` part itself was on the slot, which is an antipattern. This PR fixes that, allowing for consistent label styling, e.g.:

Before
before

After
after

I noticed this when reviewing https://github.com/shoelace-style/webawesome/issues/2614, which I _think_ is what the OP is referring to.

Closes #2614

### DIFF
--- a/packages/webawesome/docs/docs/components/input.md
+++ b/packages/webawesome/docs/docs/components/input.md
@@ -177,7 +177,7 @@ Use [CSS parts](#css-parts) to customize the way form controls are drawn. This e
       align-items: center;
     }
 
-    ::part(label) {
+    ::part(form-control-label) {
       text-align: right;
     }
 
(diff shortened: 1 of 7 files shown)


talbs (dismissed):
Image

talbs (approved):

PR #2726, opened 2026-08-07
Textarea: Follow-ups to Hint Part Fix

talbs:
A few follow-ups to https://github.com/shoelace-style/webawesome/pull/2720. 

The first one is an actual regression. The other three are small.

### The Regression
Once `part="hint"` came off the slot, the slot fell back to its UA default of `display: contents`. This means no box. So `.footer.has-count .hint { flex: 1 1 auto; min-width: 0 }` was styling nothing.

You see it with a hint that can't wrap. In a 300px textarea, the character count lands about 162px past the right edge of the control. Prior to this, it stayed put.

I added `display: block,` to the slot, which fixes it. And I added a test that fails without that line.

### Smaller Cleanups
- In a test, `should lay the hint and character count out on one line` was checking the count against its own parent, which is always true. It passed with the two stacked on top of each other. Now it checks the `.hint` slot.
- `'has-hint': this.withCount` says there's a hint when there's only a count. It's load-bearing, but it lies. `form-control.styles.ts` takes `.has-count` in the same `:not()` now, so the flag can go.
- Changelog entry gets `[issue:2614]`, and says out loud that `::part(hint)` covers the hint and the character count both.


claviska (approved):
good catch, thanks!

PR #2729, opened 2026-08-08
Native Styles: Align Table Row Headers with Their Cells

talbs:
`<th scope="row">` was picking up `--wa-font-size-smaller`, the type treatment meant for column headers. That made the row-header text inconsistent with the  `<td>`s sitting next to it.

### What Went Wrong
- `native.css` already tells column headers apart from row headers. The stronger bottom border is scoped with `:not([scope='row']):not(:has(+ td))`. `font-size` just sat outside that block and hit every `th`.
- It misaligns twice over, which is why you can see it at all. `td, th` share `padding: 0.75em` and `vertical-align: top`, so shrinking the font shrinks the padding *and* the line box. 

### The Fix
- Move `font-size` into the column-header block, next to the border it should have been sitting with. 
- Row headers stay bold and take the body size.

### Docs
The native tables page had no `scope="row"` example, so nothing pointed anyone at the attribute that settles this. So I added one.

### DIFF
--- a/packages/webawesome/docs/docs/resources/changelog.md
+++ b/packages/webawesome/docs/docs/resources/changelog.md
@@ -33,6 +33,7 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 
 :::fixed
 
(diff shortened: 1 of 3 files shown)


omnifides (approved):
Good call!

Issue #2727, opened 2026-08-07
Password toggle on <wa-input> not keyboard accessible


adamchaboryk:
### Describe the bug
The toggle password button within the `<wa-input type="password">` is not keyboard accessible (assumingely due to the `tabindex="-1"`). 

### To Reproduce

1. Go to [Toggle Password](https://webawesome.com/docs/components/input/#toggle-password)
2. Use the tab key to access Toggle Password button (eye icon).
3. Confirm toggle password is not keyboard focusable. 

### Screenshots

<img width="853" height="430" alt="Screenshot of WA Input password toggle component and corresponding HTML code." src="https://github.com/user-attachments/assets/36ce0ae2-5989-4016-9f0c-9eca519238f3" />

### Browser / OS
 - OS: MacOS
 - Browser: Chrome
 - Browser version: 151

claviska:
Hey, Adam. A very similar issue has been brought up a number of times in the Shoelace repo about the input's clear button. I want to share a prior response because it's the same principal:

> This has been discussed before in https://github.com/shoelace-style/shoelace/issues/727#issuecomment-1098072111. The decision to not focus the clear button was based on how native browser inputs work and on potentially destructive operations resulting from adding them to the tab order.
> 
> > I agree with this statement. It feels dangerous, especially since Enter can trigger a form submit as well as activate the clear button. User might expect that tabbing will reach the next field and press it hoping for a submit, but instead their data will be erased. Perhaps this is why browsers omit it from the tab order.
> 
> The clear button remains discoverable to screen readers and keyboard users can clear it quickly without the need to focus. Additionally, the extra focus is often unexpected and confusing to users. As such, I consider this a non-issue and working as intended. You can read more about this pattern here.
> 
> https://scottaohara.github.io/clear-text-field-button/

[Source](https://github.com/shoelace-style/shoelace/discussions/1905#discussioncomment-8671591)

I'm closing this as a wontfix, but happy to discuss things further and always open to suggestions!

adamchaboryk:
Hey @claviska, thanks for the quick response!! Sorry, I did a quick search in this repo first to see if this issue was reported, but didn't think to check the old Shoelace repo.

Happy to discuss this one further!

I can understand the rationale for not making the clear buttons keyboard focusable based on your and Scott's findings, especially since a keyboard user can still use backspace to clear a field.

However, I think the show password button is intrinsically different from a clear button for a few reasons:

[...]

PR #2736, opened 2026-08-14
Make password toggle focusable

claviska:
Fixes the accessibility bug described in https://github.com/shoelace-style/webawesome/issues/2727


claviska:
Pinging @talbs for a review to make sure the focus styles are 👌 

CleanShot 2026-08-14 at 12 37 46@2x

packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -56,7 +56,8 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
 
 :::changed
 
-- Improved accessibility of `<wa-dropdown>` by adding `aria-posinset` and `aria-setsize` so supportive screen readers can correctly announce the number of dropdown items []
+- Improved accessibility of `<wa-dropdown>` by adding `aria-posinset` and `aria-setsize` so supportive screen readers can correctly announce the number of dropdown items [#2697]
```
talbs: 👀

talbs (dismissed):

talbs:
@claviska, I added a commit tightening the focus ring. 

The toggle was inheriting the field's full height (20×41px, around a 16px icon), so the outline was drawn entirely outside the field's top and bottom borders. It's now a 24×24 square that sits inside the field.

Two calls for you to check out: 
- the box applies to `.clear` as well so both icons match
- an `::after` keeps the click target at full field height so pointer behavior doesn't change

---

Also, I rebased this work against `next`. Your `[#2697]` was updated to `[issue:2697]` to match the rest of the Changelog's entries.

claviska:
> an ::after keeps the click target at full field height so pointer behavior doesn't change

This is fantastic! Thanks :D

talbs (approved):

Issue #2723, opened 2026-08-07
Axe Audit Found an Issue with Wa-Card.


kelseythejackson:
A user raised an issue in Discord about `wa-card` failing an audit. I was able to verify that it does fail when this test is added to the card tests:

```  it('repro: card header outside sectioning content', async () => {
          const el = await fixture(html`
            <div>
              <header>Site header</header>
              <wa-card>
                <div slot="header">Card header</div>
                Body content
              </wa-card>
            </div>
          `);
          await expect(el).to.be.accessible();
        });```

Failed messages:
``` ❌ <wa-card> > with "client-only" rendering > accessibility > repro: card header outside sectioning content
      Error: Accessibility Violations
      ---
      Rule: landmark-no-duplicate-banner
      Impact: moderate
      Document should not have more than one banner landmark (https://dequeuniversity.com/rules/axe/4.11/landmark-no-duplicate-banner?application=axeAPI)
      
      Issue target: header
      Context: <header>Site header</header>
      Fix any of the following:
        Document has more than one banner landmark
      ---
      Rule: landmark-unique
      Impact: moderate
      Landmarks should have a unique role or role/label/title (i.e. accessible name) combination (https://dequeuniversity.com/rules/axe/4.11/landmark-unique?application=axeAPI)
      
      Issue target: header
      Context: <header>Site header</header>
      Fix any of the following:
        The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable
      ---

 ❌ <wa-card> > with "ssr-client-hydrated" rendering > accessibility > repro: card header outside sectioning content
      Error: Accessibility Violations
      ---
      Rule: landmark-no-duplicate-banner
      Impact: moderate
      Document should not have more than one banner landmark (https://dequeuniversity.com/rules/axe/4.11/landmark-no-duplicate-banner?application=axeAPI)
      
      Issue target: header
      Context: <header>Site header</header>
      Fix any of the following:
        Document has more than one banner landmark
      ---
      Rule: landmark-unique
      Impact: moderate
      Landmarks should have a unique role or role/label/title (i.e. accessible name) combination (https://dequeuniversity.com/rules/axe/4.11/landmark-unique?application=axeAPI)
      
      Issue target: header
      Context: <header>Site header</header>
      Fix any of the following:
        The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable```

PR #2737, opened 2026-08-14
Off with their head[ers and footers]!

claviska:
Using header and footer inside of these components was causing an accessibility violation as described in #2723. This reverts the affected components to use `<div>` elements internally, allowing consumers to slot their own structured content in and preventing the duplicate landmarks bug reported by Axe.


talbs (dismissed):

jaredcwhite:
@claviska Epic PR title! 👏🏻

PR #2746, opened 2026-08-17
Fix tooltip closing on slotted content

### LINKED ISSUE #2745: Wa-Tooltip closes when hovering slotted content
### Describe the bug
A Wa-Tooltip can contain any kind of slotted html content.
That also means it can itself contain another slot from a parent component.

`<wa-tooltip>
    <slot name="content"></slot>
 </wa-tooltip>`

When something is then slotted into "content" and the user hovers over the slotted node, the tooltip closes,
since the "contains()" check cannot resolve that it is within the tooltip.

### Demo
A demo of this behaviour can be seen in this [Codepen example](https://codepen.io/GuymanDudeson/pen/BypgQeP)

### Browser / OS
 - OS: Windows 11
 - Browser: Chrome
 - Browser version: 151.0.7922.138 (Official Build) (64-bit)

Shivajith754:
Fixes #2745
Added a composed-tree containment helper that traverses slot assignments and shadow-root hosts, and replaced the tooltip hover mouseout containment checks with it. Added a focused regression test for content forwarded through a parent slot.

### DIFF
--- a/packages/webawesome/src/components/tooltip/tooltip.test.ts
+++ b/packages/webawesome/src/components/tooltip/tooltip.test.ts
@@ -474,6 +474,43 @@ describe('<wa-tooltip>', () => {
       await waitUntil(() => !tooltip.open);
       expect(tooltip.open).to.be.false;
     });
+
+    it('should remain open when the pointer moves onto content forwarded through a parent slot', async () => {
+      const el = await fixtures[0]<HTMLDivElement>(html`<div></div>`);
+      const shadowRoot = el.attachShadow({ mode: 'open' });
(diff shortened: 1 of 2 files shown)


GuymanDudeson:
Thanks for implementing this. Any update on this? I need this fix in my application.

claviska:
@GuymanDudeson I'll make sure this gets reviewed this week and lands in our next release if the fix is good.

GuymanDudeson:
Bumping this again since it is still a blocker for me. Thanks.

claviska (approved):
Fix looks good. I'll get a changelog entry added. Thanks for the PR!

PR #2750, opened 2026-08-19
Tidy Up Loose Ends Post-Focus Ring Work

talbs:
Four small follow-ups to https://github.com/shoelace-style/webawesome/pull/2736 and https://github.com/shoelace-style/webawesome/pull/2749.

- The changelog entry for the switch focus ring landed under `3.0.0-alpha.5` instead of Unreleased. Moved to where it belongs.
-`<wa-switch>`'s focus rule selected `~ .switch`, which reads like the `switch` part but actually matches `<span part="control">`. The `switch` part is the outer `<label>`. It's the same element either way, now named honestly, and it matches how `<wa-checkbox>` selects `[part~='control']`.
- Dropped a redundant comment above the `::after` in `<wa-input>`.
- `<wa-input>`'s icon buttons grew a box for their focus ring, which pushed their glyphs 2px inboard of the trailing padding edge every other form control sits on. Compensated with margins so the box grows around the glyph instead of shoving it.


claviska (approved):
:shipit:

PR #2759, opened 2026-08-21
Docs: Align Side Labels in the Label Position Example

### LINKED ISSUE #2614: <wa-input> Customizing label position does not work

talbs:
The "Customizing Label Position" example on the input and number input pages never aligned its labels the way its own CSS claimed. Fixed here. Closes #2614.

### What was broken

The example set `text-align: right` on the label part. The label part is a flex container, and `text-align` does not move a flex container's contents sideways. The rule did nothing, so the labels stayed left while the code said right. `justify-content: end` is the property that works here.

### Changes

| What | Before | After |
| --- | --- | --- |
| Label alignment | `text-align: right`, which did nothing | `justify-content: end` |
| Textarea label position | Centered against the full height of the box | Sits on the first line of text |
| Second input's label | `Email` | `Email address` |

### Why the label text changed

"Name", "Email", and "Bio" are nearly the same width. Right-aligning them shifts two labels by a few pixels. The fix would be invisible, and someone would file the same bug again. "Email address" is long enough to make the alignment obvious.

### The textarea label

It used to float halfway down a tall box, nowhere near the line you type on. Now it lines up with the first line of text. Single-line inputs keep their centered labels, since for them the middle of the box is the line of text.

`align-items: baseline` looks like the elegant answer but is not. The browser builds the field's baseline from the bottom edge of the box, so every label drops to the bottom.

### Docs text

Both pages now say the label part is a flex container and takes `justify-content`, not `text-align`. That is the part that tripped up the person who reported this. There is also a short note explaining why the labels align to the end.

No changelog entry, since nothing outside `docs/` changed.

### DIFF
--- a/packages/webawesome/docs/docs/components/number-input.md
+++ b/packages/webawesome/docs/docs/components/number-input.md
@@ -169,7 +169,12 @@ Use the `start` and `end` slots to add presentational elements like `<wa-icon>` 
 
 ### Customizing Label Position
 
-Use [CSS parts](#css-parts) to customize the way form controls are drawn. This example uses CSS grid to position the label to the left of the control, but the possible orientations are nearly endless. The same technique works for inputs, textareas, radio groups, and similar form controls. The label part is a flex container, so use `justify-content` to align its text. `text-align` has no effect.
+Use [CSS parts](#css-parts) to customize the way form controls are drawn. This example uses CSS grid to position the label to the left of the control, but the possible orientations are nearly endless. The same technique works for inputs, textareas, radio groups, and similar form controls. The label part is a flex container, so align it with `justify-content` rather than `text-align`.
+
+:::info
+<strong>Labels sit at the end of their column so every one is the same distance from its field.</strong><br />
+Start-aligned labels leave a different-sized gap on each row, which makes it harder to see which label belongs to which field.
+:::
 
 ```html {.example}
 <div class="label-on-left">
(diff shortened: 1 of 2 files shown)


claviska (dismissed):

packages/webawesome/docs/docs/components/number-input.md
```diff
@@ -169,7 +169,12 @@ Use the `start` and `end` slots to add presentational elements like `<wa-icon>`
 
 ### Customizing Label Position
 
-Use [CSS parts](#css-parts) to customize the way form controls are drawn. This example uses CSS grid to position the label to the left of the control, but the possible orientations are nearly endless. The same technique works for inputs, textareas, radio groups, and similar form controls.
+Use [CSS parts](#css-parts) to customize the way form controls are drawn. This example uses CSS grid to position the label to the left of the control, but the possible orientations are nearly endless. The same technique works for inputs, textareas, radio groups, and similar form controls. The label part is a flex container, so align it with `justify-content` rather than `text-align`.
+
+:::info
+<strong>Labels sit at the end of their column so every one is the same distance from its field.</strong><br />
+Start-aligned labels leave a different-sized gap on each row, which makes it harder to see which label belongs to which field.
+:::
```
claviska, line 177: Markdown should be fine here:

```suggestion
:::info
**Labels sit at the end of their column so every one is the same distance from its field.**

Start-aligned labels leave a different-sized gap on each row, which makes it harder to see which label belongs to which field.
:::
```

PR #2760, opened 2026-08-21
Site Search: Remove Individual Recent Searches

talbs:
Recent searches now have an X that removes them one at a time. Before this, a typo like `fart` stayed in the list until five more searches pushed it out.

### Removing a search

- Hover a recent search, or arrow onto it, and an X appears at the end of the row.
- Click the X to remove that search.
- From the keyboard, press Shift+Delete on the selected row.
- Shift+Backspace does the same, because MacBooks have no forward-delete key.

Recent rows also lost their chevron. A chevron says "this opens a page", and these rows do not open anything. They put the text back in the search box and run the search again.

### Two bugs this uncovered

Both existed before this change. Both are fixed here.

| What you did | Before | After |
| --- | --- | --- |
| Tab to a row, press Enter | Opened whichever row the arrow keys were on | Opens the row you are actually on |
| Tab with recent searches showing | Stopped on the scrolling area first, drawing a ring across the whole dialog | Goes straight to the first Quick Link |

### How the rows are built now

- The row highlight moved off the link and onto the row itself.
- That is what lets it run underneath the X.
- Quick Links and search results share the same row class.
- Their appearance is unchanged. I compared screenshots before and after and they matched exactly.

### Screen readers and keyboard

- Each row is announced with its query and nothing else.
- The X carries its own name, like `Remove dialog from recent searches`.
- Removing a search is spoken out loud.
- Focus lands on the next row's X, or the search box if the list is now empty.

### DIFF
--- a/packages/webawesome/docs/_includes/search.njk
+++ b/packages/webawesome/docs/_includes/search.njk
@@ -14,7 +14,9 @@
     </header>
 
     {# Body #}
-    <div id="site-search-body">
+    {# Opts out of Chrome's keyboard-focusable scrollers — that's for regions with
+       no focusable content, and the extra tab stop rings the whole dialog. #}
+    <div id="site-search-body" tabindex="-1">
       {# Default state — Suggested starts + Recent searches #}
       <div id="site-search-default" class="wa-stack wa-gap-0">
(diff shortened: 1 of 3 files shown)


talbs:
@claude, review this PR.

PR #2761, opened 2026-08-21
Site Search: Drop the Chevrons

talbs:
Removes the chevron from every row in the site search dialog. Follow-up to #2760, which removed it from recent searches only. Rebased onto `next` now that #2760 has merged.

A chevron means "this opens something in place", like a section expanding or a panel sliding in. Every row here opens a different page and closes the dialog. The arrow was pointing at the wrong idea.

### What changes

- Quick Links, search results, and recent searches no longer show a chevron.
- Nothing else about the rows changes.
- The `.site-search-result-caret` class is now unused and deleted.

### A small bonus

Search result descriptions get about 29 pixels back per row. That is the chevron plus the gap in front of it. One example, same search:

| | Description shown |
| --- | --- |
| Before | Buttons represent actions the user can take, such as sub... |
| After | Buttons represent actions the user can take, such as submitti... |

### Worth knowing

On phones and tablets the chevron was already invisible. It only appeared when you hovered a row with a mouse, or moved to it with arrow keys. Touch users have never seen it, so nothing changes for them.

### DIFF
--- a/packages/webawesome/docs/_includes/search.njk
+++ b/packages/webawesome/docs/_includes/search.njk
@@ -29,8 +29,6 @@
                 <div class="site-search-result-details">
                   <div class="site-search-result-title wa-font-size-s">Components</div>
                 </div>
-                <wa-icon class="site-search-result-caret" name="chevron-right" variant="regular"
-                  aria-hidden="true"></wa-icon>
               </a>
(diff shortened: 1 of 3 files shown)


talbs:
@claude, review this PR.

PR #2783, opened 2026-09-03
Docs: Re-Aim Deep Links After Components Upgrade

talbs:
Links with a `#` in them, like `/support#client-projects`, can drop you in the wrong place on the page. This fixes that.

The page is about 3,200px taller before the components load. The browser jumps to the `#` using that taller version. Then the page shrinks, and your target ends up roughly 3,700px above the top of the screen, with nothing to scroll you back.

### The Fix
`scroll.js` already waited for components before restoring your scroll position on a reload. It never did the same for `#` links. Now it does, offset by the sticky header so the heading isn't tucked underneath it.

Two guards worth knowing about:
- If a scroll position was already restored, that wins. Covers reloads, and back or forward, where the browser puts you back where you were.
- `allDefined()` only waits for components to be defined, plus one frame. It doesn't wait for them to finish rendering. So this waits for every component's render as well, then takes a frame, before measuring anything.

### What This Doesn't Change
Both unchanged from before this PR: 
- clicking an anchor while the page is still loading
- clicking back or forward with no `#`.

### DIFF
--- a/packages/webawesome/docs/assets/scripts/scroll.js
+++ b/packages/webawesome/docs/assets/scripts/scroll.js
@@ -56,10 +56,27 @@ allDefined().then(() => {
   const scrollY = sessionStorage.getItem(key);
 
   // Only restore when reloading, otherwise clear it
-  if (navigationType === 'reload' && scrollY) {
+  const didRestoreScroll = navigationType === 'reload' && scrollY;
+
+  if (didRestoreScroll) {
     window.scrollTo(0, scrollY);
   } else {
     sessionStorage.removeItem(key);
+  }
+
+  // The browser aims fragment navigation at the pre-upgrade layout, which is taller than the
+  // settled one, so a deep link lands well past its target. A restored position outranks it:
+  // ours on reload, or the browser's own on back/forward.
+  const id = location.hash.slice(1);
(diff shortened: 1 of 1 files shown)


talbs:
@claude, review this PR.

packages/webawesome/docs/assets/scripts/scroll.js
```diff
@@ -56,12 +56,29 @@ allDefined().then(() => {
...
+
+  if (didRestoreScroll) {
     window.scrollTo(0, scrollY);
   } else {
     sessionStorage.removeItem(key);
   }
 
+  // The browser aims fragment navigation at the pre-upgrade layout, which is taller than the
+  // settled one, so a deep link lands well past its target. A restored position outranks it:
+  // ours on reload, or the browser's own on back/forward.
```
KonnorRogers: I think what we should do is:


```js
const page = await document.querySelector("wa-page")
await page.updateComplete // Rendering complete
requestAnimationFrame(() => {
  // scroll stuff here
})
```

Possibly even wrap the `requestAnimationFrame` in a `setTimeout` to give layout a little bit of time to settle.

talbs:
@claude, re-review this PR.

PR #2788, opened 2026-09-06
Improve named-slot detection for components with large contents

### LINKED ISSUE #2787: Named-slot detection scans entire descendant subtrees on component updates
### Describe the bug

`HasSlotController.hasNamedSlot()` uses `querySelector(':scope > [slot="…"]')` to determine whether a direct child fills a named slot. Although the selector only *matches* direct children, the query searches through descendant subtrees. Consequently, updating a card containing a large table repeatedly traverses the table to check its five named slots.

The cost grows with body contents that cannot fill those slots. This also affects other users of `HasSlotController`, including dialog, input, and select.

### Reproduction

Save the following as an HTML file, serve it locally, and open the browser console. It uses the published 3.12.0 card. Each result is the median duration of 100 awaited component updates, after three warmup samples. Table construction and layout are outside the timed region.

```html
<!doctype html>
<html lang="en">
<script type="module">
  import 'https://cdn.jsdelivr.net/npm/@awesome.me/webawesome@3.12.0/dist/components/card/card.js';

  const results = [];
  for (const rows of [10, 100, 1000]) {
    const card = document.createElement('wa-card');
    card.innerHTML = '<table><tbody>' +
      ('<tr>' + '<td><span>Cell</span></td>'.repeat(8) + '</tr>').repeat(rows) +
      '</tbody></table>';
    document.body.append(card);
    await card.updateComplete;
    await new Promise(requestAnimationFrame);
    await new Promise(requestAnimationFrame);
    const samples = [];
    for (let run = 0; run < 10; run++) {
      const start = performance.now();
      for (let i = 0; i < 100; i++) {
        card.requestUpdate();
        await card.updateComplete;
      }
      if (run >= 3) samples.push(performance.now() - start);
    }
    samples.sort((a, b) => a - b);
    results.push({ rows, descendants: card.querySelectorAll('*').length, ms: samples[3] });
    card.remove();
  }
  console.table(results);
</script>
</html>
```

### Investigation

I compared production-minified bundles at `f1981f73d9c53977108da7d30ed90c6a3eda8145`, changing only the slot helper to use `host.matches(':has(> [slot="…"])')`. That expresses the existence test on the host and lets the browser restrict traversal to direct children. Web Awesome already uses `:has()` in its styles.

Results below are milliseconds per **100 updates**, using the repository's pinned Playwright browsers on macOS arm64. Each cell is baseline → proposed change. The benchmark uses separate pages and bundles in A/B/B/A order, with 14 retained measurements per variant. These are scripting measurements for this workload, not page-load or interaction-latency claims.

| Contents | Chromium 136 | Firefox 137 | WebKit 18.4 |
| --- | ---: | ---: | ---: |
| Card, 10 table rows | 0.585 → 0.118 | 0.610 → 0.480 | 0.460 → 0.180 |
| Card, 100 table rows | 4.910 → 0.108 | 2.380 → 0.420 | 3.030 → 0.140 |
| Card, 1,000 table rows | 48.605 → 0.112 | 22.990 → 0.410 | 33.110 → 0.120 |
| Select, 10 direct options | 0.608 → 0.620 | 1.150 → 1.080 | 0.710 → 0.690 |
| Select, 100 direct options | 1.163 → 1.170 | 1.370 → 1.380 | 1.320 → 1.460 |
| Select, 1,000 direct options | 5.605 → 5.570 | 8.150 → 8.630 | 6.070 → 6.930 |

There is a tradeoff for very wide, shallow content: the 1,000-option select is about 6% slower in Firefox and 14% slower in WebKit in this benchmark, or approximately 0.005–0.009 ms per update. Chromium is effectively unchanged. I also tried a JavaScript child loop, but discarded it because its select overhead was substantially larger.

I have a small patch using the native selector, with coverage for nested slots, direct slots after unslotted content, and dynamic addition/renaming/removal/reparenting. It preserves the existing `with-*` hydration fallback and does not introduce caching or a public API change.

gregjotau:
Updating a card containing a large table currently searches through the table five times to detect named slots, even though only direct children can fill those slots. Match `:has(> [slot="…"])` on the host instead of querying descendants with `:scope > [slot="…"]` so the check no longer grows with nested body contents.

Fixes #2787.

The change is limited to `HasSlotController.hasNamedSlot()`. It keeps both the existing `with-*` SSR/hydration fallback and the optimistic named-slot assumption when server DOM APIs are unavailable. It introduces no cache, dependency or public API. Web Awesome already uses `:has()` in component styles and it fits the browser support policy.

Added card coverage runs in both client-only and SSR-client-hydrated fixtures: nested named slots, direct slots following unslotted content, addition/rename/removal, and reparenting between direct and nested positions. A controller regression test covers the existing server fallback when DOM APIs are unavailable. The changelog describes the performance improvement.

### Performance

Milliseconds per **100 awaited Lit updates**, baseline → patch:

| Contents | Chromium 136 | Firefox 137 | WebKit 18.4 |
| --- | ---: | ---: | ---: |
| Card, 100 table rows | 4.910 → 0.108 | 2.380 → 0.420 | 3.030 → 0.140 |
| Card, 1,000 table rows | 48.605 → 0.112 | 22.990 → 0.410 | 33.110 → 0.120 |
| Select, 100 direct options | 1.163 → 1.170 | 1.370 → 1.380 | 1.320 → 1.460 |
| Select, 1,000 direct options | 5.605 → 5.570 | 8.150 → 8.630 | 6.070 → 6.930 |

There is a tradeoff for wide, shallow content: the 1,000-option select adds approximately 0.005 ms per update in Firefox and 0.009 ms in WebKit (about 6% and 14% in this workload); Chromium is effectively unchanged. A JavaScript direct-child loop had a larger select regression, so this patch uses the native selector.

Measured on macOS arm64 / Node 24.19.0 with repository-pinned Playwright browsers. Production-minified bundles differ only in the slot helper, compared against `f1981f73d9c53977108da7d30ed90c6a3eda8145` in A/B/B/A order with fresh pages, three warmups per page and 14 retained samples per variant. Construction and layout are outside timing. These are component scripting measurements, not whole-page speedups. The issue has the reproduction and all six card/select sizes.

### Validation

- Formatting, TypeScript/package/docs build, `test:parts`, `test:anatomy`, and `git diff --check` pass.
- The final version passes `npm run test:hydration`: all 70 docs pages hydrate cleanly, including reloads. The upstream Hydration Tests workflow also passes.
- The new server-fallback controller test passes in all three browsers.
- `npx web-test-runner --group card --reporter dot`: 56 passed per browser in Chromium, Firefox and WebKit, covering both rendering modes.
- Final complete 77-file suite with `CI=true` (Chromium/Firefox, both rendering modes): Chromium 4,831 passed / 0 failed; Firefox 4,828 passed / 3 failed, with 38 skips in each. The Firefox failures are one copy-button custom-trigger hover timeout and two tooltip pointer/light-dismiss assertions.
- All of those Firefox failures pass on isolated rerun: 84 copy-button tests and 68 tooltip tests pass per browser in Chromium, Firefox and WebKit.
- An earlier full run also encountered a Firefox toast-item hover/timer failure; all 68 toast-item tests pass per browser in isolation.
- All five WebKit failures reproduce on an unchanged worktree at the base commit: two button click-coordinate errors, one carousel loop timeout, and two toast SSR tests (viewport placement and keeping the stack visible). The toast lifecycle failure is intermittent on the baseline as well.
- `npm run verify` reaches `verify:skills` and stops on the same 15 errors on both checkouts: public-clone references to missing Pro components and the existing `scripts/design-skill/SKILL.md` callout-reference link. The browser suite was therefore run separately; the full verify command is not green.


gregjotau:
@claviska  FYI. I just asked Astra GPT-6 to benchmark and find improvements on our accounting system app: reai.no :)

packages/webawesome/src/internal/slot.ts
```diff
@@ -41,7 +41,9 @@ export class HasSlotController implements ReactiveController {
   }
 
   private hasNamedSlot(name: string) {
-    return this.host.querySelector?.(`:scope > [slot="${name}"]`) !== null;
+    // Match on the host to avoid searching descendant subtrees for direct-child slots.
+    // Preserve the existing assumption that named slots are present when rendering without DOM APIs.
+    return this.host.matches?.(`:has(> [slot="${name}"])`) ?? true;
```
KonnorRogers, line 46: If you _really_ want the most performant option, walk the `children`, and then early return so use an iterator.

```js
private hasNamedSlot(name: string) {
  // early return if no children found.
  if (!this.host?.children?.length) { return false }

  for (const el of this.host.children) {
    if (el.slot === name) {
      return true
    }
  }

  return false
}
```


> // Preserve the existing assumption that named slots are present when rendering without DOM APIs.

Also, it should be `?? false`, as the original assumption without DOM APIs is actually false.

KonnorRogers (changes requested):

packages/webawesome/src/internal/slot.ts
gregjotau: Might be!

I just asked Astra to benchmark what changes would help our particular app. If you find a better method, that is great 🚀

gregjotau: Feel free to make new PR / change it.

PR #2794, opened 2026-09-08
Add <wa-stepper> and <wa-step> components

kelseythejackson:
## Summary

- Adds `<wa-stepper>` and `<wa-step>` for checkout flows, multi-step forms, onboarding, and setup wizards.
- Steps use a single roving tab stop: Tab moves focus into/out of the stepper as one stop, Arrow keys (or Home/End) move focus between steps, and Enter/Space (or a click) activates the focused step. The structure is `<nav><ol role="list">` with `aria-current="step"` on the active step.
- Supports a `linear` mode (steps must be completed in order; the next unreachable step renders `locked`), `disabled` steps, a `loading` state (with `aria-busy`), a `completed` checkmark state, and a `bullet` slot for custom markers (e.g. `<wa-icon>`).
- `wa-step-change` (cancelable, fires before the change) and `wa-after-step-change` events, plus `goTo()`/`next()`/`previous()` methods and `data-stepper="{command} {id}"` declarative invokers, matches `<wa-dialog>`'s `data-dialog="open {id}"` convention.
- Horizontal steppers scroll instead of squeezing labels when the viewport shrinks, borrowed from `<wa-tab-group>`'s scrollable-nav pattern: prev/next scroll buttons, `ResizeObserver`, auto-scroll, and a `without-scroll-controls` opt-out.
- Marker and connector appearance is customizable with `--marker-*`/`--connector-*`/`--gap` custom properties and a cosmetic `variant` (tints an otherwise untouched marker), plus exported CSS parts for deeper overrides.


claviska:
This is looking good!

We should probably use success for successful steps. The current design makes it look the same as active.

CleanShot 2026-09-15 at 5 38 34 PM@2x

Re: navigation: right now, users can click on any step and go to it and the only way to prevent it is for developers to wire up JavaScript, listen for events, and cancel them. What if we added a property that makes navigation opt in instead of opt out? I think it's more common for the software to guide people through than for people to guide themselves and jump around, so it's probably a better default.

claviska:
I just noticed the skip to pan button doesn't work in this example. 

CleanShot 2026-09-16 at 3 55 54 PM@2x

And thanks for updating the completed steps to success. I think [Brian's PR](https://github.com/shoelace-style/webawesome/pull/2815) does a much better job than my suggestion so I'll defer that decision to him 😆

claviska:
Looks like we just have one example to fix up and this one is ready to go.

Great work, everyone!

claviska:
On second look it seems that the issue with the example was solved.

claviska (approved):

PR #2796, opened 2026-09-09
Add the experimental `<wa-tag-input>` component

claviska:
This PR adds the experimental `<wa-tag-input>` components, which is a form-associated element that accepts an arbitrary number of tags with APIs to let users limit and validate what goes into each tag.

Suggestions (e.g. a combobox-style dropdown of pre-populated items) are deliberately out of scope for the initial commit.


kelseythejackson:
@claviska you mentioned pasting values in the delimiter section. Would an example of pasting values work here, or is it generally understood that this component can handle that?

packages/webawesome/scripts/agent-skill/choosing-components.md
```diff
@@ -113,6 +113,7 @@ The user types or uploads.
 | Single-line text (incl. email, password, etc.) | `<wa-input>` with the appropriate `type`                     |
 | A number with stepper buttons                  | `<wa-number-input>` (richer than `<wa-input type="number">`) |
 | Multi-line text                                | `<wa-textarea>`                                              |
+| A free-form list of short values (keywords, emails, labels) | `<wa-tag-input>`                                              |
```
talbs: NIT: This blew out the table formatting - can we clean it up?

talbs (dismissed):
Looks good and behaves as expected. I'm cool with this shipping as experimental.

Thanks for getting my tweaks and polish in. I left one NIT that is not a blocker.

kelseythejackson (dismissed):

packages/webawesome/scripts/agent-skill/choosing-components.md
claviska: unblowed it 👌

PR #2805, opened 2026-09-15
Tag Input: Inset the Focus Ring + Step Down Tag Corners

talbs:
Focused tags drew their ring on the box border, and nested tags shared the box's radius. This insets the ring and steps tags down to `--wa-border-radius-s` in tag input, select, and combobox.

- select's `+n` overflow tag now honors `pill` and `size`
- drops two `::slotted` rules in select that never matched
- pairs with https://github.com/shoelace-style/webawesome-pro/pull/307 that tackles Combobox

| Before | After | Details |
|--------|--------|--------|
| focused tag, before | focused tag, after | Focused tag's ring moves inside the tag. Drawn outside, it sat on the box border. |
| image | image | Tag corners 6px → 3px. Same radius as the box didn't read as nested; `--wa-border-radius-s` is what WA already uses for things inside a field. |
| image | image | Same step-down in select, so its tags match tag input's when they sit in one form. |
| image | image | The `+n` overflow tag now gets `pill` and `size` like the tags beside it. |


talbs:
@claude, review this PR.

claviska (approved):

PR #2814, opened 2026-09-16
fix(option): avoid retaining options for unused controllers

### LINKED ISSUE #2813: wa-option retains detached selects while waiting for an unused wa-combobox
### Describe the bug

We use Web Awesome at work for speaker dropdowns in a ProseMirror editor and noticed detached dropdowns accumulating in memory. This reproduces with the free package alone, without ProseMirror.

`wa-option` waits for both `wa-select` and `wa-combobox` before checking its ancestors. When combobox is never registered, its pending `whenDefined()` callbacks retain the option and its removed parent select.

Expected: updating an option in a regular select should not wait for an unrelated combobox or retain discarded dropdowns.

### To Reproduce

1. Open the demo and click **Create and remove 100 selects**.
2. In Chrome DevTools → Memory, click **Collect garbage**, then **Refresh counts after GC** in the demo.
3. The 100 removed selects remain alive, with 200 combobox waits and no combobox registered. Repeat to see accumulation.
4. Change `version` to `3.0.0` in the HTML panel and click Run to compare.

### Demo

https://jsfiddle.net/x0758mL2/

The demo imports only select/option and stores WeakRefs, not strong references to removed elements. Counts before GC alone aren't evidence of a leak.

### Browser / OS

macOS, Apple Silicon. A separate Chromium 136 check of the same create/update/remove sequence after forced GC gave:

| Build | Removed selects still alive | Combobox waits |
| --- | ---: | ---: |
| Published 3.12.0 | 100/100 | 200 |
| Published 3.0.0 | 0/100 | 0 |
| Patched `next` | 0/100 | 0 |

### Additional information

The combobox wait was added in #1838 (3.1.0) and is still present on `next`. This affects pages that never register combobox, rather than depending on subscription status itself.

A fix can find the actual select/combobox ancestor first and wait only for that controller. #2579/#2591 addressed a missing controller method; this is retention while an unrelated definition never resolves.

1rvyn:
Fixes #2813.

Find the option's nearest select/combobox ancestor before calling `whenDefined()`, and wait only for that controller. This prevents unused combobox callbacks from retaining detached options and selects when combobox is never registered. Waiting for an actual parent to be registered is preserved.

Adds regression tests for standalone options and select-contained options, including label/value changes and selected-label updates. Only source and tests are changed.

Validation:
- Build, Prettier, and `git diff --check` pass.
- `CI=true npx web-test-runner --group option`: 54 passed per browser.
- `CI=true npx web-test-runner --group select`: 153 passed, 4 existing skips per browser.
- Both suites run in Chromium and Firefox, with client-only and hydrated fixtures. The new cases fail on unpatched `next` in both browsers/modes.
- Separate Chromium GC check: published 3.12.0 retains 100/100 removed selects; patched `next` collects 100/100 and makes no combobox waits.

The proprietary Pro combobox implementation wasn't tested.


claviska (approved):
Thanks for the detailed report and fix!

I reproduced the leak on `next` and confirmed your change resolves it. Looks good to merge. I'll add the changelog entry on this end!

PR #2832, opened 2026-09-21
add allow, name, and label attributes to zoomable frame

claviska:
This worked stemmed from a [Discord discussion](https://discord.com/channels/739437527907303535/1020332605599662181/1551364396440100865). Also added `label` for better a11y when the iframe's content isn't clear from its context.


packages/webawesome/docs/docs/resources/changelog.md
```diff
@@ -31,8 +31,16 @@ Web Awesome follows <a href="https://semver.org/" class="appearance-plain">Seman
...
+:::added
+
+- Added the `allow`, `name`, and `label` attributes to `<wa-zoomable-frame>`
+
+:::
+
 :::fixed
 
+- Fixed a bug in `<wa-zoomable-frame>` where the internal iframe rendered `referrerpolicy="undefined"` when no referrer policy was set
+- Fixed the React wrapper types for properties whose setters accept more than their getters return, so TypeScript no longer rejects documented values such as a `Date` for `<WaDateInput value>` and `<WaTimeInput value>` or `null` for `<WaSlider value>`
```
talbs: This entry describes the React wrapper types fix, but that change isn't in this PR and isn't on `next` yet. Is it from another branch? Or am I missing something?

packages/webawesome/src/components/zoomable-frame/zoomable-frame.test.ts
```diff
@@ -87,6 +87,47 @@ describe('<wa-zoomable-frame>', () => {
...
+        it('should pass allow through to the iframe', async () => {
+          const el = await fixture<WaZoomableFrame>(
+            html`<wa-zoomable-frame allow="fullscreen; clipboard-write"></wa-zoomable-frame>`,
+          );
+          const iframe = el.shadowRoot!.querySelector('iframe')!;
+          expect(iframe.getAttribute('allow')).to.equal('fullscreen; clipboard-write');
+        });
+
+        it('should apply allow and sandbox before the frame navigates', async () => {
+          const el = await fixture<WaZoomableFrame>(
```
talbs, line 99: Claude sniffed something a bit fishy locally here...

> This test doesn't check timing. It only checks that the attributes exist after render.
>
> A sandboxed frame has no `contentDocument` (it's `null`). Asserting that proves the sandbox was on when the frame loaded.

Disregard if its 👃 was malfunctioning, but I thought it was worth raising.

talbs (dismissed):
A few gut checks on the diff, but if you're cool with them, I'm cool with this.

packages/webawesome/docs/docs/resources/changelog.md
claviska:
```suggestion
```

